PR #183, opened 2023-12-12
chore: Refactor to implement model plugin - part 1

Fei-Guo:
This is the first PR of a series of refactoring changes for moving all model specific configurations into preset directory so that the core controllers will be model agnostic. 

- This change introduces Kaito model plugin.
- This change adds the common model interface.
- This change adds corresponding inference implementation for llama-2-chat model family. 

NOTE: the added code is not activated (used) in the controller yet.

### DIFF
--- a/pkg/utils/plugin/plugin.go
+++ b/pkg/utils/plugin/plugin.go
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+package plugin
+
+import (
+	"sync"
+
+	"github.com/azure/kaito/pkg/inference"
+)
+
+type Registration struct {
+	Name     string
+	Instance inference.Model
+}
+
+type ModelRegister struct {
+	sync.RWMutex
+	models map[string]*Registration
+}
+
+var KaitoModelRegister ModelRegister
+
+// Register allows model to be added
+func (reg *ModelRegister) Register(r *Registration) {
+	reg.Lock()
+	defer reg.Unlock()
+	if r.Name == "" {
+		panic("model name is not specified")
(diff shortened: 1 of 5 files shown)


pkg/utils/plugin/plugin.go
```diff
@@ -0,0 +1,45 @@
...
+	reg.models[r.Name] = r
+}
+
+func (reg *ModelRegister) MustGet(name string) inference.Model {
+	reg.Lock()
+	defer reg.Unlock()
+	if _, ok := reg.models[name]; ok {
+		return reg.models[name].Instance
+	}
+	panic("model is not registered")
```
ishaansehgal99, line 44: should we panic? maybe we just return error here

Fei-Guo: With "Must", we panic.

pkg/utils/plugin/plugin.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/azure/kaito/pkg/inference"
+)
+
+type Registration struct {
+	Name     string
+	Instance inference.Model
+}
+
+type ModelRegister struct {
+	sync.RWMutex
```
ishaansehgal99, line 17: sync.Map?
https://stackoverflow.com/questions/61355008/is-sync-map-the-right-tool-for-this-use-case-or-is-a-rw-mutex-lock-better

Fei-Guo: They are interchangeable. With sync.Map, you need to call Load/Store() and you cannot easily implement MustGet.

ishaansehgal99 (approved):

PR #201, opened 2024-01-08
feat: Part 1 - Add FineTuning API

ishaansehgal99:
PEFT - LoRA for Fine-Tuning LLMs

This PR is Part 1 in PRs that will integrate fine tuning into Kaito. This PR is base API code. Future PRs will allow you to specify custom dataset, load config from configmap, and upload training results as image to ACR.


Fei-Guo (commented):
When you change the location of docker file, you need to update the READMEs that refer to the directory such as https://github.com/Azure/kaito/tree/main/presets/models/llama2.

Replace all string "training" to "tuning", here and there.

Fei-Guo (approved):

PR #208, opened 2024-01-12
docs: update install instructions

pauldotyu:
Some Azure CLI command cleanup and update Helm chart installation instructions to use a values.override.yaml file as some may run into inconsistent results using `yq` to modify values.yaml file in place.


Fei-Guo (approved):

Fei-Guo:
LGTM. Thanks a lot for the clean up!

PR #216, opened 2024-01-25
feat: Organize Inference Files - Part 1

ishaansehgal99:
This is the first part towards the new inference pipeline - organizing inference files into their own folder.

This way we can easier trigger build pipelines (edits to presets/models/inference) as well as maintain/add new types of inference


presets/models/inference/llama2-chat/inference-api.py
```diff
@@ -1,21 +1,21 @@
 # Copyright (c) Microsoft Corporation.
```
ishaansehgal99, line 1: Python just ordered the libraries alphabetically here upon saving - no changes to inference

Fei-Guo (approved):

PR #219, opened 2024-01-26
feat: Part 6 - Add Mistral Model

ishaansehgal99:
This PR adds the Mistral model, and, in effect, provides a blueprint for how models can be added in the future.


ishaansehgal99:
Covered in https://github.com/Azure/kaito/pull/249

PR #259, opened 2024-02-27
fix: Ensured nodes being deleted aren't considered Qualified Nodes

ishaansehgal99:


ishaansehgal99:
Included in https://github.com/Azure/kaito/pull/256

PR #280, opened 2024-03-05
fix: more disk space

ishaansehgal99:
Models require more disk space


Fei-Guo (commented):
You don't change mistral7bInst?

ishaansehgal99:
Included in https://github.com/Azure/kaito/pull/299

PR #298, opened 2024-03-13
feat: Add sample front end

ishaansehgal99:
Falcon-7b-Instruct Example

image


ishaansehgal99:
Included in https://github.com/Azure/kaito/pull/320

PR #318, opened 2024-03-25
security: Apply security best practices

helayoty:
Fixes #313


.github/workflows/preset-image-build.yml
```diff
@@ -81,31 +86,36 @@ jobs:
...
+
       - name: 'Az CLI login'
-        uses: azure/login@v1.6.1
+        uses: azure/login@cb79c773a3cfa27f31f25eb3f677781210c9ce3d # v1.6.1
         with:
-            client-id: ${{ secrets.AZURE_CLIENT_ID }}
-            tenant-id: ${{ secrets.AZURE_TENANT_ID }}
-            allow-no-subscriptions: true
-        
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
```
Tatsinnit, line 115: just small side note: if and when this will get converted into 1es ruler the client and tenanted will not be needed - Depends on the case, I will treat this as outside scope of this PR as this is purely security PR. thanks .

Tatsinnit (approved):
Thank you for this, just minor FYI comment. I think most will be step security wizard generated, also, we can also add depdenabot if that not enabled for future ease.

rajanbajaj (approved):

ishaansehgal99 (approved):

Fei-Guo (approved):

PR #334, opened 2024-04-03
docs: update workspace install namespace

philwelz:
**Reason for Change**:

If you install the the workspace controller to the `kaito-workspace` namespace as described in the docs, you get the following error when installing the workspace model (falcon-7b-instruct for example):

`Error from server (InternalError): error when creating "./kaito/falcon.yaml": Internal error occurred: failed calling webhook "validation.workspace.kaito.sh": failed to call webhook: Post "https://workspace.workspace.svc:9443/?timeout=10s": service "workspace" not found`

So this PR fixes the namespace for the workspace controller.

### DIFF
--- a/charts/kaito/workspace/README.md
+++ b/charts/kaito/workspace/README.md
@@ -8,7 +8,7 @@ export IMG_NAME=workspace
 export IMG_TAG=0.2.2
 helm install workspace ./charts/kaito/workspace  \
 --set image.repository=${REGISTRY}/$(IMG_NAME) --set image.tag=$(IMG_TAG) \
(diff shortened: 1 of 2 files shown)


ishaansehgal99 (changes requested):
Thanks for the PR! I couldn't replicate the issue you mentioned. Could you open an issue with the exact steps you took during the installation? It'd help us troubleshoot this better.

philwelz:
@ishaansehgal99  sure here you go : https://github.com/Azure/kaito/issues/336

helayoty:
The issue has been fixed via #337

PR #338, opened 2024-04-04
docs: Release name fix

ishaansehgal99:
**Reason for Change**:
Correction to readme. Keeps it consistent with how e2e deploys helm chart.


helayoty:
The issue has been fixed via #337

PR #342, opened 2024-04-05
feat: Add parsing logic to fine tuning - Part 5

ishaansehgal99:
**Reason for Change**:
This PR introduces expected data format from the configmap and hence the parsing logic from configmap. It also includes ability to override configmap params using command line args. 

**Notes for Reviewers**:
Note command line args is prefixed by its associated class. I thought this may make it more explicit for users if they chose to edit params which associated class is edited, however this could be removed in favor of no prefixes.


ishaansehgal99:
Logic for parsing has been moved into the controller

PR #346, opened 2024-04-10
feat: Add default configmaps - Part 5

ishaansehgal99:
**Reason for Change**:
Provide a default configmaps in the helm chart.

**Notes for Reviewers**:
An env variable release namespace is passed in for future use so that the controller will be able to find where the configmap is deployed.


charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -21,6 +21,9 @@ rules:
   - apiGroups: [ "" ]
     resources: [ "pods"]
     verbs: ["get","list","watch","create", "update", "patch" ]
+  - apiGroups: [ "" ]
+    resources: [ "configmaps" ]
+    verbs: [ "get","list","watch","create", "delete" ]
```
Fei-Guo, line 26: Do we really need to "create" or "delete" configmaps in the kaito controller?

charts/kaito/workspace/templates/deployment.yaml
```diff
@@ -43,6 +43,8 @@ spec:
                   fieldPath: metadata.namespace
             - name: PRESET_REGISTRY_NAME
               value: {{ .Values.presetRegistryName }}
+            - name: RELEASE_NAMESPACE
```
Fei-Guo, line 46: You don't need this. The controller can find the namespace in the Pod's local directory. 

/var/run/secrets/kubernetes.io/serviceaccount/namespace

ishaansehgal99: ok cool removed

charts/kaito/workspace/templates/clusterrole.yaml
ishaansehgal99: no removed

Fei-Guo (approved):

PR #350, opened 2024-04-12
feat: Setup params for plugins

ishaansehgal99:
**Reason for Change**:
Minor setup tuning image versions and run params. Falcon currently populated, phi-2 and mistral todos.


presets/models/falcon/model.go
```diff
@@ -43,11 +44,19 @@ var (
 		"Falcon40BInstruct": "0.0.5",
 	}
 
+	PresetTuningFalconTagMap = map[string]string{
```
Fei-Guo, line 47: Why do we need a separate tag map for tuning?

ishaansehgal99:
Tuning and Inference combined in same image

PR #352, opened 2024-04-15
feat: Simple Configmap Validation Checks - Part 6

ishaansehgal99:
**Reason for Change**:
This PR adds two simple checks to the tuning configmap - makes sure based on the method specified (LoRa or QLoRa) the correct params are included.  Also checks to make sure all the sections specified in the configmap are recognized.


ishaansehgal99:
Updated to PR https://github.com/Azure/kaito/pull/355

PR #355, opened 2024-04-17
feat: Simple Configmap Validation Checks - Part 6 

ishaansehgal99:
**Reason for Change**:
This PR adds two simple checks to the tuning configmap - makes sure based on the method specified (LoRa or QLoRa) the correct params are included. Also checks to make sure all the sections specified in the configmap are recognized.


api/v1alpha1/params_validation.go
```diff
@@ -0,0 +1,151 @@
+package v1alpha1
```
Fei-Guo, line 1: File header

api/v1alpha1/params_validation.go
```diff
@@ -0,0 +1,151 @@
...
+	DatasetConfig      map[string]interface{} `yaml:"DatasetConfig"`
+	DataCollator       map[string]interface{} `yaml:"DataCollator"`
+}
+
+func validateMethodViaConfigMap(cm *corev1.ConfigMap, methodLowerCase string) *apis.FieldError {
+	trainingConfigYAML, ok := cm.Data["training_config.yaml"]
+	if !ok {
+		return apis.ErrGeneric(fmt.Sprintf("ConfigMap '%s' does not contain 'training_config.yaml' in namespace '%s'", cm.Name, cm.Namespace), "config")
+	}
+	fmt.Printf("1Failed validateMethodViaConfigMap")
```
Fei-Guo: what is the printf?

api/v1alpha1/params_validation.go
```diff
@@ -0,0 +1,151 @@
...
+
+	var trainingConfig TrainingConfig
+	if err := yaml.Unmarshal([]byte(trainingConfigYAML), &trainingConfig); err != nil {
+		return apis.ErrGeneric(fmt.Sprintf("Failed to parse 'training_config.yaml' in ConfigMap '%s' in namespace '%s': %v", cm.Name, cm.Namespace, err), "config")
+	}
+	fmt.Printf("1Failed validateMethodViaConfigMap")
```
Fei-Guo: test log here?

api/v1alpha1/params_validation.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// Validate QuantizationConfig if it exists
+	quantConfig := trainingConfig.QuantizationConfig
+	if quantConfig != nil {
+		// Dynamic field search for quantization settings within ModelConfig
+		loadIn4bit, _ := utils.SearchMap(quantConfig, "load_in_4bit")
+		loadIn8bit, _ := utils.SearchMap(quantConfig, "load_in_8bit")
+
+		loadIn4bitBool, ok4bitBool := loadIn4bit.(bool)
```
Fei-Guo, line 45: You need to do a null check for loadIn4bit and loadIn8bit. 

I don't quite understand the logic here. loadIn4bit should be a string "true" or "false" , right? How to convert a string to a bool? or the yaml deserialization will turn ture/false to 1/0?

api/v1alpha1/params_validation.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+		} else if methodLowerCase == string(TuningMethodQLora) {
+			if !(ok4bitBool && loadIn4bitBool) && !(ok8bitBool && loadIn8bitBool) {
+				return apis.ErrMissingField(fmt.Sprintf("For method 'qlora', either 'load_in_4bit' or 'load_in_8bit' must be true in ConfigMap '%s'", cm.Name), "QuantizationConfig")
+			}
+		}
+	} else if methodLowerCase == string(TuningMethodQLora) {
+		return apis.ErrMissingField(fmt.Sprintf("For method 'qlora', either 'load_in_4bit' or 'load_in_8bit' must be true in ConfigMap '%s'", cm.Name), "QuantizationConfig")
+	}
+	fmt.Printf("HIT validateMethodViaConfigMap!!!")
```
Fei-Guo: test log?

api/v1alpha1/workspace_validation.go
```diff
@@ -86,7 +83,18 @@ func (w *Workspace) validateUpdate(old *Workspace) (errs *apis.FieldError) {
 	return errs
 }
 
-func (r *TuningSpec) validateCreate() (errs *apis.FieldError) {
+func (r *TuningSpec) validateCreate(ctx context.Context) (errs *apis.FieldError) {
+	methodLowerCase := strings.ToLower(string(r.Method))
+	if methodLowerCase != string(TuningMethodLora) && methodLowerCase != string(TuningMethodQLora) {
+		errs = errs.Also(apis.ErrInvalidValue(r.Method, "Method"))
+	}
+	if r.Config == "" {
```
Fei-Guo, line 91: We need to use default template here.

api/v1alpha1/params_validation.go
Fei-Guo: I think if !ok4bitbool, you can fail the check to simplify the follow up code.

Fei-Guo (commented):
Are you going to have two default templates: one for lora and one for qlora? In this pr, you need to upload the default template to the helm deployment.

api/v1alpha1/workspace_validation.go
```diff
@@ -86,7 +83,18 @@ func (w *Workspace) validateUpdate(old *Workspace) (errs *apis.FieldError) {
...
+		errs = errs.Also(apis.ErrMissingField("Config"))
+	} else {
+		if err := r.validateConfigMap(ctx, methodLowerCase); err != nil {
+			errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to evaluate validateConfigMap: %v", err), "Config"))
```
Fei-Guo, line 95: If user specifies the configmap template, the template should be inside the workspace CR's namespace, not the release namespace.

api/v1alpha1/params_validation.go
ishaansehgal99: There are 3 different combinations of results from the casting. 

`loadIn4bitBool, ok4bitBool := loadIn4bit.(bool)`

```
in the case loadIn4bit is nil
loadIn4bitBool is false, and ok4bitBool is false

in the case loadIn4bit is false
loadIn4bitBool is false, and ok4bitBool is true

in the case loadIn4bit is true
loadIn4bitBool is true, and ok4bitBool is true
```

api/v1alpha1/workspace_validation.go
ishaansehgal99: Yes good point, if user uses a template - configmap will exist in both release and workspace namespace (configmap can only be mounted to a pod in the same namespace). 

If a user specifies own template - they are responsible for creating configmap in advance in workspace namespace.

api/v1alpha1/workspace_validation.go
ishaansehgal99: Got it, message printed. Controller will handle using default.

api/v1alpha1/params_validation.go
ishaansehgal99: Removed

api/v1alpha1/params_validation.go
ishaansehgal99: Removed

api/v1alpha1/params_validation.go
ishaansehgal99: Removed

api/v1alpha1/workspace_validation.go
```diff
@@ -86,7 +88,27 @@ func (w *Workspace) validateUpdate(old *Workspace) (errs *apis.FieldError) {
 	return errs
 }
 
-func (r *TuningSpec) validateCreate() (errs *apis.FieldError) {
+func (r *TuningSpec) validateCreate(ctx context.Context, workspaceNamespace string) (errs *apis.FieldError) {
+	methodLowerCase := strings.ToLower(string(r.Method))
+	if methodLowerCase != string(TuningMethodLora) && methodLowerCase != string(TuningMethodQLora) {
+		errs = errs.Also(apis.ErrInvalidValue(r.Method, "Method"))
+	}
+	if r.ConfigTemplate == "" {
```
Fei-Guo, line 96: Here is the suggestion:

If the congfigTemplate is empty, we still validate the defaultLoraConfigMap and defaultQloraCongMap because users may change the contents.

As long as the configTemplate is specified, the configMap has to be in WS's namespace. This is consistent with the API definition.

api/v1alpha1/workspace_validation.go
```diff
@@ -21,6 +23,9 @@ import (
 const (
 	N_SERIES_PREFIX = "Standard_N"
 	D_SERIES_PREFIX = "Standard_D"
+
+	DefaultLoraConfigMap  = "lora-params"
```
Fei-Guo, line 27: DefaultLoraConfigMap->DefaultLoraConfigMap-Template,

"lora-params"-> "lora-params-template"

To make it clear, we need to add keyword template in the names.

ishaansehgal99: done

api/v1alpha1/workspace_validation.go
ishaansehgal99: done

Fei-Guo (approved):

PR #357, opened 2024-04-19
feat: [SKU modularization] sku handler interface

smritidahal653:
**Reason for Change**:
Adding a CloudSKUHandler interface that will be used to modularize gpu skus across multiple cloud platforms

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:
The cloud specific implementations and fixes/updates to current code will be part of future PRs.


ishaansehgal99 (commented):
Now that we are extending this functionality, I would suggest we place this and sku_config.go in `pkg` folder. not api

Fei-Guo:
> Now that we are extending this functionality, I would suggest we place this and sku_config.go in `pkg/resources` folder. or introducing a new folder like `pkg/cloud` or `pkg/nodes`

Yes. or just pkg/sku.

api/v1alpha1/sku_handler/cloud_sku_handler.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package skuhandler
+
+type CloudSKUHandler interface {
+	GetSupportedSKUs() string
+	IsSupportedSKU(sku string) (GPUConfig, bool)
```
Fei-Guo, line 8: Why return GPUConfig?

api/v1alpha1/sku_handler/cloud_sku_handler.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package skuhandler
+
+type CloudSKUHandler interface {
+	GetSupportedSKUs() string
```
Fei-Guo, line 7: Why return a string, instead of an array of strings?

api/v1alpha1/sku_handler/cloud_sku_handler.go
smritidahal653: During validation, we need the GPUcount and GPUMem information: https://github.com/Azure/kaito/blob/6be8a0d57225f776d1fe01bd6069f7d3f894eff0/api/v1alpha1/workspace_validation.go#L247

api/v1alpha1/sku_handler/cloud_sku_handler.go
smritidahal653: This method is currently used to print out the supported sku list to users: https://github.com/Azure/kaito/blob/6be8a0d57225f776d1fe01bd6069f7d3f894eff0/api/v1alpha1/workspace_validation.go#L274

smritidahal653:
> Now that we are extending this functionality, I would suggest we place this and sku_config.go in `pkg/resources` folder. or introducing a new folder like `pkg/cloud` or `pkg/nodes`

I am going to move cloud_sku_handler.go to pkg/sku. However, I am going to leave sku_config as is right now since the file will be eventually deleted. I will move all implementations of CloudSKUHandler to the same package though.

api/v1alpha1/sku_handler/cloud_sku_handler.go
Fei-Guo: The inference method name "IsSupportedSKU" clearly indicates that the return type should be just a bool.

api/v1alpha1/sku_handler/cloud_sku_handler.go
Fei-Guo: Are you going to use a string type to represent a list? This is a poor design for an interface method.

smritidahal653: That makes sense, I have changed it to an array

api/v1alpha1/sku_handler/cloud_sku_handler.go
smritidahal653: I gave this some thought and changed the method to GetGPUConfigMap which just returns the map as is. Maybe we won't need the GetSupportedSKUs method after this but I think its better to keep it to keep the code clean

pkg/sku/cloud_sku_handler.go
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package sku
+
+type CloudSKUHandler interface {
+	GetSupportedSKUs() []string
+	GetGPUConfigMap() map[string]GPUConfig
```
Fei-Guo, line 8: GetGPUConfigMap -> GetGPUConfig.    Map is a data type name which should not be included in the interface method name.

pkg/sku/cloud_sku_handler.go
```diff
@@ -0,0 +1,16 @@
...
+}
+
+type GPUConfig struct {
+	SKU      string
+	GPUCount float64
```
Fei-Guo, line 13: Why would GPUCount be float64 type?

smritidahal653: There are some Azure GPU skus with fractional GPU count https://learn.microsoft.com/en-us/azure/virtual-machines/ngads-v-620-series

smritidahal653: changed to int as we decided partial gpu skus won't be supported

Fei-Guo (approved):

PR #358, opened 2024-04-19
feat: Setup Preset Tuning Util Functions and miscellaneous validation/logging - Part 7

ishaansehgal99:
**Reason for Change**:
This PR introduces some of the util functions that are to be used by the function

func CreatePresetTuning(ctx context.Context, workspaceObj *kaitov1alpha1.Workspace,
	tuningObj *model.PresetParam, kubeClient client.Client) (client.Object, error)
	

This function will be in charge of handling launching the training job using the dataset source and uploading results to dataset destination. It is also incharge of ensuring configmap tuning parameters are in the right namespace so they are ready to be used by the fine_tuning_api.py file.

### DIFF
--- a/api/v1alpha1/workspace_validation.go
+++ b/api/v1alpha1/workspace_validation.go
@@ -163,6 +163,12 @@ func (r *DataSource) validateCreate() (errs *apis.FieldError) {
 		sourcesSpecified++
 	}
 	if r.Image != "" {
+		if !strings.Contains(r.Image, ":") {
+			klog.Infof("Tag not specified for image destination (%s), defaulting to 'latest' tag.", r.Image)
+		}
+		if !strings.Contains(r.Image, "/") {
+			klog.Warningf("Input image specification may be incomplete or incorrect: %s. A complete image URL is recommended.", r.Image)
+		}
 		sourcesSpecified++
 	}
 
@@ -216,6 +222,9 @@ func (r *DataDestination) validateCreate() (errs *apis.FieldError) {
 	if r.Image != "" {
 		if !strings.Contains(r.Image, ":") {
 			klog.Infof("Tag not specified for image destination (%s), defaulting to 'latest' tag.", r.Image)
+		}
+		if !strings.Contains(r.Image, "/") {
+			klog.Warningf("Output image specification may be incomplete or incorrect: %s. A complete image URL is recommended.", r.Image)
 		}
 		// Cloud Provider requires credentials to push image
 		if r.ImagePushSecret == "" {
--- a/api/v1alpha1/workspace_validation.go
+++ b/api/v1alpha1/workspace_validation.go
@@ -214,6 +214,10 @@ func (r *DataDestination) validateCreate() (errs *apis.FieldError) {
 		destinationsSpecified++
 	}
 	if r.Image != "" {
+		// Cloud Provider requires credentials to push image
+		if r.ImagePushSecret == "" {
+			errs = errs.Also(apis.ErrMissingField("Must specify imagePushSecret with destination image"))
+		}
 		destinationsSpecified++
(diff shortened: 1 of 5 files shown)


charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -23,7 +23,7 @@ rules:
     verbs: ["get","list","watch","create", "update", "patch" ]
   - apiGroups: [ "" ]
     resources: [ "configmaps" ]
-    verbs: [ "get","list","watch" ]
+    verbs: [ "get","list","watch","create", "delete" ]
```
ishaansehgal99, line 26: Needed to now create duplicate configmap from release to workspace namespace

pkg/tuning/preset-tuning.go
```diff
@@ -2,11 +2,142 @@ package tuning
...
+	registryName := os.Getenv("PRESET_REGISTRY_NAME")
+	return formatRegistryImagePath(registryName, "kaito-tuning-"+string(wObj.Tuning.Preset.Name), presetObj.Tag)
+}
+
+func GetDataSrcImageInfo(ctx context.Context, wObj *kaitov1alpha1.Workspace) (string, []corev1.LocalObjectReference) {
+	imagePullSecretRefs := make([]corev1.LocalObjectReference, len(wObj.Tuning.Input.ImagePullSecrets))
+	for i, secretName := range wObj.Tuning.Input.ImagePullSecrets {
+		imagePullSecretRefs[i] = corev1.LocalObjectReference{Name: secretName}
+	}
+	registryName := os.Getenv("PRESET_REGISTRY_NAME")
```
Fei-Guo, line 92: Why do we need to get registry name from env?  Can we get the registry from the Image, 
or in the other word, wObj.Tuning.Input.Image should be a valid image url already.

pkg/tuning/preset-tuning.go
```diff
@@ -2,11 +2,142 @@ package tuning
...
+}
+
+func GetDataDestImageInfo(ctx context.Context, wObj *kaitov1alpha1.Workspace) (string, []corev1.LocalObjectReference) {
+	registryName := os.Getenv("PRESET_REGISTRY_NAME")
+	imageName := fmt.Sprintf("%s/%s", registryName, wObj.Tuning.Output.Image)
+	imagePushSecretRefs := []corev1.LocalObjectReference{{Name: wObj.Tuning.Output.ImagePushSecret}}
+	return imageName, imagePushSecretRefs
+}
+
+func CreatePresetConfigMap(ctx context.Context, workspaceObj *kaitov1alpha1.Workspace,
```
Fei-Guo, line 104: Usually, if we check if the configmap exists before creating a new one, we call this method "EnsurePresetConfigmap".

How about EnsureTuningConfigMap?

Fei-Guo (commented):
remove deadcode.

pkg/tuning/preset-tuning.go
ishaansehgal99: Ok - User will specify full url

pkg/tuning/preset-tuning.go
ishaansehgal99: Changed

api/v1alpha1/workspace_validation.go
```diff
@@ -163,6 +163,12 @@ func (r *DataSource) validateCreate() (errs *apis.FieldError) {
 		sourcesSpecified++
 	}
 	if r.Image != "" {
+		if !strings.Contains(r.Image, ":") {
```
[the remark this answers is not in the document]
ishaansehgal99: added regex

Fei-Guo (approved):

PR #360, opened 2024-04-25
feat: [SKU modularization] adding azure sku handler

smritidahal653:
**Reason for Change**:
Adding implementation for Azure sku handler along with unit test

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/sku/azure_sku_handler.go
+++ b/pkg/sku/azure_sku_handler.go
@@ -1,0 +1,58 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package sku
+
+type AzureSKUHandler struct {
+	supportedSKUs map[string]GPUConfig
+}
+
+func NewAzureSKUHandler() *AzureSKUHandler {
+	return &AzureSKUHandler{
+		supportedSKUs: azureSupportedGPUConfigs,
+	}
+}
+
+func (a *AzureSKUHandler) GetSupportedSKUs() []string {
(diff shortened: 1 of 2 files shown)


pkg/sku/azure_sku_handler.go
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package sku
+
+type AzureSKUHandler struct {
+	supportedSKUs map[string]GPUConfig
```
Fei-Guo, line 7: Do we need a lock to protect the supportedSKUs field,

pkg/sku/azure_sku_handler.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+}
+
+func (a *AzureSKUHandler) GetSupportedSKUs() []string {
+	skus := make([]string, 0, len(azureSupportedGPUConfigs))
+	for sku := range azureSupportedGPUConfigs {
+		skus = append(skus, sku)
+	}
+	return skus
+}
```
ishaansehgal99, line 22: recreates a slice from a map, could be used elsewhere as well. move to utills.

ishaansehgal99: nvm this is ok

smritidahal653: This logic would be reused by other cloud provider implementations as well so I moved it to a helper file sku_utils

Fei-Guo (approved):

PR #365, opened 2024-04-29
feat: Add URL as data source

ishaansehgal99:
**Reason for Change**:
Add URL as a data source
**Requirements**


Fei-Guo (approved):
Not in this change, but can you refactor the ConfigDataVolume method?  "" input is hard to understand. Maybe changing the input hostpath to *string type and use nil if hostpath is not configured.

Issue #339, opened 2024-04-04
Kaito workspace-controller support for Karpenter nodeclaim


philwelz:
**Describe the bug**

Kaito workspace controller seems to be only compatible with Karpenter version prior [v0.33.0 ](https://github.com/kubernetes-sigs/karpenter/releases/tag/v0.33.0 ) as they deprecated the machine CRD (karpenter.sh/v1alpha5) in that release and the controller seems to relay on that machine CRD for spinning up the workspace.

**Steps To Reproduce**

1. [Create an AKS cluster with NAP](https://learn.microsoft.com/en-us/azure/aks/node-autoprovision?tabs=azure-cli) enabled
2. Install Kaito workspace-controller `helm install workspace kaito/workspace --namespace workspace --create-namespace`
3. Add an workspace `kubectl apply -f https://raw.githubusercontent.com/Azure/kaito/main/examples/inference/kaito_workspace_phi-2.yaml`
4. Nothing happens, run `kubectl api-resouces` and see there is no CRD `machines karpenter.sh/v1alpha5`

**Expected behavior**

As Kaito per docs supports node provisioning controllers that supports Karpenter-core APIs, it should also support the new Karpenter CRDs for nodeClaims (machine CRD was deprecated in December). In best-case scenario Kaito (workspace-controller) should also run on AKS with NAP enabled and should be aware of the Karpenter version (or which CRD is available: machines vs nodeclaim).

**Logs**
Controller starts and prints this:

`2024-04-04T10:07:49Z	INFO	Starting EventSource	{"controller": "workspace", "controllerGroup": "kaito.sh", "controllerKind": "Workspace", "source": "kind source: *v1alpha5.Machine"}
`

**Environment**

- Kubernetes version (use `kubectl version`): 1.28.5
- Install tools: AKS NAP (Karpenter)

### Support Karpenter Tasks
- [x] Add NodeClaim API #362
- [x] Update workspace controller to support NodeClaim #366
- [x] Update pipelines #569
- [x] update chart #570
- [x] Add e2e tests  #375
- [ ] Add label to the nodeclaims to identify the node created by kaito
- [ ] Update doc

Fei-Guo:
Thanks for filing the issue. We are working towards supporting Karpenter new APIs.

zhuangqh:
All core functionality for Karpenter NodeClaim support has been implemented and tested. The remaining documentation task can be tracked separately if needed. Closing as completed. Thanks everyone! 🎉

PR #366, opened 2024-04-30
feat: Support Karpenter NodeClaim API in workspace

helayoty:
**Reason for Change**:
- Workspace controller can support `nodeClaims` API.

Workspace controller won't migrate the current `machine` nodes. The behavior for CRUD operation:

**Create**
   - Both `nodeClaims` or `Machine` APIs will be used to create a node based on featureGate,

**Ownership**
   - Both `machines` and `nodeClaims` objects which created by the same workspace will be associated to it until got deleted. 

**Delete**
   - Deleting the workspace will delete both (if any) old `machines` objects and `nodeClaims`.

- Add unit tests for supporting node claims, and managing migrated clusters (both machines and nodeClaims).
- Replace the deprecated func `FromInt` with `FromInt32`.
- Update error message lint warning, start with lowercase characters.

**Requirements**

**Issue Fixed**:
#339 

**Notes for Reviewers**:

### DIFF
--- a/pkg/controllers/workspace_controller.go
+++ b/pkg/controllers/workspace_controller.go
@@ -229,7 +229,7 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
 
 	if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
 		// Check nodeClaims only if the k8s version is greater than 1.29. As nodeClaim is supported from 1.29.
-		if c.KubernetesVersion.Major >= consts.KubernetesSupportedMajorVersion {
+		if c.KubernetesVersion.Major >= consts.RequiredKubernetesVersionForNodeClaim {
 			// Wait for pending nodeClaims if any before we decide whether to create new node or not.
 			if err := nodeclaim.WaitForPendingNodeClaims(ctx, wObj, c.Client); err != nil {
 				return err
@@ -249,15 +249,16 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
 
 	if newNodesCount > 0 {
 		klog.InfoS("need to create more nodes", "NodeCount", newNodesCount)
-		if featuregates.FeatureGates[consts.FeatureFlagKarpenter] &&
-			c.KubernetesVersion.Major < consts.KubernetesSupportedMajorVersion {
-			if err := c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeMachineStatus, metav1.ConditionUnknown,
+		if c.KubernetesVersion.Major < consts.RequiredKubernetesVersionForNodeClaim {
+			if err := c.updateStatusConditionIfNotMatch(ctx, wObj,
+				kaitov1alpha1.WorkspaceConditionTypeMachineStatus, metav1.ConditionUnknown,
 				"CreateMachinePending", fmt.Sprintf("creating %d machines", newNodesCount)); err != nil {
 				klog.ErrorS(err, "failed to update workspace status", "workspace", klog.KObj(wObj))
 				return err
 			}
-		} else {
-			if err := c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeNodeClaimStatus, metav1.ConditionUnknown,
+		} else if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
+			if err := c.updateStatusConditionIfNotMatch(ctx, wObj,
+				kaitov1alpha1.WorkspaceConditionTypeNodeClaimStatus, metav1.ConditionUnknown,
 				"CreateNodeClaimPending", fmt.Sprintf("creating %d nodeClaims", newNodesCount)); err != nil {
 				klog.ErrorS(err, "failed to update workspace status", "workspace", klog.KObj(wObj))
 				return err
@@ -293,15 +294,16 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
 		}
 	}
 
-	if featuregates.FeatureGates[consts.FeatureFlagKarpenter] &&
-		c.KubernetesVersion.Major < consts.KubernetesSupportedMajorVersion {
-		if err = c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeMachineStatus, metav1.ConditionTrue,
+	if c.KubernetesVersion.Major < consts.RequiredKubernetesVersionForNodeClaim {
+		if err = c.updateStatusConditionIfNotMatch(ctx, wObj,
+			kaitov1alpha1.WorkspaceConditionTypeMachineStatus, metav1.ConditionTrue,
 			"installNodePluginsSuccess", "machines plugins have been installed successfully"); err != nil {
 			klog.ErrorS(err, "failed to update workspace status", "workspace", klog.KObj(wObj))
 			return err
 		}
-	} else {
-		if err = c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeNodeClaimStatus, metav1.ConditionTrue,
+	} else if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
+		if err = c.updateStatusConditionIfNotMatch(ctx, wObj,
+			kaitov1alpha1.WorkspaceConditionTypeNodeClaimStatus, metav1.ConditionTrue,
 			"installNodePluginsSuccess", "nodeClaim plugins have been installed successfully"); err != nil {
 			klog.ErrorS(err, "failed to update workspace status", "workspace", klog.KObj(wObj))
 			return err
@@ -376,18 +378,20 @@ func (c *WorkspaceReconciler) createAndValidateNode(ctx context.Context, wObj *k
 	var nodeOSDiskSize string
 	if wObj.Inference != nil && wObj.Inference.Preset != nil && wObj.Inference.Preset.Name != "" {
 		presetName := string(wObj.Inference.Preset.Name)
-		nodeOSDiskSize = plugin.KaitoModelRegister.MustGet(presetName).GetInferenceParameters().DiskStorageRequirement
+		nodeOSDiskSize = plugin.KaitoModelRegister.MustGet(presetName).
+			GetInferenceParameters().DiskStorageRequirement
 	}
 	if nodeOSDiskSize == "" {
 		nodeOSDiskSize = "0" // The default OS size is used
 	}
 
 	// If	the k8s version is less than 1.29, create machine, else create nodeClaim.
-	if c.KubernetesVersion.Major < "29" {
+	if c.KubernetesVersion.Major < consts.RequiredKubernetesVersionForNodeClaim {
 		return c.CreateMachine(ctx, wObj, nodeOSDiskSize)
-	} else {
+	} else if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
 		return c.CreateNodeClaim(ctx, wObj, nodeOSDiskSize)
 	}
+	return nil, fmt.Errorf("no node has been created")
 }
 
 func (c *WorkspaceReconciler) CreateMachine(ctx context.Context, wObj *kaitov1alpha1.Workspace, nodeOSDiskSize string) (*corev1.Node, error) {
@@ -396,7 +400,7 @@ RetryWithDifferentName:
 
 	if err := machine.CreateMachine(ctx, newMachine, c.Client); err != nil {
 		if apierrors.IsAlreadyExists(err) {
-			klog.InfoS("There exists a machine with the same name, retry with a different name", "machine", klog.KObj(newMachine))
+			klog.InfoS("A machine exists with the same name, retry with a different name", "machine", klog.KObj(newMachine))
 			goto RetryWithDifferentName
 		} else {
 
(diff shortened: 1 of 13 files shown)


bangqipropel:
what happened between k8s 1.27 and 1.29?

pkg/controllers/workspace_controller.go
```diff
@@ -313,24 +364,33 @@ func (c *WorkspaceReconciler) validateNodeInstanceType(ctx context.Context, wObj
...
-	if machineOSDiskSize == "" {
-		machineOSDiskSize = "0" // The default OS size is used
+	if nodeOSDiskSize == "" {
+		nodeOSDiskSize = "0" // The default OS size is used
 	}
 
-Retry_withdifferentname:
-	newMachine := machine.GenerateMachineManifest(ctx, machineOSDiskSize, wObj)
+	// If	the k8s version is less than 1.29, create machine, else create nodeClaim.
+	if c.KubernetesVersion.Major < "29" {
```
ishaansehgal99, line 379: move "29" toconstant

helayoty: addressed

helayoty:
> what happened between k8s 1.27 and 1.29?

@bangqipropel any k8s version before 1.29 will use `machine` API. Otherwise, the `nodeClaim` API will be used.

bangqipropel:
> > what happened between k8s 1.27 and 1.29?
> 
> @bangqipropel any k8s version before 1.29 will use `machine` API. Otherwise, the `nodeClaim` API will be used.

I see, before I saw 1.27 also, that was why I asked

pkg/controllers/workspace_controller.go
```diff
@@ -62,6 +68,18 @@ func (c *WorkspaceReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 
 	klog.InfoS("Reconciling", "workspace", req.NamespacedName)
 
+	// Get the current API Server version.
```
Fei-Guo, line 71: Guard this code with feature  gate?

pkg/controllers/workspace_controller.go
```diff
@@ -168,25 +186,36 @@ func selectWorkspaceNodes(qualified []*corev1.Node, preferred []string, previous
...
 				// Choose node created by gpu-provisioner since it is more likely to be empty to use.
-				if iCreatedByKaito && !jCreatedByKaito {
+				if iCreatedByGPUProvisioner && !jCreatedByGPUProvisioner {
 					return true
-				} else if !iCreatedByKaito && jCreatedByKaito {
+				} else if !iCreatedByGPUProvisioner && jCreatedByGPUProvisioner {
 					return false
 				} else {
-					return qualified[i].Name < qualified[j].Name
+					var iCreatedByKarpenter, jCreatedByKarpenter bool
```
Fei-Guo, line 198: Guard this code with feature gate?

pkg/controllers/workspace_controller.go
```diff
@@ -198,22 +227,41 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
 		return err
 	}
 
-	// Find all nodes that match the labelSelector and instanceType, they are not necessarily created by machines.
+	if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
+		// Check nodeClaims only if the k8s version is greater than 1.29. As nodeClaim is supported from 1.29.
+		if c.KubernetesVersion.Major >= consts.KubernetesSupportedMajorVersion {
```
Fei-Guo, line 232: KubernetesSupportedMajorVersion, this name is confusing. Shoudn't it be requiredVersionForNodeClaim?

pkg/controllers/workspace_controller.go
```diff
@@ -198,22 +227,41 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
...
 
 	newNodesCount := lo.FromPtr(wObj.Resource.Count) - len(selectedNodes)
 
 	if newNodesCount > 0 {
 		klog.InfoS("need to create more nodes", "NodeCount", newNodesCount)
-		if err := c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeMachineStatus, metav1.ConditionUnknown,
-			"CreateMachinePending", fmt.Sprintf("creating %d machines", newNodesCount)); err != nil {
-			klog.ErrorS(err, "failed to update workspace status", "workspace", klog.KObj(wObj))
-			return err
+		if featuregates.FeatureGates[consts.FeatureFlagKarpenter] &&
```
Fei-Guo, line 252: The logic seems to be wrong. If featuregate is enabled, we should go to the else path, right?

pkg/controllers/workspace_controller.go
```diff
@@ -245,10 +293,19 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
 		}
 	}
 
-	if err = c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeMachineStatus, metav1.ConditionTrue,
-		"installNodePluginsSuccess", "machines plugins have been installed successfully"); err != nil {
-		klog.ErrorS(err, "failed to update workspace status", "workspace", klog.KObj(wObj))
-		return err
+	if featuregates.FeatureGates[consts.FeatureFlagKarpenter] &&
```
Fei-Guo, line 296: Same here, if featuregate is enabled, the code should go to else path.

pkg/controllers/workspace_controller.go
```diff
@@ -314,24 +371,33 @@ func (c *WorkspaceReconciler) validateNodeInstanceType(ctx context.Context, wObj
...
-	if machineOSDiskSize == "" {
-		machineOSDiskSize = "0" // The default OS size is used
+	if nodeOSDiskSize == "" {
+		nodeOSDiskSize = "0" // The default OS size is used
 	}
 
-Retry_withdifferentname:
-	newMachine := machine.GenerateMachineManifest(ctx, machineOSDiskSize, wObj)
+	// If	the k8s version is less than 1.29, create machine, else create nodeClaim.
+	if c.KubernetesVersion.Major < "29" {
```
Fei-Guo, line 386: Add feature gate.

pkg/controllers/workspace_controller.go
```diff
@@ -549,7 +649,8 @@ func (c *WorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&kaitov1alpha1.Workspace{}).
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.StatefulSet{}).
-		Watches(&v1alpha5.Machine{}, c.watchMachines()).
+		Watches(&v1alpha5.Machine{}, c.watchMachines()).    // watches for machine with labels indicating workspace name.
+		Watches(&v1beta1.NodeClaim{}, c.watchNodeClaims()). // watches for nodeClaim with labels indicating workspace name.
```
Fei-Guo, line 653: Guard with feature gate. The watch call requires the nodeclaim CRD to be installed, otherwise, it will throw error.

pkg/controllers/workspace_gc_finalizer.go
```diff
@@ -30,6 +32,22 @@ func (c *WorkspaceReconciler) garbageCollectWorkspace(ctx context.Context, wObj
 		}
 	}
 
+	if c.KubernetesVersion.Major >= "29" {
```
Fei-Guo, line 35: guard with featuregate. We suppose to remove hardcode 29?

pkg/controllers/workspace_controller.go
helayoty: There is no harm from getting the k8s version with or without the feature gates.

pkg/controllers/workspace_controller.go
helayoty: they will be always nil as no nodeclaims are available but I'm ok to add feature gate.

pkg/controllers/workspace_controller.go
helayoty: The feature flag condition should go to the else. Otherwise, the logic is correct.

pkg/controllers/workspace_controller.go
helayoty: Renamed to `RequiredKubernetesVersionForNodeClaim`

pkg/controllers/workspace_controller.go
helayoty: The feature flag condition should go to the else (addressed). Otherwise, the logic is correct.

pkg/controllers/workspace_controller.go
helayoty: Addressed

pkg/controllers/workspace_gc_finalizer.go
helayoty: Addressed.

pkg/controllers/workspace_controller.go
```diff
@@ -168,13 +173,20 @@ func selectWorkspaceNodes(qualified []*corev1.Node, preferred []string, previous
...
+				_, iCreatedByGPUProvisioner = qualified[i].Labels[machine.LabelGPUProvisionerCustom]
+				_, jCreatedByGPUProvisioner = qualified[j].Labels[machine.LabelGPUProvisionerCustom]
+				// Choose node created by gpu-provisioner and karpenter since it is more likely to be empty to use.
+				var iCreatedByKarpenter, jCreatedByKarpenter bool
+				if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
+					_, iCreatedByKarpenter = qualified[i].Labels[nodeclaim.LabelNodePool]
+					_, jCreatedByKarpenter = qualified[j].Labels[nodeclaim.LabelNodePool]
+				}
+				if (iCreatedByGPUProvisioner && !jCreatedByGPUProvisioner) ||
+					(iCreatedByKarpenter && !jCreatedByKarpenter) {
```
ishaansehgal99, line 186: So this is handling the selection criteria of eligible nodes if there are both gpu-provisioner and karpenter nodes?

pkg/controllers/workspace_controller.go
```diff
@@ -149,7 +154,7 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
 	return c.garbageCollectWorkspace(ctx, wObj)
 }
 
-func selectWorkspaceNodes(qualified []*corev1.Node, preferred []string, previous []string, count int) []*corev1.Node {
+func (c *WorkspaceReconciler) selectWorkspaceNodes(qualified []*corev1.Node, preferred []string, previous []string, count int) []*corev1.Node {
 
 	sort.Slice(qualified, func(i, j int) bool {
```
ishaansehgal99, line 159: I recommend adding a comment here on selection critieria - something high level, like 1. Preferred 2. Previous usage 3. karpenter made etc.

pkg/controllers/workspace_controller.go
ishaansehgal99: Oh I see, all good

Fei-Guo (approved):

pkg/controllers/workspace_controller.go
ishaansehgal99, line 159:
```suggestion
func (c *WorkspaceReconciler) selectWorkspaceNodes(qualified []*corev1.Node, preferred []string, previous []string, count int) []*corev1.Node {
	isPreferred := func(node *corev1.Node) bool {
		return utils.Contains(preferred, node.Name)
	}
	isPrevious := func(node *corev1.Node) bool {
		return utils.Contains(previous, node.Name)
	}
	isCreatedByGPUProvisioner := func(node *corev1.Node) bool {
		_, exists := node.Labels[machine.LabelGPUProvisionerCustom]
		return exists
	}
	isCreatedByKarpenter := func(node *corev1.Node) bool {
		if featuregates.FeatureGates[consts.FeatureFlagKarpenter] {
			_, exists := node.Labels[nodeclaim.LabelNodePool]
			return exists
		}
		return false
	}


	sort.Slice(qualified, func(i, j int) bool {
		// Priority order: preferred, previous, GPUProvisioner, Karpenter, name
		if isPreferred(qualified[i]) != isPreferred(qualified[j]) {
			return isPreferred(qualified[i])
		}
		if isPrevious(qualified[i]) != isPrevious(qualified[j]) {
			return isPrevious(qualified[i])
		}
		if isCreatedByGPUProvisioner(qualified[i]) != isCreatedByGPUProvisioner(qualified[j]) {
			return isCreatedByGPUProvisioner(qualified[i])
		}
		if isCreatedByKarpenter(qualified[i]) != isCreatedByKarpenter(qualified[j]) {
			return isCreatedByKarpenter(qualified[i])
		}
		// Default to alphabetical order if all else equal
		return qualified[i].Name < qualified[j].Name
	})

	// Return all qualified nodes if requested count is >= to list size.
	if len(qualified) <= count {
		return qualified
	}
	return qualified[:count]
}

```

pkg/controllers/workspace_controller.go
```diff
@@ -359,6 +403,41 @@ Retry_withdifferentname:
...
 }
 
+func (c *WorkspaceReconciler) CreateNodeClaim(ctx context.Context, wObj *kaitov1alpha1.Workspace, nodeOSDiskSize string) (*corev1.Node, error) {
+RetryWithDifferentName:
+	newNodeClaim := nodeclaim.GenerateNodeClaimManifest(ctx, nodeOSDiskSize, wObj)
+
+	if err := nodeclaim.CreateNodeClaim(ctx, newNodeClaim, c.Client); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			klog.InfoS("There exists a nodeClaim with the same name, retry with a different name", "nodeClaim", klog.KObj(newNodeClaim))
+			goto RetryWithDifferentName
```
ishaansehgal99, line 413: Consider for loop...to avoid infinite loop and better code readability

const maxRetries = 5
for i := 0; i < maxRetries; i++ {
newNodeClaim := nodeclaim.GenerateNodeClaimManifest(ctx, nodeOSDiskSize, wObj)

       if err := nodeclaim.CreateNodeClaim(ctx, newNodeClaim, c.Client); err != nil {
            if apierrors.IsAlreadyExists(err) {
                klog.InfoS("There exists a nodeClaim with the same name, retrying with a different name", "nodeClaim", klog.KObj(newNodeClaim))
                continue
            }
            ....

pkg/controllers/workspace_controller.go
```diff
@@ -359,6 +403,41 @@ Retry_withdifferentname:
...
+		} else {
+
+			klog.ErrorS(err, "failed to create nodeClaim", "nodeClaim", newNodeClaim.Name)
+			if updateErr := c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeNodeClaimStatus, metav1.ConditionFalse,
+				"nodeClaimFailedCreation", err.Error()); updateErr != nil {
+				klog.ErrorS(updateErr, "failed to update workspace status", "workspace", klog.KObj(wObj))
+				return nil, updateErr
+			}
```
ishaansehgal99, line 421: The pattern of printing error, updating status and catching update status error is repeated. Consider making this own function like
func (c *WorkspaceReconciler) handleError(ctx context.Context, wObj *kaitov1alpha1.Workspace, reason string, err error) error {
klog error
update status
klog if any update status error
return err
}

pkg/controllers/workspace_controller.go
```diff
@@ -577,3 +661,27 @@ func (c *WorkspaceReconciler) watchMachines() handler.EventHandler {
...
+			return []reconcile.Request{
+				{
+					NamespacedName: client.ObjectKey{
+						Name:      name,
+						Namespace: namespace,
+					},
+				},
+			}
+		})
+}
```
ishaansehgal99, line 687: wow super cool. 
I was at first scratching my head...I thought maybe if nodeclaim gets updated automatically other objects would get reconciled. But I realize now not. 

if you modify a nodeclaim it would only run reconcile on that object. What this is doing is also ensuring reconcile on workspace object! Cool!

pkg/controllers/workspace_controller.go
helayoty: But we don't want to quit until we find a new name.

pkg/controllers/workspace_controller.go
helayoty: Will add in a follow up pr

PR #367, opened 2024-04-30
feat: Fine tune (Part 9) - Handling image data destination

ishaansehgal99:
**Reason for Change**:
Sets up the data destination outline. This PR introduces image as first data destination. It includes sidecar container script for detecting finished fine tuning results and uploading results to image data destination

PR also fixes a flaky test - caused by environment variables getting carried over between tests.


charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -39,7 +39,10 @@ rules:
   - apiGroups: ["admissionregistration.k8s.io"]
     resources: ["validatingwebhookconfigurations"]
     verbs: ["get","list","watch"]
+  - apiGroups: [ "batch" ]
+    resources: [ "jobs" ]
+    verbs: [ "get", "list", "watch", "create", "delete", "update", "patch" ]
```
helayoty, line 44: Do we need create, delete, update, and patch roles?

charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -39,7 +39,10 @@ rules:
...
   - apiGroups: ["admissionregistration.k8s.io"]
     resources: ["validatingwebhookconfigurations"]
     verbs: ["update"]
-    resourceNames: ["validation.workspace.kaito.sh"]
+    resourceNames: ["validation.workspace.kaito.sh"]
```
helayoty, line 48: nit: add new line. [why?](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)

charts/kaito/workspace/templates/clusterrole.yaml
ishaansehgal99: removed

ishaansehgal99: Upon further debugging we actually need these roles - just like the deployments and statefulsets workloads managed by the controller. So is job for fine tuning. If user for example deletes their job controller needs permission to recreate it.

pkg/resources/manifests.go
```diff
@@ -184,18 +185,74 @@ func GenerateStatefulSetManifest(ctx context.Context, workspaceObj *kaitov1alpha
 	return ss
 }
 
+func dockerSidecarScriptPushToVolume(arg interface{}) string {
```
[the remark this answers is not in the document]
ishaansehgal99: Yes I refactored code to be more modular now for this.

pkg/tuning/preset-tuning.go
```diff
@@ -129,10 +162,55 @@ func CreatePresetTuning(ctx context.Context, workspaceObj *kaitov1alpha1.Workspa
...
+		initContainers = append(initContainers, initContainer)
+	}
+
+	sidecarContainer, imagePushSecret, dataDestVolume, dataDestVolumeMount, err := prepareDataDestination(ctx, workspaceObj)
+	if err != nil {
+		return nil, err
+	}
+	volumes = append(volumes, dataDestVolume)
+	volumeMounts = append(volumeMounts, dataDestVolumeMount)
+	if sidecarContainer.Name != "" {
```
Fei-Guo, line 202: I still prefer a nil check over checking for empty string. sidecarContainer can be defined as a pointer type.
The same for imagePushSecret.

ishaansehgal99: fixed!

Fei-Guo (approved):

PR #368, opened 2024-04-30
feat: Add feature gates flag

helayoty:
**Reason for Change**:
- Add `feature-gates` flag to the Kaito controller.
- Support the `karpenter` feature gate with default value, `false`. 
- Enable setting the feature gate flag via helm chart.

In addition to some chores:
- Exclude the auto generated and mock files from unit test coverage.
- Upgrade k8s version using in the tests to the current supported version.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/featuregates/featuregates.go
+++ b/pkg/featuregates/featuregates.go
@@ -5,22 +5,23 @@ package featuregates
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/azure/kaito/pkg/utils/consts"
 	cliflag "k8s.io/component-base/cli/flag"
 )
 
 var (
-	// FeatureGates is a map that holds	the feature gates and their values for Kaito.
+	// FeatureGates is a map that holds	the feature gates and their default values for Kaito.
 	FeatureGates = map[string]bool{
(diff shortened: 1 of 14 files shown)


pkg/featuregates/featuregates.go
```diff
@@ -0,0 +1,39 @@
...
+	gateMap := map[string]bool{}
+
+	if err := cliflag.NewMapStringBool(&gateMap).Set(featureGates); err != nil {
+		return err
+	}
+	if len(gateMap) == 0 {
+		// no feature gates set
+		return nil
+	}
+	if val, ok := gateMap[consts.FeatureFlagKarpenter]; ok {
```
Fei-Guo, line 31: You'd do sth like this:

for k, v := range gateMap {
       if _, ok := FeatureGates[k]; !ok {
              return errors.New("invalid feature gate")
       }
       FeatureGates[k]=v
}

helayoty: Addressed but with a return of all invalid features. Added more UT.

Fei-Guo (approved):

PR #376, opened 2024-05-09
ci: Add Env variable for supported models yaml

ishaansehgal99:
**Reason for Change**:
Add Env variable for supported models yaml


.github/workflows/e2e-workflow.yml
```diff
@@ -198,6 +198,7 @@ jobs:
           RUN_LLAMA_13B: ${{ env.RUN_LLAMA_13B }}
           AI_MODELS_REGISTRY: ${{ secrets.E2E_ACR_AMRT_USERNAME }}.azurecr.io
           AI_MODELS_REGISTRY_SECRET: ${{ secrets.E2E_AMRT_SECRET_NAME }}
+          SUPPORTED_MODELS_YAML_PATH: "/home/runner/work/kaito/kaito/presets/models/supported_models.yaml"
```
helayoty, line 201: Can we add the default value for the file path in the makefile?

helayoty (changes requested):

helayoty (approved):

PR #439, opened 2024-05-28
feat: Add support for target_modules List

ishaansehgal99:
**Reason for Change**:
`target_modules` LoRAConfig parameter specifies which parts of the model should be adapted using LoRA. It can either be a string or string array. This PR adds support for array type arguments. It also does a validation check that target_modules is either of type string or string array. 

Example usage of target modules here

https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/config.py#L51-L58

https://stackoverflow.com/questions/76768226/target-modules-for-applying-peft-lora-on-different-models

### DIFF
--- a/api/v1alpha1/params_validation.go
+++ b/api/v1alpha1/params_validation.go
@@ -108,6 +108,38 @@ func UnmarshalTrainingConfig(cm *corev1.ConfigMap) (*Config, *apis.FieldError) {
 	return &config, nil
 }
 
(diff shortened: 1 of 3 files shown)


Fei-Guo (commented):
Q: should we validate the content of the string? I don't see much value in validating the data type to be string or array of string.

ishaansehgal99:
Unecessary Validation check

PR #443, opened 2024-05-30
feat: Add Min GPU Memory Requirement

ishaansehgal99:
**Reason for Change**:
1. Add interface field `TuningMethodGPUMemory` - Min GPU memory per tuning method (batch size 1).
2. Fix bug(s) in tuning ConfigMap creation

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -15,23 +15,21 @@ type Model interface {
 
 // PresetParam defines the preset inference parameters for a model.
 type PresetParam struct {
-	ModelFamilyName           string            // The name of the model family.
-	ImageAccessMode           string            // Defines where the Image is Public or Private.
-	DiskStorageRequirement    string            // Disk storage requirements for the model.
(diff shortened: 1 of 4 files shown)


pkg/model/interface.go
```diff
@@ -21,6 +21,7 @@ type PresetParam struct {
 	GPUCountRequirement       string            // Number of GPUs required for the Preset.
 	TotalGPUMemoryRequirement string            // Total GPU memory required for the Preset.
 	PerGPUMemoryRequirement   string            // GPU memory required per GPU.
+	TuningMethodGPUMemory     map[string]int    // Min GPU memory per tuning method (batch size 1).
```
Fei-Guo, line 24: TuningMethodGPUMemory->TuningPerGPUMemoryRequirement

Fei-Guo: Please also update the comments of other requirements in ths struct indicating they are used for inference.

ishaansehgal99: Resolved

Fei-Guo (approved):

PR #470, opened 2024-06-14
fix: Refactor Naming Conventions, Update Dependencies, Enhance Examples, and Add Volume Validation Check

ishaansehgal99:
**Reason for Change**:
This PR includes a bunch of renaming and chore changes. It includes: 

1. Package renaming to avoid hyphens in package names for phi-2 and phi-3.
2. Updating pip dependencies required for phi-3
3. Use huggingface naming for plugins with hyphens
4. Add phi3 to examples folder
5. Readme: Update phi-3 example requests 
6. Change ConfigMapTemplate back to Config and update the comments
7. Fix a device error encountered by inference API UTs by explicitly specifying device
8. Raise not implemented error in validation for using volume


examples/inference/kaito_workspace_phi-3-adapter.yaml
bangqipropel, line 1: seems like still different with other yaml file names in examples/inference, shall we make them consistent?

Fei-Guo (commented):
Nit: naming consistency for example yaml file. Using underscore instead of hyphen.

Fei-Guo (approved):

PR #476, opened 2024-06-18
feat: Add E2E Tests for Phi-3 and Tuning

ishaansehgal99:
**Reason for Change**:
1. Add E2E tests for Phi-3 and Tuning
2. Uses temporary ACR for uploading e2e results
3. Allows a public preset to be used in private accessmode with private ACR path 
4. Adds Validation check for count > 1 for Tuning


pkg/controllers/workspace_controller.go
```diff
@@ -492,18 +503,20 @@ func (c *WorkspaceReconciler) ensureService(ctx context.Context, wObj *kaitov1al
...
-		err = resources.CreateResource(ctx, serviceObj, c.Client)
-		if err != nil {
+		if err := resources.CreateResource(ctx, serviceObj, c.Client); err != nil {
 			return err
 		}
 		if model.SupportDistributedInference() {
 			headlessService := resources.GenerateHeadlessServiceManifest(ctx, wObj)
-			err = resources.CreateResource(ctx, headlessService, c.Client)
-			if err != nil {
+			if err := resources.CreateResource(ctx, headlessService, c.Client); err != nil {
```
ishaansehgal99, line 519: These code changes ensure workspace service is created for tuning....however we could remove this given tuning has no endpoints...

pkg/controllers/workspace_controller.go
```diff
@@ -471,6 +471,17 @@ func (c *WorkspaceReconciler) ensureNodePlugins(ctx context.Context, wObj *kaito
 	}
 }
 
+// getPresetName returns the preset name from wObj if available
+func getPresetName(wObj *kaitov1alpha1.Workspace) *string {
```
Fei-Guo, line 475: "A string has two values in it: pointer to an array, and the string length. When you pass string as an argument, those two values are copied, not the underlying array." 

It is ok to return string type to simplify code, not much overhead.

pkg/controllers/workspace_controller.go
Fei-Guo: We should skip the entire ensureService for tuning.

api/v1alpha1/workspace_validation.go
```diff
@@ -45,7 +45,7 @@ func (w *Workspace) Validate(ctx context.Context) (errs *apis.FieldError) {
 		errs = errs.Also(w.validateCreate().ViaField("spec"))
 		if w.Inference != nil {
 			// TODO: Add Adapter Spec Validation - Including DataSource Validation for Adapter
-			errs = errs.Also(w.Resource.validateCreate(*w.Inference).ViaField("resource"),
+			errs = errs.Also(w.Resource.validateCreate(w.Inference, w.Tuning).ViaField("resource"),
 				w.Inference.validateCreate().ViaField("inference"))
 		}
 		if w.Tuning != nil {
```
Fei-Guo, line 51: We need to have separate resource check for tuning.

api/v1alpha1/workspace_validation.go
```diff
@@ -299,16 +299,19 @@ func (r *DataDestination) validateUpdate(old *DataDestination) (errs *apis.Field
...
 }
 
-func (r *ResourceSpec) validateCreate(inference InferenceSpec) (errs *apis.FieldError) {
+func (r *ResourceSpec) validateCreate(inference *InferenceSpec, tuning *TuningSpec) (errs *apis.FieldError) {
 	var presetName string
-	if inference.Preset != nil {
+	if inference != nil {
 		presetName = strings.ToLower(string(inference.Preset.Name))
 	}
+	if tuning != nil && *r.Count > 1 {
```
Fei-Guo, line 307: Let us move this to another method for checking tuning resource.

api/v1alpha1/workspace_validation.go
```diff
@@ -45,7 +45,7 @@ func (w *Workspace) Validate(ctx context.Context) (errs *apis.FieldError) {
 		errs = errs.Also(w.validateCreate().ViaField("spec"))
 		if w.Inference != nil {
```
Fei-Guo, line 46: We need to ensure either one of tunning or inference spec should be set, not both.

pkg/inference/preset-inferences.go
```diff
@@ -96,7 +96,8 @@ func updateTorchParamsForDistributedInference(ctx context.Context, kubeClient cl
 
 func GetInferenceImageInfo(ctx context.Context, workspaceObj *kaitov1alpha1.Workspace, presetObj *model.PresetParam) (string, []corev1.LocalObjectReference) {
 	imagePullSecretRefs := []corev1.LocalObjectReference{}
-	if presetObj.ImageAccessMode == string(kaitov1alpha1.ModelImageAccessModePrivate) {
+	if presetObj.ImageAccessMode == string(kaitov1alpha1.ModelImageAccessModePrivate) ||
+		string(workspaceObj.Inference.Preset.AccessMode) == string(kaitov1alpha1.ModelImageAccessModePrivate) {
```
Fei-Guo, line 100: Why do we need the extra check?

test/e2e/preset_test.go
```diff
@@ -109,16 +118,76 @@ func createCustomWorkspaceWithPresetCustomMode(imageName string, numOfNode int)
...
 				MatchLabels: map[string]string{"kaito-workspace": "private-preset-e2e-test-custom"},
-			}, nil, "", utils.InferenceModeCustomTemplate, nil, utils.GeneratePodTemplate(uniqueID, namespaceName, imageName, nil))
+			}, nil, "", utils.CustomTemplateAccessMode, nil, utils.GeneratePodTemplate(uniqueID, namespaceName, imageName, nil))
 
 		createAndValidateWorkspace(workspaceObj)
 	})
 	return workspaceObj
 }
 
+func createPhi3WorkspaceWithPresetPrivateMode(registry, registrySecret, imageVersion string, numOfNode int) *kaitov1alpha1.Workspace {
```
Fei-Guo, line 131: Can we use public image for e2e test?

test/e2e/utils/utils.go
```diff
@@ -36,6 +42,15 @@ func GetEnv(envVar string) string {
 	return env
 }
 
+func GenerateRandomString(n int) string {
```
Fei-Guo, line 45: Maybe make it simpler, just a random number between 0 and 1000 and convert to string.

api/v1alpha1/workspace_validation.go
ishaansehgal99: Added

pkg/controllers/workspace_controller.go
ishaansehgal99: We return a pointer to perform nil check

pkg/inference/preset-inferences.go
ishaansehgal99: This check allows public models to run in private mode.

api/v1alpha1/workspace_validation.go
ishaansehgal99: func (w *Workspace) validateCreate() (errs *apis.FieldError) {
	if w.Inference == nil && w.Tuning == nil {
		errs = errs.Also(apis.ErrGeneric("Either Inference or Tuning must be specified, not neither", ""))
	}
	if w.Inference != nil && w.Tuning != nil {
		errs = errs.Also(apis.ErrGeneric("Either Inference or Tuning must be specified, but not both", ""))
	}
	return errs
}

pkg/controllers/workspace_controller.go
ishaansehgal99: Resolved

presets/models/falcon/model.go
```diff
@@ -75,7 +75,7 @@ func (*falcon7b) GetTuningParameters() *model.PresetParam {
 		ModelFamilyName:           "Falcon",
 		ImageAccessMode:           string(kaitov1alpha1.ModelImageAccessModePublic),
 		DiskStorageRequirement:    "50Gi",
-		GPUCountRequirement:       "2",
+		GPUCountRequirement:       "1",
```
bangqipropel, line 78: Just question, why we change it back to 1?

bangqipropel (approved):
Approved

presets/models/falcon/model.go
ishaansehgal99: We can run falcon tuning with 1 gpu, its just that its slower, but still possible

Fei-Guo (approved):

PR #485, opened 2024-06-27
feat: Add Util funcs, updating func names, logs, configs and ensure service requirement

ishaansehgal99:
**Reason for Change**:
Renaming functions, updating configs and services

### DIFF
--- a/presets/inference/text-generation/requirements.txt
+++ b/presets/inference/text-generation/requirements.txt
@@ -14,7 +14,6 @@ numpy==1.22.4
 bitsandbytes==0.42.0
 
 # Less critical, can be latest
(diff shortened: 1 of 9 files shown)


presets/inference/text-generation/requirements.txt
```diff
@@ -14,7 +14,6 @@ numpy==1.22.4
 bitsandbytes==0.42.0
 
 # Less critical, can be latest
-deepspeed
```
ishaansehgal99, line 17: Including deepspeed package causes breaking version with transformers - to be included in future release

Fei-Guo (approved):

PR #505, opened 2024-07-11
fix: Update timeout values with string

ishaansehgal99:
**Reason for Change**:
Update timeout values with string


Fei-Guo:
Fix is in #506

PR #507, opened 2024-07-11
fix: increase e2e test suite timeout

Fei-Guo:
Since we start to add more and more e2e tests, the 60m suite timeout is too short. Increase the timeout to 3h.


bangqipropel (approved):
LGTM

Fei-Guo:
already addressed by #506

PR #509, opened 2024-07-11
chore: Update nodeClaim manifest and add nodeClass

helayoty:
**Reason for Change**:
- Update nodeClaim manifest:
  - Use `ResourceEphemeralStorage` instead of `ResourceStorage`
  - Add `LabelInstanceTypeStable` to the Node Selector.
- Check if NodeClass for either Azure or AWS cases has been created. if not, the controller will create it automatically.
- Add nodeClaim error handle for plugin error case.
- Update go toolchain to 1.22.5 to be able to import karpenter/aws
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/nodeclaim/nodeclaim.go
```diff
@@ -76,50 +80,54 @@ func GenerateNodeClaimManifest(ctx context.Context, storageRequirement string, w
 		},
 		Spec: v1beta1.NodeClaimSpec{
 			NodeClassRef: &v1beta1.NodeClassReference{
-				Name: "default",
+				Name: NodeClassName,
 				Kind: nodeClassRefKind,
 			},
+			Taints: []v1.Taint{
+				{
+					Key:    utils.SKUString,
```
Fei-Guo, line 88: Can you move SKUString and GPUString to utils/const?

pkg/utils/common.go
```diff
@@ -101,3 +102,7 @@ func GetReleaseNamespace() (string, error) {
 	}
 	return "", fmt.Errorf("failed to determine release namespace from file %s and env var %s", namespaceFilePath, consts.DefaultReleaseNamespaceEnvVar)
 }
+
+func GetCloudProviderName(ctx context.Context) string {
```
Fei-Guo, line 106: ctx is not used anywhere, why do you need to pass it in?

pkg/nodeclaim/nodeclaim.go
```diff
@@ -250,3 +320,18 @@ func CheckNodeClaimStatus(ctx context.Context, nodeClaimObj *v1beta1.NodeClaim,
 		}
 	}
 }
+
+func IsNodeClassAvailable(ctx context.Context, kubeClient client.Client) bool {
```
Fei-Guo, line 324: This client is controller-runtime client. If you don't watch the nodeclaim object, every GET will become a direct API call. I suggest you move this code out of reconciler (to main function, just execute once when controller starts). The nodeclass is not associated with any workspace. It is weird to see it is reconciled whenever we reconcile a workspace.

Fei-Guo (commented):
Can you move all aws related code to another PR (more UT tests may be needed to test AWS  
 enablement) and the PR will be more focused.

Fei-Guo (approved):

PR #524, opened 2024-07-18
feat: Add controllerrevision for workspaceController

bangqipropel:
**Reason for Change**:
Add contropllerrevision for workspace controller to mark the revision
Add unit test for contropllerrevision

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/controllers/workspace_controller.go
```diff
@@ -149,10 +175,87 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+	c.revisionCache = append(c.revisionCache, currentHash)
+	const maxRevisionHistoryLimit = 10
+	if len(c.revisionCache) > maxRevisionHistoryLimit {
+		c.revisionCache = c.revisionCache[1:]
+	}
+	return nil
+}
+
+func computeHash(w *kaitov1alpha1.Workspace) string {
+	hasher := md5.New()
```
ishaansehgal99, line 251: Can use sha256.New() - more collision resistant

pkg/controllers/workspace_controller.go
```diff
@@ -149,10 +175,87 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+		if currentHash != lastCachedHash {
+			needNewRevision = true
+		}
+	}
+	if !needNewRevision {
+		return nil
+	}
+	if currentHash == "" {
+		currentHash = computeHash(wObj)
+	}
```
ishaansehgal99, line 218: Can be replaced with something like

currentHash := computeHash(wObj)
lastHash := wObj.Annotations["controller-revision-hash"]
if lastHash == currentHash { return nil }

pkg/controllers/workspace_controller.go
```diff
@@ -149,10 +175,87 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+		Data:     runtime.RawExtension{Raw: jsonData},
+	}
+	if err := c.Client.Create(ctx, newRevision); err != nil {
+		return fmt.Errorf("failed to create new ControllerRevision: %w", err)
+	}
+	c.revisionCache = append(c.revisionCache, currentHash)
+	const maxRevisionHistoryLimit = 10
+	if len(c.revisionCache) > maxRevisionHistoryLimit {
+		c.revisionCache = c.revisionCache[1:]
+	}
```
ishaansehgal99, line 246: Can be replaced with something that adds the "controller-revision-hash" to the annotations and then updates the object.

pkg/controllers/workspace_controller.go
```diff
@@ -149,10 +175,87 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+	if err := c.deleteControllerRevisions(ctx, wObj); err != nil {
+		klog.ErrorS(err, "failed to delete ControllerRevisions", "workspace", klog.KObj(wObj))
+		return reconcile.Result{}, err
+	}
 	return c.garbageCollectWorkspace(ctx, wObj)
 }
 
+func (c *WorkspaceReconciler) deleteControllerRevisions(ctx context.Context, wObj *kaitov1alpha1.Workspace) error {
+	revisionList := &appsv1.ControllerRevisionList{}
+	if err := c.Client.List(ctx, revisionList, client.InNamespace(wObj.Namespace), client.MatchingLabels{
```
ishaansehgal99, line 187: As fei mentioned you may not need the delete

pkg/controllers/workspace_controller.go
```diff
@@ -155,6 +179,81 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+		return fmt.Errorf("failed to marshal revision data: %w", err)
+	}
+
+	revisions := &appsv1.ControllerRevisionList{}
+	if err := c.List(ctx, revisions, client.InNamespace(wObj.Namespace), client.MatchingLabels{WorkspaceNameLabel: wObj.Name}); err != nil {
+		return fmt.Errorf("failed to list revisions: %w", err)
+	}
+	sort.Slice(revisions.Items, func(i, j int) bool {
+		return revisions.Items[i].Revision > revisions.Items[j].Revision
+	})
```
ishaansehgal99, line 203: Could also add the latest revision number in the annotations...this would prevent need to sort

ishaansehgal99 (approved):

pkg/controllers/workspace_controller.go
```diff
@@ -180,6 +204,81 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+	if err := c.List(ctx, revisions, client.InNamespace(wObj.Namespace), client.MatchingLabels{WorkspaceNameLabel: wObj.Name}); err != nil {
+		return fmt.Errorf("failed to list revisions: %w", err)
+	}
+	sort.Slice(revisions.Items, func(i, j int) bool {
+		return revisions.Items[i].Revision < revisions.Items[j].Revision
+	})
+
+	revisionNum := int64(1)
+
+	if exists {
```
Fei-Guo, line 232: This code is dangerous. You can have a case where the annotation exists, but all controller revisions are deleted by user. Then this code hits a crash because len(revisions.Items) == 0.  You should always check the sorted list and pick the largest revision number from there.

pkg/controllers/workspace_controller.go
```diff
@@ -180,6 +204,81 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
+	wObj.SetAnnotations(annotations)
+	if err := c.Update(ctx, wObj); err != nil {
+		return fmt.Errorf("failed to update Workspace annotations: %w", err)
+	}
+
+	if err := c.Create(ctx, newRevision); err != nil {
+		return fmt.Errorf("failed to create new ControllerRevision: %w", err)
+	}
+
+	const maxRevisionHistoryLimit = 10
```
Fei-Guo, line 264: Can you move this to a global const value? https://github.com/Azure/kaito/blob/main/pkg/utils/consts/consts.go

pkg/utils/test/testUtils.go
```diff
@@ -72,6 +72,114 @@ var (
 	}
 )
 
+var (
+	MockWorkspaceWithDeleteOldCR = v1alpha1.Workspace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "testWorkspace",
+			Namespace: "kaito",
+			Annotations: map[string]string{
+				"workspace.kaito.io/revision": "DeleteOldCR",
```
Fei-Guo, line 81: I don't understand why the value of "workspace.kaito.io/revision" is not a hash value?

pkg/controllers/workspace_controller.go
ishaansehgal99: To add, when you first make controller there would be zero controller revision objects. So this line below will panic - out of bounds.

bangqipropel: > To add, when you first make controller there would be zero controller revision objects. So this line below will panic - out of bounds. tricky bug I didn't spot it at first either :)

First have controller revision? In this case I think it should not, since it will not go into  `if exists`

Fei-Guo (approved):

PR #529, opened 2024-07-19
feat: Add workspace status to present the tuning job status

Fei-Guo:
This PR made a few changes in terms of the workspace status update.
- Since workspace CRD now supports both inference and tuning, the `WorkspaceConditionTypeReady` status is renamed to `WorkspaceConditionTypeSucceeded`, which can be used for both use cases to represent the overall status. 
- Add a `WorkspaceConditionTypeTuningJobStatus` to indicate that the job pod has started without error.
- Since the controller has a watcher for the job, once job has changed its status to "succeeded", the reconciler will update the `WorkspaceConditionTypeSucceeded` status to true.
- The CRD printed columns are also changed accordingly.


pkg/controllers/workspace_controller.go
```diff
@@ -561,6 +582,18 @@ func (c *WorkspaceReconciler) applyTuning(ctx context.Context, wObj *kaitov1alph
...
 
 	if err != nil {
+		if updateErr := c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1alpha1.WorkspaceConditionTypeTuningJobStatus, metav1.ConditionFalse,
+			"WorkspaceTuningJobStatusFailed", err.Error()); updateErr != nil {
+			klog.ErrorS(updateErr, "failed to update workspace status", "workspace", klog.KObj(wObj))
+			return updateErr
+		} else {
+			return err
+		}
+	}
```
ishaansehgal99, line 592:
```suggestion
		}
		return err
	}
```

charts/kaito/workspace/crds/kaito.sh_workspaces.yaml
```diff
@@ -29,8 +29,11 @@ spec:
     - jsonPath: .status.conditions[?(@.type=="InferenceReady")].status
       name: InferenceReady
       type: string
-    - jsonPath: .status.conditions[?(@.type=="WorkspaceReady")].status
-      name: WorkspaceReady
+    - jsonPath: .status.conditions[?(@.type=="JobStarted")].status
+      name: JobStarted
```
ishaansehgal99, line 33: So this job started column will be present even for inference workloads?

ishaansehgal99 (approved):

charts/kaito/workspace/crds/kaito.sh_workspaces.yaml
Fei-Guo: yes

PR #540, opened 2024-07-26
feat: Update deployment object when adapter changes are detected

bangqipropel:
**Reason for Change**:
1. Update deployment object when adapter changes are detected
2. Include logic for comparing adapters 
3. Add unit test for deployment update

### DIFF
--- a/pkg/controllers/workspace_controller.go
+++ b/pkg/controllers/workspace_controller.go
@@ -9,10 +9,12 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"reflect"
 	"sort"
 	"strings"
 	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/azure/kaito/pkg/featuregates"
 	"github.com/azure/kaito/pkg/nodeclaim"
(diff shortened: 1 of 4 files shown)


pkg/controllers/workspace_controller.go
```diff
@@ -278,6 +318,35 @@ func computeHash(w *kaitov1alpha1.Workspace) string {
 	return hex.EncodeToString(hasher.Sum(nil))
 }
 
+func compareAdapters(oldAdapters, newAdapters []kaitov1alpha1.AdapterSpec) bool {
```
Fei-Guo, line 321: Add sufficient UT for this method.

pkg/controllers/workspace_controller.go
```diff
@@ -253,6 +269,30 @@ func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj
...
+	if wObj.Inference != nil {
+		if previousWObj != nil && !compareAdapters(previousWObj.Inference.Adapters, wObj.Inference.Adapters) {
+			deployment := &appsv1.Deployment{}
+			if err := c.Get(ctx, types.NamespacedName{
+				Name:      wObj.Name,
+				Namespace: wObj.Namespace,
+			}, deployment); err != nil {
+				if !errors.IsNotFound(err) {
+					klog.ErrorS(err, "failed to get deployment", "deplyment", wObj.Name)
+				}
```
Fei-Guo, line 281: You can add an annotation with revhash in the deployment, if the revhash is equal to the hash you plan to apply, you can skip the update.

Fei-Guo (commented):
We need to consider how the workspace status may change when we allow to update the inference spec. This can be done in next change.

pkg/controllers/workspace_controller.go
bangqipropel: I can add it, but if the code is going to this position but still find the same revHash, it must have some errors, so I am reporting an error here

Fei-Guo (approved):

PR #541, opened 2024-07-29
fix: Use SKU GPU Count in DeploymentSpec

ishaansehgal99:
**Reason for Change**:
Cases were workspace was not utilizing all available GPUs on machine by default this was causing slower inference requests and unused resources. This is the fix.

### DIFF
--- a/pkg/inference/preset-inferences.go
+++ b/pkg/inference/preset-inferences.go
@@ -150,7 +150,6 @@ func CreatePresetInference(ctx context.Context, workspaceObj *kaitov1alpha1.Work
 	if skuExists {
 		skuNumGPUs = fmt.Sprintf("%d", skuConfig.GPUCount)
 	} else {
-		// Fetch Machine objects with the specified label
 		machineList, err := machine.ListMachinesByWorkspace(context.TODO(), workspaceObj, kubeClient)
 		if err != nil {
 			return nil, fmt.Errorf("Failed to list Machine objects: %v", err)
--- a/pkg/inference/preset-inferences.go
+++ b/pkg/inference/preset-inferences.go
@@ -152,7 +152,7 @@ func CreatePresetInference(ctx context.Context, workspaceObj *kaitov1alpha1.Work
 	} else {
 		nodeNames := workspaceObj.Status.WorkerNodes
 		if len(nodeNames) == 0 {
-			return nil, fmt.Errorf("No worker nodes found in the workspace")
+			return nil, fmt.Errorf("no worker nodes found in the workspace")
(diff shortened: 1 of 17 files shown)


pkg/utils/common.go
```diff
@@ -114,3 +116,36 @@ func GetSKUHandler() (sku.CloudSKUHandler, error) {
...
+			fmt.Printf("Failed to list Node object %s: %v\n", nodeName, err)
+			continue
+		}
+		allNodes.Items = append(allNodes.Items, nodeList.Items...)
+	}
+
+	return GetGPUCountFromNodes(&allNodes), nil
+}
+
+func GetGPUCountFromNodes(nodeList *v1.NodeList) string {
```
Fei-Guo, line 143: GetGPUCountFromNodes -> GetPerNodeGPUCountFromNodes

pkg/inference/preset-inferences.go
```diff
@@ -138,7 +139,25 @@ func CreatePresetInference(ctx context.Context, workspaceObj *kaitov1alpha1.Work
 		volumeMounts = append(volumeMounts, adapterVolumeMount)
 	}
 
-	commands, resourceReq := prepareInferenceParameters(ctx, inferenceObj)
+	skuHandler, err := utils.GetSKUHandler()
+	if err != nil {
```
Fei-Guo, line 143: Can you use a common function for GetSKUGPU? Instead of copying the same code in two places?

pkg/tuning/preset-tuning.go
```diff
@@ -310,7 +310,26 @@ func CreatePresetTuning(ctx context.Context, workspaceObj *kaitov1alpha1.Workspa
...
+	skuNumGPUs := tuningObj.GPUCountRequirement // Default to using tuningObj in case sku information not available
+	skuConfig, skuExists := skuHandler.GetGPUConfigs()[workspaceObj.Resource.InstanceType]
+	if skuExists {
+		skuNumGPUs = fmt.Sprintf("%d", skuConfig.GPUCount)
+	} else {
+		skuGPUCount, err := utils.FetchGPUCountFromNodes(ctx, kubeClient, workspaceObj.Status.WorkerNodes)
+		if err != nil {
+			fmt.Printf("Failed to fetch GPU count from nodes %v", err)
+		} else if skuGPUCount != "" {
+			skuNumGPUs = fmt.Sprintf("%d", skuGPUCount)
```
Fei-Guo, line 328: skuCPUCount is a string already, right?

Fei-Guo (approved):

PR #546, opened 2024-08-02
chore: patch for volumeMounts

bangqipropel:
**Reason for Change**:
add volumeMounts and volume patch for initContainers
add nil check for annotations

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/controllers/workspace_controller.go
+++ b/pkg/controllers/workspace_controller.go
@@ -207,7 +207,7 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
 	return c.garbageCollectWorkspace(ctx, wObj)
 }
 
-func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj *kaitov1alpha1.Workspace) error {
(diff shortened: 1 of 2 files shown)


pkg/controllers/workspace_controller.go
```diff
@@ -287,11 +289,29 @@ func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj
...
+				var volumeMounts []corev1.VolumeMount
+				shmVolume, shmVolumeMount := utils.ConfigSHMVolume(*wObj.Resource.Count)
+				if shmVolume.Name != "" {
+					volumes = append(volumes, shmVolume)
+				}
+				if shmVolumeMount.Name != "" {
+					volumeMounts = append(volumeMounts, shmVolumeMount)
+				}
+
+				if len(wObj.Inference.Adapters) > 0 {
```
Fei-Guo, line 302: Ok, I see you also want to support a case for adapters are being removed from the spec, in this case, the initcontainer sure just be empty.  Can you adjust the code to achieve that?

Fei-Guo (approved):

PR #547, opened 2024-08-02
feat: Add OOM GPU Troubleshooting Section

ishaansehgal99:
**Reason for Change**:
This PR adds OOM GPU Troubleshooting Section for tuning and inference


presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
 ## Distributed inference
 
 For models that support distributed inference, when the node count is larger than one, [torch distributed elastic](https://pytorch.org/docs/stable/distributed.elastic.html) is configured with master/worker pods running in multiple nodes and the service endpoint is the master pod.
+
+## OOM Troubleshooting
+
+### Gathering Debug Information
+1. Set CUDA Environment Variables:
+   - `CUDA_LAUNCH_BLOCKING=1`: This variable provides more detailed debug information
+   - Reduce GPU Memory Fragmentation by setting:
```
Fei-Guo, line 25: Is this about debugging setup? This looks like some defaults that we should use in the pod?

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+
+
+### Preliminary Steps
+If setting environment variables doesn't resolve the issue, consider these steps:
+1. Reduce Batch Size: 
+   - The default batch size is set to the smallest size of 1. If you haven't modified this, you are already using the minimum batch size.
+2. Enable QLoRa (Quantization)
+   - Refer to `../charts/kaito/templates/qlora-params.yaml` for an example. (TODO: example GRaph)
+3. Truncation and Padding
+   - Truncation reduces data sample sizes, which can help prevent OOM errors. For more information, refer to [Hugging Face's documentation on padding and truncation](https://huggingface.co/docs/transformers/en/pad_truncation).
```
Fei-Guo, line 41: You need to add some pro/cons here, what is the problem of truncating the input? How should cx do it? Preprocessing the input data via some scripts?

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+
+There are several ways to go about debugging GPU OOM issues during the fine-tuning process. 
+
+
+### Observing and Adjusting the Dataset
+
+If issues persist, examine your tuning dataset for outliers:
```
Fei-Guo, line 48: For every figure, you should explain the x and y axis. How do you do the experiment and how do you collect the data.

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+- For example see this dataset below ![img.png](dataset_sizes.png)
+- Identify and filter out larger data samples that may be causing OOM issues.
```
Fei-Guo, line 50: How do you remove it in practice? You need add your rational about this recommendation? why large input data will cause OOM?

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+- Consider truncating data samples to manage their size.
+
+### Memory Management
+
+In Kaito, we call torch.cuda.empty_cache() at every tuning step to prevent OOM errors. Here’s an experiment demonstrating the impact of this approach:
+
+1. Memory Usage Over Time:
+   - Running a tuning job for Phi-3 on an A100 GPU shows memory usage patterns. ![mem_usage_no_cache_clear.png](mem_usage_no_cache_clear.png)
```
Fei-Guo, line 58: Again, explain all figures with sufficient details. I suggestion you use a separate figure to show the pattern of input token size (it should not be batch size). It is hard to read all three metrics in one figure.

Fei-Guo: You need to explain the findings from the figure.

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+
+### Observing and Adjusting the Dataset
+
+If issues persist, examine your tuning dataset for outliers:
+- For example see this dataset below ![img.png](dataset_sizes.png)
+- Identify and filter out larger data samples that may be causing OOM issues.
+- Consider truncating data samples to manage their size.
+
+### Memory Management
+
```
Fei-Guo, line 54: You definitely need to explain the reserved memory and allocated memory here and how they will affect the OOM.

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+In Kaito, we call torch.cuda.empty_cache() at every tuning step to prevent OOM errors. Here’s an experiment demonstrating the impact of this approach:
+
+1. Memory Usage Over Time:
+   - Running a tuning job for Phi-3 on an A100 GPU shows memory usage patterns. ![mem_usage_no_cache_clear.png](mem_usage_no_cache_clear.png)
+2. Effect of Clearing Cache: 
+   - Clearing the cache at every step significantly improves memory management. ![mem_usage_cache_clear_1.png](mem_usage_cache_clear_1.png)
+3. Detailed View:
```
Fei-Guo, line 61: You may switch the order and move the truncate the input part here. The detail view can prove the relationship between input data size and the memory consumption. Move the entire the section below the Performance Consideration section.

presets/README.md
Fei-Guo: Are you recommending cx to set these? How can they set it? pros and cons of setting them?

presets/README.md
```diff
@@ -16,3 +16,69 @@ Each preset model has its own hardware requirements in terms of GPU count and GP
...
+3. Detailed View: 
+   - A closer look reveals the correlation between reserved memory and batch content size. ![mem_usage_cache_clear_2.png](mem_usage_cache_clear_2.png)
+
+### Performance Consideration 
+Clearing the cache at every step introduces minimal performance overhead:
+![img_3.png](cache_clearing_performance.png)
+
+Given the minimal performance penalty, it is acceptable to call `torch.cuda.empty_cache()` at every step to manage memory and prevent OOM errors effectively.
+
+
```
Fei-Guo, line 70: You need to add your overall conclusion here: due to the way CUDA to deal with memory reserved, you may have to preserve large GPU to prepare for larger input samples which may burst the GPU memory usage.

presets/README.md
ishaansehgal99: Updated

presets/README.md
ishaansehgal99: Included pros and cons and how to do it with example script

ishaansehgal99:
This all got covered in a separate repo in its - [README.md](https://github.com/ishaansehgal99/GPU-Memory-Experiments/blob/master/README.md )

PR #552, opened 2024-08-08
fix: Keep Docker Sidecar Alive for Data Retrieval

ishaansehgal99:
**Reason for Change**:
- Keep the docker sidecar container alive so incase ACR push fails we can still exec into the container to retrieve completed tuning job files in the /mnt/results folder.

- Perform validation check for output images that are uppercase

- Use projected volume for mounting docker config secrets - mount at directory location /root/.docker/config

- Add secret for E2E Temp ACR

### DIFF
--- a/api/v1alpha1/workspace_validation.go
+++ b/api/v1alpha1/workspace_validation.go
@@ -271,7 +271,19 @@ func (r *DataDestination) validateCreate() (errs *apis.FieldError) {
 		re := regexp.MustCompile(`^(.+/[^:/]+):([^:/]+)$`)
 		if !re.MatchString(r.Image) {
 			errs = errs.Also(apis.ErrInvalidValue("Invalid image format, require full output image URL", "Image"))
+		} else {	// Executes if image is of correct format
+			// Extract repository name (part after the last / and before the colon :)
+			// Suppose r.Image == modelsregistry.azurecr.io/ADAPTER_HERE:0.0.1
+			parts := strings.Split(r.Image, "/")
+			lastPart := parts[len(parts)-1]           // Extracts "ADAPTER_HERE:0.0.1"
+			repoName := strings.Split(lastPart, ":")[0] // Extracts "ADAPTER_HERE"
+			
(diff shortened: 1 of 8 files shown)


api/v1alpha1/workspace_validation.go
```diff
@@ -271,7 +271,19 @@ func (r *DataDestination) validateCreate() (errs *apis.FieldError) {
 		re := regexp.MustCompile(`^(.+/[^:/]+):([^:/]+)$`)
 		if !re.MatchString(r.Image) {
 			errs = errs.Also(apis.ErrInvalidValue("Invalid image format, require full output image URL", "Image"))
+		} else {	// Executes if image is of correct format
```
Fei-Guo, line 274: Should you do the same check for the data source as well? or maybe have a common function due to the duplicated code.

Fei-Guo: And add UT pls

test/e2e/preset_test.go
```diff
@@ -56,6 +56,7 @@ func loadTestEnvVars() {
 	aiModelsRegistry = utils.GetEnv("AI_MODELS_REGISTRY")
 	aiModelsRegistrySecret = utils.GetEnv("AI_MODELS_REGISTRY_SECRET")
 	supportedModelsYamlPath = utils.GetEnv("SUPPORTED_MODELS_YAML_PATH")
+	azureClusterName = utils.GetEnv("AZURE_CLUSTER_NAME")
```
Fei-Guo, line 59: Is this used anywhere?

ishaansehgal99: Yes this is used by fine tuning e2e - outputRegistryUrl. e2e for fine tuning previously was uploading to local ACR now we are using temp ACR

api/v1alpha1/workspace_validation.go
ishaansehgal99: cool added function and added UTs

Fei-Guo (approved):

PR #559, opened 2024-08-14
feat: Update controllerRevision and deployment lifecycle 

bangqipropel:
**Reason for Change**:
1. Change "workspace.kaito.io/revision" to revision number, and make the hash result as "workspace.kaito.io/hash"
2. Sync contropllerrevision and workspace at the beginning of the Reconcile
3. Update the tests for the changes

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/controllers/workspace_controller.go
+++ b/pkg/controllers/workspace_controller.go
@@ -9,8 +9,8 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -52,10 +52,10 @@ import (
 )
 
 const (
-	gpuSkuPrefix                = "Standard_N"
-	nodePluginInstallTimeout    = 60 * time.Second
-	WorkspaceRevisionAnnotation = "workspace.kaito.io/revision"
-	WorkspaceNameLabel          = "workspace.kaito.io/name"
+	gpuSkuPrefix             = "Standard_N"
+	nodePluginInstallTimeout = 60 * time.Second
+	WorkspaceHashAnnotation  = "workspace.kaito.io/hash"
+	WorkspaceNameLabel       = "workspace.kaito.io/name"
 )
 
 type WorkspaceReconciler struct {
@@ -85,6 +85,10 @@ func (c *WorkspaceReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 
 	klog.InfoS("Reconciling", "workspace", req.NamespacedName)
 
+	if err := c.syncControllerRevision(ctx, workspaceObj); err != nil {
+		return reconcile.Result{}, err
+	}
+
 	if err := c.ensureFinalizer(ctx, workspaceObj); err != nil {
 		return reconcile.Result{}, err
 	}
@@ -103,11 +107,6 @@ func (c *WorkspaceReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 	result, err := c.addOrUpdateWorkspace(ctx, workspaceObj)
 	if err != nil {
 		return result, err
-	}
-
-	if err := c.updateControllerRevision(ctx, workspaceObj); err != nil {
-		klog.ErrorS(err, "failed to update ControllerRevision", "workspace", klog.KObj(workspaceObj))
-		return reconcile.Result{}, nil
 	}
 
 	return result, nil
@@ -213,20 +212,14 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
 
 	return c.garbageCollectWorkspace(ctx, wObj)
 }
-
-func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj *kaitov1alpha1.Workspace) error { // TODO: Move non-updateControllerRevision related logic to separate functions
+func (c *WorkspaceReconciler) syncControllerRevision(ctx context.Context, wObj *kaitov1alpha1.Workspace) error {
 	currentHash := computeHash(wObj)
 	annotations := wObj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	} // nil checking.
 
-	latestHash, exists := annotations[WorkspaceRevisionAnnotation]
-	if exists && latestHash == currentHash {
-		return nil
-	}
-
-	jsonData, err := json.Marshal(wObj)
-	if err != nil {
-		return fmt.Errorf("failed to marshal revision data: %w", err)
-	}
+	revisionNum := int64(1)
 
 	revisions := &appsv1.ControllerRevisionList{}
 	if err := c.List(ctx, revisions, client.InNamespace(wObj.Namespace), client.MatchingLabels{WorkspaceNameLabel: wObj.Name}); err != nil {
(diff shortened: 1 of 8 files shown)


pkg/controllers/workspace_controller.go
```diff
@@ -862,6 +867,36 @@ func (c *WorkspaceReconciler) applyInference(ctx context.Context, wObj *kaitov1a
...
+				WorkspaceNameLabel: wObj.Name,
+			},
+			Annotations: map[string]string{
+				WorkspaceRevisionAnnotation: workloadHash,
+			},
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(wObj, kaitov1alpha1.GroupVersion.WithKind("Workspace")),
+			},
+		},
+		Revision: revisionNum,
```
ishaansehgal99, line 890: Just checking - is this always supposed to be hardcoded to 1?

pkg/resources/manifests.go
```diff
@@ -291,9 +295,25 @@ func GenerateDeploymentManifest(ctx context.Context, workspaceObj *kaitov1alpha1
...
+				Type: appsv1.RollingUpdateDeploymentStrategyType,
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxSurge: &intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: 0,
+					},
+					MaxUnavailable: &intstr.IntOrString{
+						Type:   intstr.Int,
+						IntVal: 1,
+					},
```
ishaansehgal99, line 314: Can you explain the MaxSurge 0 and MaxUnavailable: 1? Provide comment. I know you discussed with Fei - trying to catch up

pkg/controllers/workspace_controller.go
```diff
@@ -270,74 +262,55 @@ func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj
...
-				var volumeMounts []corev1.VolumeMount
-				shmVolume, shmVolumeMount := utils.ConfigSHMVolume(*wObj.Resource.Count)
-				if shmVolume.Name != "" {
-					volumes = append(volumes, shmVolume)
-				}
-				if shmVolumeMount.Name != "" {
-					volumeMounts = append(volumeMounts, shmVolumeMount)
+			if len(revisions.Items) > consts.MaxRevisionHistoryLimit {
+				if err := c.Delete(ctx, &revisions.Items[0]); err != nil {
+					return fmt.Errorf("failed to delete old revision: %w", err)
```
ishaansehgal99, line 280: slight nit. Can you change to this? Easier to read


```

if err != nil {
	if errors.IsNotFound(err) {
		if err := c.Create(ctx, newRevision); err != nil {
			return fmt.Errorf("failed to create new ControllerRevision: %w", err)
		}

		// Manage revision history after creation
		if len(revisions.Items) > consts.MaxRevisionHistoryLimit {
			if err := c.Delete(ctx, &revisions.Items[0]); err != nil {
				return fmt.Errorf("failed to delete old revision: %w", err)
			}
		}
	} else {
		return fmt.Errorf("failed to get controller revision: %w", err)
	}
} else {
	// If the ControllerRevision was found, check for annotation conflicts
	if controllerRevision.Annotations[WorkspaceHashAnnotation] != newRevision.Annotations[WorkspaceHashAnnotation] {
		return fmt.Errorf("revision has the same name and namespace but different workspaces")
	}
}
```

pkg/controllers/workspace_controller.go
```diff
@@ -827,16 +800,45 @@ func (c *WorkspaceReconciler) applyInference(ctx context.Context, wObj *kaitov1a
...
+					}
+					initContainers, envs := resources.GenerateInitContainers(wObj, volumeMounts)
+					spec := &deployment.Spec
+					spec.Template.Spec.InitContainers = initContainers
+					spec.Template.Spec.Containers[0].Env = envs
+					spec.Template.Spec.Containers[0].VolumeMounts = volumeMounts
+					deployment.Annotations[kaitov1alpha1.WorkspaceRevisionAnnotation] = revisionStr
+					spec.Template.Spec.Volumes = volumes
+
+					if err := c.Update(ctx, deployment); err != nil {
```
ishaansehgal99, line 831: Tests needed for testing update mounts

ishaansehgal99: e.g. test initial volume mounts are correct, apply update, test updated volume mounts are correct

pkg/controllers/workspace_controller.go
```diff
@@ -213,20 +213,15 @@ func (c *WorkspaceReconciler) deleteWorkspace(ctx context.Context, wObj *kaitov1
...
 	return c.garbageCollectWorkspace(ctx, wObj)
 }
-
-func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj *kaitov1alpha1.Workspace) error { // TODO: Move non-updateControllerRevision related logic to separate functions
+func (c *WorkspaceReconciler) syncControllerRevision(ctx context.Context, wObj *kaitov1alpha1.Workspace) error {
 	currentHash := computeHash(wObj)
 	annotations := wObj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+		annotations[kaitov1alpha1.WorkspaceRevisionAnnotation] = "1"
```
Fei-Guo, line 221: Do you need to assign the revision 1 here? If not, please remove. It is confusing.

pkg/controllers/workspace_controller.go
```diff
@@ -274,75 +261,56 @@ func (c *WorkspaceReconciler) updateControllerRevision(ctx context.Context, wObj
...
+					return fmt.Errorf("failed to delete old revision: %w", err)
 				}
 			}
-
+		} else {
+			return fmt.Errorf("failed to get controller revision: %w", err)
+		}
+	} else {
+		if controllerRevision.Annotations[WorkspaceHashAnnotation] != newRevision.Annotations[WorkspaceHashAnnotation] {
+			return fmt.Errorf("revision has the same name and namespace but different workspaces")
```
Fei-Guo, line 288: "revision has the same name and namespace but different workspaces" -> "revision name conflicts, the hash values are different"

pkg/controllers/workspace_controller.go
```diff
@@ -831,16 +770,52 @@ func (c *WorkspaceReconciler) applyInference(ctx context.Context, wObj *kaitov1a
 				existingObj = &appsv1.Deployment{}
 
 			}
-
+			revisionStr := wObj.Annotations[kaitov1alpha1.WorkspaceRevisionAnnotation]
 			if err = resources.GetResource(ctx, wObj.Name, wObj.Namespace, c.Client, existingObj); err == nil {
 				klog.InfoS("An inference workload already exists for workspace", "workspace", klog.KObj(wObj))
+				switch deployment := existingObj.(type) {
+				case *appsv1.Deployment:
```
Fei-Guo, line 777: You can use the condition in line 767 instead of casting the type. 

You may also remove the comment in line 764.

Fei-Guo: And we discussed that you only do deployment update when workspace revision number is different from deployment revision number, right?

bangqipropel: > And we discussed that you only do deployment update when workspace revision number is different from deployment revision number, right?

Yes, the comparison is on line 793

Fei-Guo (approved):

PR #567, opened 2024-08-16
feat: Add Adapter Loading Test for E2E Image Preset Workflow

ishaansehgal99:
**Reason for Change**:
This PR adds a test for loading a mental health adapter onto the falcon-7b model. And testing to make sure the adapter was loaded onto the model successfully using `inference_api.py` logs.


bangqipropel (approved):
LGTM

PR #570, opened 2024-08-19
feat: Update karpenter nodeclass

helayoty:
**Reason for Change**:
- Enable workspace controller to manage karpenter providers objects to be able to create the NodeClass.
- Remove the node affinity condition that applies on gpu-provisioner and will not allow Karpenter. 
- add scheme registration for Karpenter CRDs
- cleanup the NodeClass code.
- Add unit tests for NodeClass.
- Update the suite parameter to `node_provisioner`

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


cmd/main.go
```diff
@@ -154,31 +161,22 @@ func main() {
...
 		time.Sleep(2 * time.Second)
 
-		if err = featuregates.ParseAndValidateFeatureGates(featureGates); err != nil {
+		if err := featuregates.ParseAndValidateFeatureGates(featureGates); err != nil {
 			klog.ErrorS(err, "unable to set `feature-gates` flag")
 			exitWithErrorFunc()
 		}
 	}
 
+	err = nodeclaim.CheckNodeClass(ctx, kClient)
```
Fei-Guo, line 170: Why do you remove the feature gate check here?

helayoty: Because it is redundant in this case. The nodeclaim won't get triggered unless the feature gate flag is on

helayoty: Fixed.

Fei-Guo (approved):

PR #591, opened 2024-09-09
feat: set dataset_text_field for more than one column

bangqipropel:
**Reason for Change**:

In the previous, if the column is larger than one, it will report an error: "Unable to load dataset". This is for fixing this case.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/tuning/text-generation/dataset.py
+++ b/presets/tuning/text-generation/dataset.py
@@ -43,6 +43,9 @@ class DatasetManager:
         try:
             self.dataset = load_dataset(file_ext, data_files=dataset_path, split="train")
             print(f"Dataset loaded successfully from {dataset_path} with file type '{file_ext}'.")
+
+            self.analyze_dataset_structure()
+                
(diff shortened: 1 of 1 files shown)


presets/tuning/text-generation/dataset.py
```diff
@@ -56,6 +59,22 @@ def find_valid_dataset(self, data_dir):
...
+        if self.dataset is None:
+            raise ValueError("Dataset is not loaded. Please load the dataset first.")
+        
+        columns = self.dataset.column_names
+
+        if len(columns) > 1:
+            self.dataset_text_field = self.find_text_field(columns)
+    
+    def find_text_field(self, columns):
+        text_words = ["text", "content", "input", "question"]
```
zhuangqh, line 73: what about
```python
text_words = {"text", "content", "input", "question"}
for column in columns:
  if column.lower() in text_words:
    return column
```

PR #628, opened 2024-10-14
feat: Part 1 - Add RAG Embedding Interface

ishaansehgal99:
**Reason for Change**:
This series of PR will integrate llamaindex RAG service for Kaito. 

This first PR introduces the Embedding Model Interface and two classes that implement it `huggingface_local` and `huggingface_remote`. These allows users to specify both local and remote HuggingFace models for text embeddings.

### DIFF
--- a/ragengine/embedding/huggingface_local.py
+++ b/ragengine/embedding/huggingface_local.py
@@ -1,0 +1,19 @@
+from llama_index.embeddings.huggingface import HuggingFaceEmbedding
+
+from .base import BaseEmbeddingModel
+
+
+class LocalHuggingFaceEmbedding(BaseEmbeddingModel):
+    def __init__(self, model_name: str):
+        self.model = HuggingFaceEmbedding(model_name=model_name) # TODO: Ensure/test loads on GPU (when available)
+
+    def get_text_embedding(self, text: str):
+        """Returns the text embedding for a given input string."""
+        return self.model.get_text_embedding(text)
+
+    def get_embedding_dimension(self) -> int:
+        """Infers the embedding dimension by making a local call to get the embedding of a dummy text."""
+        dummy_input = "This is a dummy sentence."
+        embedding = self.get_text_embedding(dummy_input)
+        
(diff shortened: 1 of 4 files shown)


ragengine/embedding/huggingface_local.py
```diff
@@ -0,0 +1,19 @@
...
+    def __init__(self, model_name: str):
+        self.model = HuggingFaceEmbedding(model_name=model_name) # TODO: Ensure/test loads on GPU (when available)
+
+    def get_text_embedding(self, text: str):
+        """Returns the text embedding for a given input string."""
+        return self.model.get_text_embedding(text)
+
+    def get_embedding_dimension(self) -> int:
+        """Infers the embedding dimension by making a local call to get the embedding of a dummy text."""
+        dummy_input = "This is a dummy sentence."
```
bangqipropel, line 16: I know this is for test functionality, but can we add the specific idea here? like "Test model functionality and output dimensions."?

ragengine/embedding/huggingface_remote.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, model_name: str, api_key: str):
+        self.model = HuggingFaceInferenceAPIEmbedding(model_name=model_name, token=api_key)
+
+    def get_text_embedding(self, text: str):
+        """Returns the text embedding for a given input string."""
+        return self.model.get_text_embedding(text)
+    
+    def get_embedding_dimension(self) -> int:
+        """Infers the embedding dimension by making a remote call to get the embedding of a dummy text."""
+        dummy_input = "This is a dummy sentence."
```
bangqipropel, line 17: same question for here

bangqipropel (approved):

ragengine/embedding/huggingface_local.py
ishaansehgal99: Actually, this method isn't for testing, but to initialize objects with the correct embedding dimension, which is why I think the current comment works fine

PR #630, opened 2024-10-14
feat: Part 2 - Add custom LLM inference class

ishaansehgal99:
**Reason for Change**:
This series of PR will integrate llamaindex RAG service for Kaito.

This PR contains the custom LLM inference class for llamaindex. We need this class because we use custom HTTP endpoint or OpenAI API for handling LLM requests so we need a custom LLM inference class.

https://docs.llamaindex.ai/en/stable/module_guides/models/llms/usage_custom/#example-using-a-custom-llm-model-advanced

### DIFF
--- a/ragengine/config.py
+++ b/ragengine/config.py
@@ -1,6 +1,6 @@
 # config.py
 
-# Variables are set via environment variables, typically from the RAGEngine CR
+# Variables are set via environment variables from the RAGEngine CR
 # and exposed to the pod. For example, InferenceURL is specified in the CR and 
 # passed to the pod via env variables.
 
--- a/ragengine/config.py
+++ b/ragengine/config.py
@@ -1,0 +1,15 @@
+# config.py
+import os
+
+EMBEDDING_TYPE = os.getenv("EMBEDDING_TYPE", "local")
+EMBEDDING_URL = os.getenv("EMBEDDING_URL")
+
+INFERENCE_URL = os.getenv("INFERENCE_URL", "http://52.190.41.209/chat")
+INFERENCE_ACCESS_SECRET = os.getenv("AccessSecret", "default-inference-secret")
+# RESPONSE_FIELD = os.getenv("RESPONSE_FIELD", "result")
+
+MODEL_ID = os.getenv("MODEL_ID", "BAAI/bge-small-en-v1.5")
+VECTOR_DB_TYPE = os.getenv("VECTOR_DB_TYPE", "faiss")
+INDEX_SERVICE_NAME = os.getenv("INDEX_SERVICE_NAME", "default-index-service")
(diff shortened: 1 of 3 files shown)


bangqipropel (approved):

ragengine/config.py
```diff
@@ -0,0 +1,15 @@
+# config.py
+import os
+
+EMBEDDING_TYPE = os.getenv("EMBEDDING_TYPE", "local")
+EMBEDDING_URL = os.getenv("EMBEDDING_URL")
+
+INFERENCE_URL = os.getenv("INFERENCE_URL", "http://52.190.41.209/chat")
```
zhuangqh, line 7: why hardcode an ip here?

ragengine/config.py
```diff
@@ -0,0 +1,15 @@
...
+INFERENCE_ACCESS_SECRET = os.getenv("AccessSecret", "default-inference-secret")
+# RESPONSE_FIELD = os.getenv("RESPONSE_FIELD", "result")
+
+MODEL_ID = os.getenv("MODEL_ID", "BAAI/bge-small-en-v1.5")
+VECTOR_DB_TYPE = os.getenv("VECTOR_DB_TYPE", "faiss")
+INDEX_SERVICE_NAME = os.getenv("INDEX_SERVICE_NAME", "default-index-service")
+ACCESS_SECRET = os.getenv("ACCESS_SECRET", "default-access-secret")
+PERSIST_DIR = "./storage"
```
zhuangqh, line 15: can we use an absolute path here? It should start with the `/workspace` prefix

ragengine/config.py
ishaansehgal99: This IP here is a dummy default incase `INFERENCE_URL` environment variable is not found

ragengine/config.py
ishaansehgal99: This is the dir where results are saved. Absolute path would require knowing pods filesystem. I think this is ok for now

ragengine/config.py
ishaansehgal99: I'll make this more clear

Fei-Guo: If it is a dummy, just use "localhost" instead of ip.

ragengine/inference/custom_inference.py
```diff
@@ -0,0 +1,53 @@
+from typing import Any, Optional
+from llama_index.core.llms import CustomLLM, CompletionResponse, LLMMetadata, CompletionResponseGen
```
Fei-Guo, line 2: Can we rename the file name and class name to be just "Inference"?

ragengine/config.py
Fei-Guo: Add a comment here mentioning that InferenceURL is specified in RAGEngine CR and expose to pod via pod env.

Fei-Guo (approved):

PR #646, opened 2024-10-23
feat: Add delete and finalizer to RAGEngine

bangqipropel:
**Reason for Change**:
Add delete and finalizer to RAGEngine

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


ishaansehgal99 (approved):
LGTM

pkg/controllers/ragengine_gc_finalizer.go
```diff
@@ -0,0 +1,64 @@
...
+	staleWObj := ragEngineObj.DeepCopy()
+	staleWObj.SetFinalizers(nil)
+	if updateErr := c.Update(ctx, staleWObj, &client.UpdateOptions{}); updateErr != nil {
+		klog.ErrorS(updateErr, "failed to remove the finalizer from the ragengine",
+			"ragengine", klog.KObj(ragEngineObj), "ragengine", klog.KObj(staleWObj))
+		return ctrl.Result{}, updateErr
+	}
+	klog.InfoS("successfully removed the ragengine finalizers",
+		"ragengine", klog.KObj(ragEngineObj))
+	controllerutil.RemoveFinalizer(ragEngineObj, consts.RAGEngineFinalizer)
```
zhuangqh, line 62: why remove the finalizer again here?

pkg/controllers/ragengine_gc_finalizer.go
```diff
@@ -0,0 +1,64 @@
...
+		for i := range ncList.Items {
+			if deleteErr := c.Delete(ctx, &ncList.Items[i], &client.DeleteOptions{}); deleteErr != nil {
+				klog.ErrorS(deleteErr, "failed to delete the nodeClaim", "nodeClaim", klog.KObj(&ncList.Items[i]))
+				return ctrl.Result{}, deleteErr
+			}
+		}
+	}
+
+	staleWObj := ragEngineObj.DeepCopy()
+	staleWObj.SetFinalizers(nil)
```
zhuangqh, line 54: we should only remove the finalizer which is set by this controller, or we may remove the finalizer set by user by accident.

pkg/controllers/ragengine_gc_finalizer.go
bangqipropel: Why we say "again" here?

pkg/controllers/ragengine_gc_finalizer.go
bangqipropel: I think from 53-62 all about removing the finalizer?

zhuangqh: Suppose finalizers is ["kaito/a", "from-user/b"], just remove the finalizer set by this controller (kaito/b), and set finalizers to ["from-user/b"].
you can take a look at this code https://github.com/openkruise/kruise/blob/master/pkg/controller/ephemeraljob/ephemeraljob_utils.go#L256

pkg/controllers/ragengine_gc_finalizer.go
zhuangqh: you have made an update to the apisever using `staleWObj`.
and update to `ragEngineObj` again

zhuangqh (approved):

Fei-Guo (approved):

PR #652, opened 2024-10-23
feat: Use PyEnv for UTs in GitHub Runner

ishaansehgal99:
**Reason for Change**:
Isolate python dependency environments between tests in the GitHub Runner.


ishaansehgal99:
Addressed in https://github.com/Azure/kaito/pull/624

PR #680, opened 2024-11-07
feat: Dockerfile for Kaito RAG Service

ishaansehgal99:
**Reason for Change**:
Dockerfile for Kaito RAG Service


bangqipropel:
It has not set python evorinment I think.

Fei-Guo (approved):

PR #682, opened 2024-11-08
feat: RAG engine controller revision

bangqipropel:
**Reason for Change**:
RAG engine controller revision

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/ragengine/controllers/ragengine_controller.go
+++ b/pkg/ragengine/controllers/ragengine_controller.go
@@ -5,7 +5,12 @@ package controllers
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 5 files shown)


pkg/ragengine/controllers/ragengine_controller.go
```diff
@@ -39,6 +46,13 @@ import (
 	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 )
 
+const (
+	nodePluginInstallTimeout = 60 * time.Second
```
Fei-Guo, line 50: Why do we need this const in this PR?

bangqipropel: delete both for rag engine and workspace

Fei-Guo (approved):

PR #702, opened 2024-11-18
Remove instanceType field in aks arc doc and rename HCI to Azure Local

qfai:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


ishaansehgal99 (approved):

ishaansehgal99 (changes requested):
one nit

docs/How-to-use-kaito-in-aks-arc.md
```diff
@@ -115,7 +115,6 @@ kind: Workspace
 metadata:
   name: workspace-falcon-7b
 resource:
-  instanceType: Standard_NC12s_v3
```
ishaansehgal99, line 118: If the preferred node cannot be found, a fallback `instanceType` is used, defaulting to `Standard_NC12s_V3`. We retain this field for transparency.

Fei-Guo: I think a default sku can be added once gpuprovisioner supports arc RP apis. For now, it is ok to remove it.

Fei-Guo (approved):

PR #705, opened 2024-11-19
fix: binary search for best context length avoiding oom

zhuangqh:
**Reason for Change**:

fix: binary search for best context length avoiding oom

**Issue Fixed**:

`find_max_available_seq_len` runs to oom when running
on the V100 16GB gpu with 128K context.

**Notes for Reviewers**:

In the worst case, it costs about 1minutes to find the best 
length (running with phi3 medium model and 128k search space).

We set the context length to a safe value to avoid oom.
If the serving server receives a request which token length is longer
than `max_model_len`, server will reject this request.

example error message: `This model's maximum context length is 2 tokens. 
However, you requested 19 tokens (9 in the messages, 10 in the completion).
Please reduce the length of the messages or completion.`

### DIFF
--- a/presets/inference/vllm/inference_api.py
+++ b/presets/inference/vllm/inference_api.py
@@ -3,6 +3,7 @@
 import logging
 import gc
 import os
+from typing import Callable
 
 import uvloop
 import torch
@@ -62,21 +63,14 @@ def find_max_available_seq_len(engine_config: EngineConfig) -> int:
         observability_config=engine_config.observability_config,
     )
 
-    # binary search for the max safe seq len
+    max_probe_steps = 6
+    if os.getenv("MAX_PROBE_STEPS") is not None:
+        max_probe_steps = int(os.getenv("MAX_PROBE_STEPS"))
+
     model_max_blocks = int(engine_config.model_config.max_model_len / engine_config.cache_config.block_size)
-    low = 0
-    high = model_max_blocks * 2
-    while low < high:
-        mid = (low + high + 1) // 2
-        if mid > model_max_blocks:
-            break
+    res = binary_search_with_limited_steps(model_max_blocks, max_probe_steps, lambda x: is_context_length_safe(executor, x))
 
-        if is_context_length_safe(executor, mid):
-            low = mid
-        else:
-            high = mid - 1
-
-    if low <= 0:
+    if res <= 0:
         raise ValueError("No available memory for the cache blocks.")
 
     # release memory
@@ -84,7 +78,25 @@ def find_max_available_seq_len(engine_config: EngineConfig) -> int:
     gc.collect()
     torch.cuda.empty_cache()
 
-    return engine_config.cache_config.block_size * low
+    return engine_config.cache_config.block_size * res
+
+def binary_search_with_limited_steps(upper: int, max_probe_steps: int, is_valid_fn: Callable[[int], bool]) -> int:
+    probe_steps = 0
+    low = 0
+    high = upper * 2
+    while low < high and probe_steps <= max_probe_steps:
+        mid = (low + high + 1) // 2
+        if mid > upper:
+            break
+
+        if is_valid_fn(mid):
+            low = mid
+        else:
+            high = mid - 1
+
+        probe_steps += 1
+
+    return low
 
 def is_context_length_safe(executor: ExecutorBase, num_gpu_blocks: int) -> bool:
     """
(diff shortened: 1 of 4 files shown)


presets/inference/vllm/inference_api.py
```diff
@@ -58,15 +62,44 @@ def find_max_available_seq_len(engine_config: EngineConfig) -> int:
...
     )
 
-    # see https://github.com/vllm-project/vllm/blob/v0.6.3/vllm/engine/llm_engine.py#L477
-    num_gpu_blocks, _ = executor.determine_num_available_blocks()
+    # binary search for the max safe seq len
+    model_max_blocks = int(engine_config.model_config.max_model_len / engine_config.cache_config.block_size)
+    low = 0
+    high = model_max_blocks * 2
+    while low < high:
+        mid = (low + high + 1) // 2
```
zhuangqh, line 70: try the possible largest context window at first.

ishaansehgal99 (approved):

presets/inference/vllm/inference_api.py
```diff
@@ -58,15 +62,44 @@ def find_max_available_seq_len(engine_config: EngineConfig) -> int:
...
+    Check if the avilable gpu blocks is enough for the given num_gpu_blocks.
+    """
+    context_length = executor.cache_config.block_size * num_gpu_blocks
+    executor.scheduler_config.max_num_batched_tokens = context_length
+
+    try:
+        # see https://github.com/vllm-project/vllm/blob/v0.6.3/vllm/engine/llm_engine.py#L477
+        available_gpu_blocks, _ = executor.determine_num_available_blocks()
+    except torch.OutOfMemoryError as e:
+        return False
```
ishaansehgal99, line 100: Consider adding generic exception handling incase error thrown is not torch.OutOfMemoryError

zhuangqh: this is an unexpected exception, just let it go.

presets/inference/vllm/tests/test_vllm_inference_api.py
```diff
@@ -111,4 +113,35 @@ def test_chat_completions_api(setup_server):
...
         assert "content" in choice["message"], "Each message should contain a 'content' key"
-        assert len(choice["message"]["content"]) > 0, "The completion text should not be empty"
\ No newline at end of file
+        assert len(choice["message"]["content"]) > 0, "The completion text should not be empty"
+
+
+def test_binary_search_with_limited_steps():
+
+    def is_safe_fn(x):
+        return x <= 10
```
ishaansehgal99, line 122: add 10 as const

zhuangqh: This is a one-time test data. 😊

presets/inference/vllm/inference_api.py
```diff
@@ -58,15 +63,55 @@ def find_max_available_seq_len(engine_config: EngineConfig) -> int:
...
 
     # release memory
     del executor
     gc.collect()
     torch.cuda.empty_cache()
 
-    return engine_config.cache_config.block_size * num_gpu_blocks
+    return engine_config.cache_config.block_size * res
+
+def binary_search_with_limited_steps(upper: int, max_probe_steps: int, is_valid_fn: Callable[[int], bool]) -> int:
```
ishaansehgal99, line 83:
```
def find_max_valid_context_length(max_context_length: int, max_probe_steps: int, is_valid_fn: Callable[[int], bool]) -> int:
    """
    Finds the maximum valid context length for inference without causing an out-of-memory (OOM) error.
    
    Parameters:
    - max_context_length (int): The initial upper bound for the context length to be tested.
    - max_probe_steps (int): Maximum number of steps to probe for a valid context length.
    - is_valid_fn (Callable[[int], bool]): A function that checks if a given context length is valid.
    
    Returns:
    - int: The maximum valid context length that does not cause an OOM error.
    """
    low, high, probe_steps = 0, max_context_length * 2, 0
    
    while low < high and probe_steps < max_probe_steps:
        mid = (low + high + 1) // 2
        if mid > max_context_length:
            break

        if is_valid_fn(mid):
            low = mid  # Move low pointer up if valid
        else:
            high = mid - 1  # Move high pointer down if not valid

        probe_steps += 1

    return low
```

ishaansehgal99 (approved):

Fei-Guo (approved):

presets/inference/vllm/inference_api.py
zhuangqh: your suggestion is good. but I don't want to introduce any context info about the invoker for this function. It's just a pure helper function. The parameters passed in are well-documented.
We can add a comment about this algorithm.

ishaansehgal99 (approved):

PR #709, opened 2024-11-20
fix: secret update patch

bangqipropel:
**Reason for Change**:
Workspace Secret Update Patch

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/workspace/inference/preset-inferences.go
```diff
@@ -99,6 +99,13 @@ func updateTorchParamsForDistributedInference(ctx context.Context, kubeClient cl
 func GetInferenceImageInfo(ctx context.Context, workspaceObj *kaitov1alpha1.Workspace, presetObj *model.PresetParam) (string, []corev1.LocalObjectReference) {
 	imagePullSecretRefs := []corev1.LocalObjectReference{}
 	// Check if the workspace preset's access mode is private
+	if len(workspaceObj.Inference.Adapters) > 0 {
+		for _, adapter := range workspaceObj.Inference.Adapters {
+			for _, secretName := range adapter.Source.ImagePullSecrets {
+				imagePullSecretRefs = append(imagePullSecretRefs, corev1.LocalObjectReference{Name: secretName})
```
zhuangqh, line 105: I am curious about a situation. if the secret provided by adapter is duplicate with the inference one. What will happen here?
That's to say there are some duplicated entries in `imagePullSecretRefs` array.

bangqipropel: I think it will handle the duplicate ones, but the way they consume it will be try all the secrets one by one.
This one lists the codes about this:
[https://stackoverflow.com/questions/67073039/multiple-imagepullsecrets-in-a-kubernetes-deployment-for-the-same-url](url)

zhuangqh (approved):

Fei-Guo (approved):

PR #710, opened 2024-11-20
feat: [SKU modularization] AWS chart changes

smritidahal653:
**Reason for Change**:

Updating kaito charts for aws integration:

- add clusterrole permission for ec2nodeclasses resource
- add new env variable in deployment CLUSTER_NAME as this environment is needed to create EC2 Node Class
- update nodeSelectorTerm for nvidia-device-plugin daemonset based on the cloud provider
- add AWS specific nodeSelectorRequirement for nodeClaim obj
- add InstanceStorePolicyRAID0 to EC2 node class to share node's ephermeral storage among pods that request it. Additional information [here](https://karpenter.sh/v1.0/concepts/nodeclasses/#specinstancestorepolicy)
- fixing a bug on nodeClaim create (kubeClient.Create) - if the nodeClaim fails to get created, we retry to create it again. However, on kubeClient.Create, certain metadata are set like resourceVersion which will be passed along in the retires if we don't pass the DeepCopy(). This metadata being populated will throw an error from KubeClient `resourceVersion should not be set on objects to be created`

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/utils/nodeclaim/nodeclaim_test.go
```diff
@@ -205,7 +205,7 @@ func TestGenerateNodeClaimManifest(t *testing.T) {
 		assert.Equal(t, nodeClaim.Labels[kaitov1alpha1.LabelWorkspaceNamespace], mockWorkspace.Namespace, "label must have same workspace namespace as workspace")
 		assert.Equal(t, nodeClaim.Labels[consts.LabelNodePool], consts.KaitoNodePoolName, "label must have same labels as workspace label selector")
 		assert.Equal(t, nodeClaim.Annotations[v1beta1.DoNotDisruptAnnotationKey], "true", "label must have do not disrupt annotation")
-		assert.Equal(t, len(nodeClaim.Spec.Requirements), 3, " NodeClaim must have 3 NodeSelector Requirements")
+		assert.Equal(t, len(nodeClaim.Spec.Requirements), 4, " NodeClaim must have 3 NodeSelector Requirements")
```
bangqipropel, line 208: I think if you change the previous to 4, I think the sentence should be " NodeClaim must have 3 NodeSelector Requirements".

Which one is the requirement you added?

smritidahal653: Fixed. And the requirement added was karpenter.k8s.aws/instance-gpu-count should be greater than 0 for aws

Fei-Guo (approved):

charts/kaito/workspace/templates/deployment.yaml
```diff
@@ -47,6 +47,8 @@ spec:
               value: {{ .Values.presetRegistryName }}
             - name: CLOUD_PROVIDER
               value: {{ .Values.cloudProviderName }}
+            - name: CLUSTER_NAME
+              value: {{ .Values.clusterName }}
```
ishaansehgal99, line 51: consider 
value: {{ .Values.clusterName | default "default-cluster-name" }}

ishaansehgal99 (approved):

zhuangqh (approved):

charts/kaito/workspace/templates/deployment.yaml
smritidahal653: added default to values.yaml

ishaansehgal99 (approved):

PR #745, opened 2024-11-29
Create the workspace service for custom models

### LINKED ISSUE #744: Custom Model Integration Doesn't Create a Workspace Service
**Describe the bug**

When following the steps at https://github.com/kaito-project/kaito/blob/main/docs/custom-model-integration/custom-model-integration-guide.md, I noticed that the workspace that gets created doesn't have the kubernetes service that gets created for the preset models. 

This may be by design, but if not it should be a simple fix around [here](https://github.com/kaito-project/kaito/blob/main/pkg/workspace/controllers/workspace_controller.go#L614).

**Steps To Reproduce**

Follow the guide at https://github.com/kaito-project/kaito/blob/main/docs/custom-model-integration/custom-model-integration-guide.md.

**Expected behavior**

Workspace creates a kubernetes service.

**Logs**

N/A

**Environment**

N/A

- Kubernetes version (use `kubectl version`): 1.30
- OS (e.g: `cat /etc/os-release`): N/A
- Install tools: N/A
- Others:  N/A

**Additional context**

zawachte:
**Reason for Change**:

This change updates the `ensureService` function in the workspace controller to create the workspace service for custom models instead of only doing it for preset models. 

The logic today only goes into the service creation code if there is a preset, but it is only using the preset information to see if the model supports distributed inference. To add support for custom models, we can just default the distributed inference support to false and always create the service preset or not. 

It seems fair to me to only support distributed inference on the preset models as that requires more setup. With this change custom models can at least have basic support without any manual Kubernetes resource creation.

Fixes #744.

**Requirements**

- [X] added unit tests and e2e tests (if applicable).

**Issue Fixed**:

Fixes #744 

**Notes for Reviewers**:

### DIFF
--- a/pkg/utils/test/testUtils.go
+++ b/pkg/utils/test/testUtils.go
@@ -70,6 +70,33 @@ var (
 			},
 		},
 	}
+	MockWorkspaceCustomModel = &v1alpha1.Workspace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "testCustomWorkspace",
+			Namespace: "kaito",
+		},
+		Resource: v1alpha1.ResourceSpec{
+			Count:        &gpuNodeCount,
+			InstanceType: "Standard_NC12s_v3",
+			LabelSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"apps": "test",
+				},
+			},
+		},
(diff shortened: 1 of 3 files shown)


zhuangqh (approved):
Thanks for your contribution!

PR #769, opened 2024-12-10
fix: machine and nodeclaim can not supported at the same time

rambohe-ch:
**Reason for Change**:

We need to select Machine or NodeClaim to connect gpu-provisioner component to provide GPU resource.
this means Machine and NodeClaim can not supported at the same time.

In this pull request, improve some sequence process of Machine and NodeClaim into `if condition`.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


ishaansehgal99 (approved):
Cool, is thi

Fei-Guo (approved):

PR #791, opened 2024-12-20
feat: add inference config api

zhuangqh:
**Reason for Change**:

- API change: add config to `workspace.inference.config`
- generate a default config if no user config specified by copying from release-namespace

**Requirements**

### DIFF
--- a/pkg/workspace/inference/preset-inferences.go
+++ b/pkg/workspace/inference/preset-inferences.go
@@ -17,6 +17,7 @@ import (
 	"github.com/kaito-project/kaito/pkg/utils/resources"
 	"github.com/kaito-project/kaito/pkg/workspace/manifests"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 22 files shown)


pkg/workspace/inference/preset-inferences.go
```diff
@@ -191,3 +203,70 @@ func CreatePresetInference(ctx context.Context, workspaceObj *kaitov1alpha1.Work
 	}
 	return depObj, nil
 }
+
```
Fei-Guo, line 206: Is there any chance to dedup this method code with the existing code in fine tuning for the similar funcitonality?

Fei-Guo (commented):
where is the default inference configmap?

zhuangqh:
> where is the default inference configmap?

https://github.com/kaito-project/kaito/blob/main/charts/kaito/workspace/templates/inference-params.yaml  here.

pkg/workspace/inference/preset-inferences.go
zhuangqh: sure

Fei-Guo (approved):

PR #795, opened 2024-12-27
test: add adapter e2e test for vllm

zhuangqh:
validate loaded model by using model list api.
https://docs.vllm.ai/en/latest/usage/lora.html#lora-model-lineage-in-model-card

### DIFF
--- a/test/e2e/inference_with_adapters_test.go
+++ b/test/e2e/inference_with_adapters_test.go
@@ -4,6 +4,7 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


test/e2e/inference_with_adapters_test.go
```diff
@@ -158,6 +159,47 @@ func validateAdapterAdded(workspaceObj *kaitov1alpha1.Workspace, deploymentName
 	})
 }
 
+func validateAdapterLoadedInVLLM(workspaceObj *kaitov1alpha1.Workspace, deploymentName string, adapterName string) {
+	execOption := corev1.PodExecOptions{
+		Command:   []string{"bash", "-c", "apt-get update && apt-get install curl -y; curl -s 127.0.0.1:5000/v1/models | grep " + adapterName},
```
Fei-Guo, line 164: "curl -s 127.0.0.1:5000/v1/models"

What does this return?

zhuangqh:
```
{
   "object":"list",
   "data":[
      {
         "id":"falcon-7b",
         "object":"model",
         "created":1735276136,
         "owned_by":"vllm",
         "root":"falcon-7b",
         "parent":"None",
         "max_model_len":1024
      },
      {
         "id":"e2e-adapter2",
         "object":"model",
         "created":1735276136,
         "owned_by":"vllm",
         "root":"/xxxxx",
         "parent":"falcon-7b",
         "max_model_len":"None"
      }
   ]
}
```

like this. the adapter's parent is the original model.

ishaansehgal99 (approved):

Fei-Guo (approved):

PR #820, opened 2025-01-12
fix: Ensure `model` provided in vLLM inference

ishaansehgal99:
**Reason for Change**:
Update to include model name in requests made to VLLM (has a /v1/models endpoint)


zhuangqh:
remember to update api spec json and docs here https://github.com/kaito-project/kaito/blob/main/docs/inference/README.md#vllm-inference-api

presets/ragengine/inference/inference.py
```diff
@@ -38,29 +41,89 @@ def complete(self, prompt: str, **kwargs) -> CompletionResponse:
...
+        except Exception as e:
+            print(f"Error fetching models from {models_url}: {e}. \"model\" parameter will not be included with inference call.")
+            return None
+
+    def _get_default_model(self) -> str:
+        """
+        Returns the cached default model if available; otherwise fetches and caches it.
+        """
+        if not self._default_model:
+            self._default_model = self._fetch_default_model()
```
zhuangqh, line 98: this will end with api call each time you use transformers one. because it's None. maybe we can add a boolean flag.

presets/ragengine/inference/inference.py
```diff
@@ -38,29 +41,89 @@ def complete(self, prompt: str, **kwargs) -> CompletionResponse:
...
 
-        # Dynamically extract the field from the response based on the specified response_field
-        # completion_text = response_data.get(RESPONSE_FIELD, "No response field found") # not necessary for now
-        return CompletionResponse(text=str(response_data))
+            response = requests.get(models_url, headers=headers)
+            response.raise_for_status()  # Raise an exception for HTTP errors
+
+            models = response.json().get("data", [])
+            return models[0].get("id") if models else None
+        except Exception as e:
```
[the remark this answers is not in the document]
ishaansehgal99: current approach (catching Exception and handling 404 implicitly) is sufficient. It already logs 404, which can occur if vLLM is initializing or if the endpoint doesn't exist (e.g., non-vLLM). Let me know if you have a specific log message in mind to add

presets/ragengine/inference/inference.py
ishaansehgal99: Fixed this with dynamic check

presets/ragengine/inference/inference.py
```diff
@@ -38,29 +51,99 @@ def complete(self, prompt: str, **kwargs) -> CompletionResponse:
...
+                    f"Response: {err_msg}. Attempting to fetch the default model..."
+                )
+                self._default_model = self._fetch_default_model()  # Fetch default model dynamically
+                if self._default_model:
+                    logger.info(f"Default model '{self._default_model}' fetched successfully. Retrying request...")
+                    data["model"] = self._default_model
+                    return self._post_request(data, headers=DEFAULT_HEADERS)
+                else:
+                    logger.error("Failed to fetch a default model. Aborting retry.")
+            raise # Re-raise the exception if not recoverable
```
ishaansehgal99, line 87: Note here we hit inference endpoint. If we are returned with a missing `model` parameter error then we attempt to hit models endpoint. 

This makes the check targeted, as it will retry hitting this endpoint as long as there is a missing models parameter error.

presets/ragengine/inference/inference.py
```diff
@@ -38,29 +51,99 @@ def complete(self, prompt: str, **kwargs) -> CompletionResponse:
...
-        response_data = response.json()
-        return CompletionResponse(text=str(response_data))
+        return self._post_request(
+            {"messages": [{"role": "user", "content": prompt}]},
+            headers={"Authorization": f"Bearer {LLM_ACCESS_SECRET}"}
+        )
 
     def _custom_api_complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
-        headers = {"Authorization": f"Bearer {LLM_ACCESS_SECRET}"}
+        model = kwargs.pop("model", self._get_default_model())
```
ishaansehgal99, line 63: note: this first gets what is user specified "model" if that is not present it gets default dynamically by hitting /v1/models endpoint

presets/ragengine/inference/inference.py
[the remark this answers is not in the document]
zhuangqh: if vLLM is initializing, you will get a network error instead of 404. because http server is not launched.

zhuangqh (approved):

bangqipropel (approved):

presets/ragengine/inference/inference.py
```diff
@@ -38,29 +52,99 @@ def complete(self, prompt: str, **kwargs) -> CompletionResponse:
             self.params = {}
 
     def _openai_complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
-        llm = OpenAI(
-            api_key=LLM_ACCESS_SECRET,
-            **kwargs  # Pass all kwargs directly; kwargs may include model, temperature, max_tokens, etc.
-        )
-        return llm.complete(prompt)
+        return OpenAI(api_key=LLM_ACCESS_SECRET, **kwargs).complete(prompt)
```
bangqipropel, line 55: I suppose this part do something similar to the vllm part but if you want to merge it now for efficiency, it is good

PR #847, opened 2025-01-29
feat: Paginate & Truncate List Documents, Add RAG FastAPI Docstrings, Optimize Multi-Document Retrieval with Async Gather

ishaansehgal99:
**Reason for Change**:
Add Pagination and Truncation for List Documents Endpoints


presets/ragengine/main.py
```diff
@@ -80,7 +80,12 @@ def list_indexes():
         raise HTTPException(status_code=500, detail=str(e))
 
 @app.get("/indexes/{index_name}/documents", response_model=ListDocumentsResponse)
-async def list_documents_in_index(index_name: str):
+async def list_documents_in_index(
```
Fei-Guo, line 83: Can you add detailed examples in the change description or code comments about how to use this API? E.g., in which cases, limit/offset/max_text_length should be used?

presets/ragengine/main.py
```diff
@@ -90,23 +95,47 @@ async def list_documents_in_index(index_name: str):
...
-    index?query       | index%3Fquery     | index?query
     """
     try:
         # Decode the index_name in case it was URL-encoded by the client
         decoded_index_name = unquote(index_name)
-        documents = await rag_ops.list_documents_in_index(decoded_index_name)
-        return ListDocumentsResponse(documents=documents)
+        documents = await rag_ops.list_documents_in_index(
+            index_name=decoded_index_name,
+            limit=limit+1, # Fetch one extra to check for more results
```
Fei-Guo, line 104: I don't understand the +1 part here.

ishaansehgal99: Yes we grab one more than the limit to see if there are more possible documents that can be returned - it is to set the has_more and next_offset fields

presets/ragengine/vector_store/base.py
```diff
@@ -170,53 +172,109 @@ async def add_document_to_index(self, index_name: str, document: Document, doc_i
...
-                # Use async retrieval for non-simple doc_store
-                doc_info = await doc_store.aget_document(doc_id)
-                doc_map[doc_info.ref_doc_id] = {
-                    "text": doc_info.text,
-                    "hash": doc_info.hash
-                }
-        return doc_map
+        return {doc["doc_id"]: doc for doc in docs if isinstance(doc, dict)}
+
+    async def list_all_documents(
```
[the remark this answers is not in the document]
ishaansehgal99: done

presets/ragengine/vector_store/base.py
```diff
@@ -170,53 +172,109 @@ async def add_document_to_index(self, index_name: str, document: Document, doc_i
...
+        if not index_names:
+            return {}  # No indexes available
 
-    async def list_all_documents(self) -> Dict[str, Dict[str, Dict[str, str]]]:
-        """Common logic for listing all documents."""
-        indexes: Dict[str, Dict[str, Dict[str, str]]] = {}
-        for index_name, vector_store_index in self.index_map.items():
+        result: Dict[str, List[Dict[str, any]]] = defaultdict(list)
+
+        # Calculate per-index limits and offsets
```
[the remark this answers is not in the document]
ishaansehgal99: done

presets/ragengine/main.py
ishaansehgal99: done added docstrings in fastapi format

presets/ragengine/vector_store/base.py
```diff
@@ -170,53 +172,109 @@ async def add_document_to_index(self, index_name: str, document: Document, doc_i
...
+    async def list_documents_paginated(
+            self,
+            limit: int,
+            offset: int,
+            max_text_length: Optional[int] = None
+    ) -> Dict[str, List[Dict[str, any]]]:
+        """
+        List documents across all indexes with fair distribution and global pagination.
+        """
+        index_names = list(self.index_map.keys())
```
zhuangqh, line 234: are these keys well sorted?

presets/ragengine/models.py
```diff
@@ -8,12 +8,19 @@
...
 class DocumentResponse(BaseModel):
     doc_id: str
     text: str
-    metadata: Optional[dict] = None
+    hash_value: Optional[str] = None
+    metadata: Optional[dict] = Field(default_factory=dict)
+    is_truncated: bool = False
+
+class ListDocumentsResponse(BaseModel):
+    documents: Dict[str, List[DocumentResponse]] # Mapping of index names to DocumentResponse
```
zhuangqh, line 21: ithink the response should be a list instead of a hashmap for a listing api.

presets/ragengine/vector_store/base.py
ishaansehgal99: we have removed this code

zhuangqh (approved):
please ensure UT is good before merge.

presets/ragengine/main.py
```diff
@@ -72,15 +142,68 @@ async def query_index(request: QueryRequest):
...
+      "count": 5
+    }
+    ```
+    """,
+)
+async def list_documents_in_index(
+    index_name: str,
+    limit: int = Query(10, ge=1, le=100, description="Maximum number of documents to return"),
+    offset: int = Query(0, ge=0, description="Starting point for the document list"),
+    max_text_length: Optional[int] = Query(1000, ge=1, description="Maximum text length to return"),
```
Fei-Guo, line 205: max_text_length means the max text length returned per document. We need to make it clear.  Otherwise, user may think it is the max_length for all documents

ishaansehgal99: ok I fixed the description, I think thats sufficient but if you think name change is needed happy to do that

ishaansehgal99: could rename it to `max_text_length_per_document` just find it wordy if we can avoid

Fei-Guo (approved):

PR #848, opened 2025-01-29
feat: Add Deepseek Model

ishaansehgal99:
**Reason for Change**:
Add Deepseek models


Fei-Guo (approved):

clarenceb:
Are there plans to add the larger DeepSeek R1 models like 32B and 70B?

PR #859, opened 2025-01-31
fix: DeepSeek Model Naming

ishaansehgal99:
**Reason for Change**:
Nit on model naming and typo


zhuangqh:
deepseekQwen14bRunParamsVLLM = map[string]string{
                "dtype": "float16",
        }

change this to bfloat16

Fei-Guo (approved):

PR #875, opened 2025-02-11
chore: RAG pipeline leaking cluster patch

bangqipropel:
**Reason for Change**:
1. Use composite steps for avoiding cleaning step is not run when interrupted.
2. Add RAG pipeline back when rag engine related codes are changed.
3. Use the image aimodelsregistrytest.azurecr.io/kaito-rag-service:0.0.7 for now, will change it back when MCR image is good
4. Changed the image in the unit test too

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/.github/actions/e2e-base-setup/action.yml
+++ b/.github/actions/e2e-base-setup/action.yml
@@ -1,0 +1,206 @@
+name: "e2e-base-setup"
+description: "Composite action for e2e base setup"
+
+inputs:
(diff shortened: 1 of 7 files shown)


Fei-Guo:
Approved, But do check the poll interval of each step in the test. The overall test runtime is longer than it normally needs.

Fei-Guo (approved):

PR #879, opened 2025-02-12
fix: skip probing for incompatible scenario

zhuangqh:
When determining the max available context length, vllm will
generate dummy inputs with `max_num_batched_tokens` tokens
and profiling peak VRAM usage.

1. skip probing for multi-gpu serving

If tensor-parallelism is enabled, there are more than 1 gpu workers. 
However, we have no way to adjust the context length for profiling
for these workers in vllm.

2. skip probing when chunked-prefill is enabled

[chunked prefill](https://arxiv.org/pdf/2308.16369) is a feature
to improve thoughput for long context serving (max_model_len > 32K).
It makes the peak memory usage nonlinearly related to max_model_len.

**Notes for Reviewers**:

test case:
- [x] single gpu with enough vram to run max-model-len
- [x] single gpu, find max-model-len after several probing
- [x] single gpu, unable to run with even small model len
- [x] multi gpu
- [x] chunked prefill enabled

### DIFF
--- a/presets/workspace/inference/vllm/inference_api.py
+++ b/presets/workspace/inference/vllm/inference_api.py
@@ -38,7 +38,7 @@ class KAITOArgumentParser(argparse.ArgumentParser):
         # They should start with "kaito-" prefix to avoid conflict with vllm args
         self.add_argument("--kaito-adapters-dir", type=str, default="/mnt/adapter", help="Directory where adapters are stored in KAITO preset.")
         self.add_argument("--kaito-config-file", type=str, default="", help="Additional args for KAITO preset.")
-        self.add_argument("--kaito-max-probe-steps", type=int, default=6, help="Maximum number of steps to find the max available seq len fitting in the GPU memory.")
+        self.add_argument("--kaito-max-probe-steps", type=int, help="Maximum number of steps to find the max available seq len fitting in the GPU memory.")
 
     def _reset_vllm_defaults(self):
         local_rank = int(os.environ.get("LOCAL_RANK",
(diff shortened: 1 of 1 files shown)


presets/workspace/inference/vllm/inference_api.py
```diff
@@ -140,15 +143,14 @@ def find_max_available_seq_len(engine_config: EngineConfig, max_probe_steps: int
         observability_config=engine_config.observability_config,
     )
 
-    model_max_blocks = int(engine_config.model_config.max_model_len / engine_config.cache_config.block_size)
```
zhuangqh, line 143: this several lines of change are just a refactoring to make it more clear.

Fei-Guo (approved):

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -127,6 +127,9 @@ def find_max_available_seq_len(engine_config: EngineConfig, max_probe_steps: int
     """
     # see https://github.com/vllm-project/vllm/blob/v0.6.3/vllm/engine/llm_engine.py#L335
     executor_class = LLMEngine._get_executor_cls(engine_config)
+    if engine_config.scheduler_config.enable_chunked_prefill:
+        logger.info("Chunked Prefill is enabled, skip probing.")
+        return engine_config.model_config.max_model_len
```
ishaansehgal99, line 132: chunked prefill - cool feature

ishaansehgal99 (approved):

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, *args, **kwargs):
         # They should start with "kaito-" prefix to avoid conflict with vllm args
         self.add_argument("--kaito-adapters-dir", type=str, default="/mnt/adapter", help="Directory where adapters are stored in KAITO preset.")
         self.add_argument("--kaito-config-file", type=str, default="", help="Additional args for KAITO preset.")
-        self.add_argument("--kaito-max-probe-steps", type=int, default=6, help="Maximum number of steps to find the max available seq len fitting in the GPU memory.")
+        self.add_argument("--kaito-max-probe-steps", type=int, help="Maximum number of steps to find the max available seq len fitting in the GPU memory.")
```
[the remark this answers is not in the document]
zhuangqh: right.

PR #904, opened 2025-02-28
feat: Introduce workspace v1beta1 API

Fei-Guo:
**Reason for Change**:
This change introduces the v1beta1 version for workspace CRD. Note that ragengine is still in v1alpha1.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

This is a relatively big change but is straightforward mostly because we did not change any workspace CRD APIs between beta and alpha. The whole files under api/v1beta are directly copied from the api/v1alpha folder. The only important change is the line 181 in workspace_types.go to make v1beta1 as the CRD storage version. 

There are some files shared by ragengine and workspace. This means some codes in v1beta1 fold are irrelevant to workspace crd. I will have a follow-up PR to clean up the irrelevant codes.

To accommodate the new CRD version, I switched to use v1beta in all workspace related controller/test code. Any code related to REGEngine API is unchanged (still use v1alpha1).

One important question is do we need to implement a conversion webhook to deal with different api versions of workspace CRs. Upon some manual testing, we found this is unnecessary. APIServer will automatically perform the conversion between v1beta1 and v1alpha1 and it works well when there are not fields changes between versions. This means the client can ask for a v1alpha1 or v1beta1 version of the object. The watcher can watch for both versions even when you install the watch for v1beta1 only. This largely simplifies the required code changes. 

For all the documents, we will switch to use v1beta1 later (after the new release of v1beta1).


zhuangqh (approved):
"APIServer will automatically perform the conversion between v1beta1 and v1alpha1 and it works well when there are not fields changes between versions. "
That's good.

bangqipropel (approved):
basically good for me. 

Don't forget to change the unit-test in the next PR:
`grep -v -e /vendor -e /api/v1alpha1/zz_generated.deepcopy.go -e /pkg/utils/test/...) \`

pkg/workspace/webhooks/webhooks.go
```diff
@@ -35,4 +36,5 @@ func NewWorkspaceCRDValidationWebhook(ctx context.Context, _ configmap.Watcher)
 
 var WorkspaceResources = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	kaitov1alpha1.GroupVersion.WithKind("Workspace"): &kaitov1alpha1.Workspace{},
+	kaitov1alpha1.GroupVersion.WithKind("Workspace"): &kaitov1beta1.Workspace{},
 }
```
ishaansehgal99, line 40:
```suggestion
	kaitov1alpha1.GroupVersion.WithKind("Workspace"): &kaitov1alpha1.Workspace{},
	kaitov1beta1.GroupVersion.WithKind("Workspace"): &kaitov1beta1.Workspace{},
}
```

Fei-Guo: fixed

Fei-Guo:
> basically good for me.
> 
> Don't forget to change the unit-test in the next PR: `grep -v -e /vendor -e /api/v1alpha1/zz_generated.deepcopy.go -e /pkg/utils/test/...) \`

fixed

PR #913, opened 2025-03-04
feat: Add load and persist endpoints tests to RAGEngine E2E

ishaansehgal99:
**Reason for Change**:

Add load and persist endpoints tests to RAGEngine E2E

### DIFF
--- a/presets/ragengine/main.py
+++ b/presets/ragengine/main.py
@@ -310,9 +310,12 @@ async def persist_index(
 )
 async def load_index(
         index_name: str,
-        path: str = Query(DEFAULT_VECTOR_DB_PERSIST_DIR, description="Path to load the index from"),
+        path: Optional[str] = Query(None, description="Path to load the index from"),
         overwrite: bool = Query(False, description="Overwrite the existing index if it already exists")
 ):
+    # If no path is provided, use the default directory joined with index_name.
+    if path is None:
+        path = os.path.join(DEFAULT_VECTOR_DB_PERSIST_DIR, index_name)
     try:
         await rag_ops.load(index_name, path, overwrite)
         return {"message": f"Successfully loaded index {index_name} from {path}."}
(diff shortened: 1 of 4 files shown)


ishaansehgal99:
Preparing review...

presets/ragengine/main.py
```diff
@@ -310,9 +310,12 @@ async def persist_index(
 )
 async def load_index(
         index_name: str,
-        path: str = Query(DEFAULT_VECTOR_DB_PERSIST_DIR, description="Path to load the index from"),
+        path: Optional[str] = Query(None, description="Path to load the index from"),
         overwrite: bool = Query(False, description="Overwrite the existing index if it already exists")
 ):
+    # If no path is provided, use the default directory joined with index_name.
+    if path is None:
+        path = os.path.join(DEFAULT_VECTOR_DB_PERSIST_DIR, index_name)
```
ishaansehgal99, line 318: Default here needed to include index name - by default we persist at /storage/index_name - so we will by default load from there as well (given user doesn't provide path)

Fei-Guo (approved):

test/rage2e/rag_test.go
```diff
@@ -471,114 +502,178 @@ func createAndValidateIndexPod(ragengineObj *kaitov1alpha1.RAGEngine) error {
...
+			logs, err := utils.GetPodLogs(coreClient, ragengineObj.Namespace, podName, "")
 			if err != nil {
-				GinkgoWriter.Printf("Failed to get logs from pod %s: %v\n", "indexpod", err)
+				GinkgoWriter.Printf("Failed to get logs from pod %s: %v\n", podName, err)
 				return false
 			}
 
 			return strings.Contains(logs, "Kaito is an operator that automates the AI/ML model inference or tuning workload in a Kubernetes cluster")
-		}, 2*time.Minute, utils.PollInterval).Should(BeTrue(), "Failed to wait for index logs to be ready")
+		}, 10*time.Minute, utils.PollInterval).Should(BeTrue(), "Failed to wait for index logs to be ready")
```
bangqipropel, line 512: Is it necessary to change it to 10 minutes? I think 10 minutes is too long

bangqipropel (commented):
All others look good to me, but how long the persist and load will take? Looks 5/10 minutes much longer than my expected

test/rage2e/rag_test.go
ishaansehgal99: Yes this is the max timeout, it will proceed sooner if it succeeds. I increased it once because curl pod was stuck in container creating past two minutes. But I  agree we should lower it, lowered to 4mins.

PR #948, opened 2025-03-25
feat: onboard arc instance type

qfai:
**Reason for Change**:

Instacne type are used to determine how many GPU are used, AKS arc also require this feature to support multiple GPUs.
The official link of AKS arc supported GPU are [here](https://learn.microsoft.com/en-us/azure/aks/aksarc/scale-requirements)
We will set cloudProviderName: "arc" in arc's environment.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo (approved):

Fei-Guo:
@qfai , can you fix test errors?

qfai:
Thanks @Fei-Guo for the review, would you mind starting the workflows for the PR? Seems like I don't have the permission.

PR #950, opened 2025-03-25
feat: reimplement pull and push of images with skopeo and oras

### LINKED ISSUE #901: Use skopeo as the image management tool to manage adapters

Currently, we use docker to push adapter images using busybox base image. It introduces a few problems:
1) Busybox base image can potentially contain CVEs, which may pollute user container registries;
2) We are not building adapter images with mutli-arch support, so it does not work in ARM GPU;
3) Docker may hit issues when used in rootless containers. 

We will change the way we load adapters in the init-containers.
- The adapter image will NOT be the image of the init-container
- We use base image of tools like "skopeo" as the init-container image, which will be used to download the adapter image locally (completely avoid the multi-arch issue) and extract the data layer out. 
- We use skopeo in the sidecar container to push container images without using busybox base image (using distroless or other safe base images)

t0rr3sp3dr0:
**Reason for Change**:
This patch allows us to get rid of DinD, which currently runs in a privileged container. Skopeo will be used to pull images, mimicking the exact pull behavior of [K8s Image Volumes](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/). While ORAS is used to build the image in a way compatible with the implementation of [K8s Image Volumes](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/) in `containerd` and `cri-o`.

**Requirements**

**Issue Fixed**:
Fixes #901

**Notes for Reviewers**:
After this patch, images produced with newer versions of Kaito won't run on older versions of Kaito. However, images produced by older versions of Kaito will run on newer versions of Kaito. We probably want to make this clear on the release notes.

### DIFF
--- a/api/v1beta1/workspace_validation_test.go
+++ b/api/v1beta1/workspace_validation_test.go
@@ -1167,8 +1167,8 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 		{
 			name: "All fields valid",
 			tuningSpec: &TuningSpec{
-				Input:  &DataSource{Name: "valid-input", Image: "AZURE_ACR.azurecr.io/test:0.0.0"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0", ImagePushSecret: "secret"},
+				Input:  &DataSource{Name: "valid-input", Image: "kaito.azurecr.io/test:0.0.0"},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0", ImagePushSecret: "secret"},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodLora,
 			},
@@ -1178,8 +1178,8 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 		{
 			name: "Verify QLoRA Config",
 			tuningSpec: &TuningSpec{
-				Input:  &DataSource{Name: "valid-input", Image: "AZURE_ACR.azurecr.io/test:0.0.0"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0", ImagePushSecret: "secret"},
+				Input:  &DataSource{Name: "valid-input", Image: "kaito.azurecr.io/test:0.0.0"},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0", ImagePushSecret: "secret"},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodQLora,
 			},
@@ -1189,7 +1189,7 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 		{
 			name: "Missing Input",
 			tuningSpec: &TuningSpec{
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0", ImagePushSecret: ""},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0", ImagePushSecret: ""},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodLora,
 			},
@@ -1210,7 +1210,7 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 			name: "Missing Preset",
 			tuningSpec: &TuningSpec{
 				Input:  &DataSource{Name: "valid-input"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0", ImagePushSecret: ""},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0", ImagePushSecret: ""},
 				Method: TuningMethodLora,
 			},
 			wantErr:   true,
@@ -1220,7 +1220,7 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 			name: "Invalid Preset",
 			tuningSpec: &TuningSpec{
 				Input:  &DataSource{Name: "valid-input"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0", ImagePushSecret: ""},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0", ImagePushSecret: ""},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("invalid-preset")}},
 				Method: TuningMethodLora,
 			},
@@ -1231,7 +1231,7 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 			name: "Invalid Method",
 			tuningSpec: &TuningSpec{
 				Input:  &DataSource{Name: "valid-input"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0", ImagePushSecret: ""},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0", ImagePushSecret: ""},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: "invalid-method",
 			},
@@ -1241,8 +1241,8 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 		{
 			name: "Invalid Input Source Casing",
 			tuningSpec: &TuningSpec{
-				Input:  &DataSource{Name: "valid-input", Image: "AZURE_ACR.azurecr.io/INPUT:0.0.0"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/output:0.0.0", ImagePushSecret: "secret"},
+				Input:  &DataSource{Name: "valid-input", Image: "kaito.azurecr.io/INPUT:0.0.0"},
+				Output: &DataDestination{Image: "kaito.azurecr.io/output:0.0.0", ImagePushSecret: "secret"},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodLora,
 			},
@@ -1252,8 +1252,8 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 		{
 			name: "Invalid Output Destination Casing",
 			tuningSpec: &TuningSpec{
-				Input:  &DataSource{Name: "valid-input", Image: "AZURE_ACR.azurecr.io/input:0.0.0"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/OUTPUT:0.0.0", ImagePushSecret: "secret"},
+				Input:  &DataSource{Name: "valid-input", Image: "kaito.azurecr.io/input:0.0.0"},
+				Output: &DataDestination{Image: "kaito.azurecr.io/OUTPUT:0.0.0", ImagePushSecret: "secret"},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodLora,
 			},
@@ -1295,13 +1295,13 @@ func TestTuningSpecValidateUpdate(t *testing.T) {
 			name: "No changes",
 			oldTuning: &TuningSpec{
 				Input:  &DataSource{Name: "input1"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0"},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0"},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodLora,
 			},
 			newTuning: &TuningSpec{
 				Input:  &DataSource{Name: "input1"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/test:0.0.0"},
+				Output: &DataDestination{Image: "kaito.azurecr.io/test:0.0.0"},
 				Preset: &PresetSpec{PresetMeta: PresetMeta{Name: ModelName("test-validation")}},
 				Method: TuningMethodLora,
 			},
@@ -1368,7 +1368,7 @@ func TestDataSourceValidateCreate(t *testing.T) {
 		{
 			name: "Volume specified only",
 			dataSource: &DataSource{
-				Image: "AZURE_ACR.azurecr.io/test:0.0.0",
+				Image: "kaito.azurecr.io/test:0.0.0",
 			},
 			wantErr: false,
 		},
@@ -1385,7 +1385,7 @@ func TestDataSourceValidateCreate(t *testing.T) {
 				Image:            "data-image:latest",
 				ImagePullSecrets: []string{"imagePushSecret"},
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 		{
 			name: "Image without Tag Specified",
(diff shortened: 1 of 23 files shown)


pkg/workspace/image/puller.go
```diff
@@ -0,0 +1,66 @@
+package image
```
Fei-Guo, line 1: License header please.

pkg/workspace/image/puller.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/sh
```
Fei-Guo, line 1: License header please.

pkg/workspace/image/puller_test.go
```diff
@@ -0,0 +1,71 @@
+package image
```
Fei-Guo, line 1: license header

pkg/workspace/image/pusher.go
```diff
@@ -0,0 +1,66 @@
+package image
```
Fei-Guo, line 1: license header

pkg/workspace/image/pusher.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
Fei-Guo, line 1: license header

pkg/workspace/image/pusher_test.go
```diff
@@ -0,0 +1,71 @@
+package image
```
Fei-Guo, line 1: license header

Fei-Guo (commented):
Add license header for all new files please.

pkg/workspace/image/puller.go
```diff
@@ -0,0 +1,66 @@
...
+	t, err := template.New("puller.sh").Parse(pullerSHTextData)
+	if err != nil {
+		panic(err)
+	}
+
+	pullerSHTemplate = t
+}
+
+func renderPullerSH(imgRef string, volDir string) (string, error) {
+	normalizedImgRef, err := reference.ParseDockerRef(imgRef)
```
Fei-Guo, line 30: Should we do this check in validation webhook? I was thinking to avoid any err path in this method.

t0rr3sp3dr0: Yeah, it makes sense for us to have that check in the webhook. I'll implement that!

But what should we do if we don't have a valid image reference at this point? I know it's the user's fault if they disable the webhook and create an invalid workspace, but I need to do something in code if that function fails. Should I just panic? Or should I continue with the original invalid value? The latter will eventually cause pull/push to fail.

t0rr3sp3dr0: Also, the rendering of the Go template can technically fail. But it won't, as long as we have good coverage on that method. So, I can safely make that a panic.

t0rr3sp3dr0: I think panic would be the best option for both situations if we don't want to have error handling on that path

zhuangqh: If cannot normalize, just use the original `imgRef`.
later steps will give users the error.

pkg/workspace/image/puller.sh
```diff
@@ -0,0 +1,58 @@
...
+
+    local DOCKER_CONFIG_D="${HOME}/.docker/config.d"
+    [ ! -e "${DOCKER_CONFIG_D}" ] || local REGISTRY_AUTH_FILES="$(find "${DOCKER_CONFIG_D}" -printf '%p ' -type 'f')"
+
+    for REGISTRY_AUTH_FILE in '' ${REGISTRY_AUTH_FILES}
+    do
+        local SKOPEO_ARGS='--dest-decompress'
+        [ -z "${REGISTRY_AUTH_FILE}" ] || SKOPEO_ARGS="${SKOPEO_ARGS} --authfile=${REGISTRY_AUTH_FILE}"
+
+        skopeo copy ${SKOPEO_ARGS} "docker://${IMG_REF}" "dir:${LAYOUT_DIR}" || continue
```
zhuangqh, line 28: I can see that oras supports image pulling too. Can we use it for pulling to avoid introduce two technical context?

pkg/workspace/image/pusher.sh
```diff
@@ -0,0 +1,67 @@
...
+#{{`
+
+wait() {
+    until [ -e "${VOL_DIR}/fine_tuning_completed.txt" ]
+    do
+        sleep 1
+    done
+}
+
+mklayer() {
```
zhuangqh, line 24: construct an oci image from scratch. It's hard for somebody who are not familar with oci spec to maintain this script.
Can you add this document https://github.com/opencontainers/image-spec/blob/main/manifest.md#example-image-manifest as a comment.

pkg/workspace/image/pusher.sh
```diff
@@ -0,0 +1,67 @@
...
+VOL_DIR="${1}"
+[ ! -z "${VOL_DIR}" ] || VOL_DIR='{{ .volDir }}'
+
+IMG_REF="${2}"
+[ ! -z "${IMG_REF}" ] || IMG_REF='{{ .imgRef }}'
+
+#{{`
+
+wait() {
+    until [ -e "${VOL_DIR}/fine_tuning_completed.txt" ]
```
zhuangqh, line 18: how about pass this sentinel file name through args? Determine it by upper control plane.
This script is expected to be simple and rarely updated.

pkg/workspace/image/pusher.sh
```diff
@@ -0,0 +1,67 @@
...
+    CONFIG_PATH="${TMPDIR}/config.json"
+
+    printf '{"rootfs":{"diff_ids":["sha256:%s"]}}' "${TAR_LAYER_DIFF}" > "${CONFIG_PATH}"
+}
+
+mklayout() {
+    LAYOUT_REF="${TMPDIR}/layout:latest"
+
+    cd "$(dirname "${TGZ_LAYER_PATH}")"
+    oras push --disable-path-validation --config "${CONFIG_PATH}:${CONFIG_MIME}" --oci-layout "${LAYOUT_REF}" "$(basename "${TGZ_LAYER_PATH}"):${TGZ_LAYER_MIME}"
```
zhuangqh, line 53: Do oras add some metadata to the image manifest? e.g. creation timestamp.
We can add some metadata to this image to make it easier to maintain.
e.g. maintainer: kaito.dev.xxx;   target model name this adapter files can be used for.

zhuangqh (commented):
Cool! Thanks for your investigation. Daemonless pull/push is good in this case.

pkg/workspace/image/puller.go
t0rr3sp3dr0: Okay, I've made the changes here. The methods in this package no longer return errors.

pkg/workspace/image/puller.sh
t0rr3sp3dr0: I discussed this briefly with Fei, and the reason for using Skopeo here is to keep the pull behavior compatible with the [K8s Image Volume](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/) implementation. The idea is that when this feature becomes GA and we no longer support older K8s versions, we will remove the custom puller and simply mount the image as a volume in the inference pod.

I've run some tests with both `containerd` and `cri-o`, and Skopeo pulls images in a way that is compatible with both container runtimes and supports images built by older versions of Kaito. These projects actually share code that is used in the pull process. While ORAS has its own implementation that differs from them, not allowing a transparent transition to K8s Image Volumes.

pkg/workspace/image/pusher.sh
t0rr3sp3dr0: By default, it only sets `org.opencontainers.image.created`.

image

t0rr3sp3dr0: We can add more stuff if we would like

pkg/workspace/image/pusher.sh
t0rr3sp3dr0: Yeah, sure!

t0rr3sp3dr0: One extra thing to consider here, and unfortunately, there's no documentation for it, is that I'm building this image in a way that is architecture-independent and compatible with the K8s Image Volume implementations of `containerd` and `cri-o`. I got here by trial and error, tweaking stuff until both implementations accepted my images.

t0rr3sp3dr0: I can write a README in this package, documenting my findings

pkg/workspace/image/pusher.sh
t0rr3sp3dr0: This was hard-coded in the previous script, so I thought it would be okay to do the same here.

Since I've written this in a way that allows local execution of the script, I can rewrite it in a way that uses the hard-coded values by default but allows these values to be changed when the script is rendered.

pkg/workspace/image/pusher.sh
zhuangqh: Can you get the model's name in controller code simply? If so, we can add a model name annotation here.
This is the most useful information for this image.

pkg/workspace/image/puller.sh
t0rr3sp3dr0: Transparent would mean that we would be able to swap our custom puller with an image volume, and there would be no behavior change at all. The most crucial factors here are how authentication and image selection based on the host architecture are handled.

I was having problems making ORAS deal with the new arch-independent image, but it turns out that it can handle the final version of the image I came up with. Another good indication that the final design of the image is widely compatible with existing tools and won't cause problems in the future.

But the main problem I have with ORAS right now is that it's not made to deal with regular Docker images, like the old ones produced by Kaito. I can make it fetch the image configuration, but I'll have to process it manually to find out the layers. I can fetch the layers with it, but I'll have to deal with decompression myself. In the end, we would have to install extra tools on the fly or build a new image with those tools and write more code to accomplish what Skopeo does by itself. Fei would like to avoid that and keep this simple.

pkg/workspace/image/pusher.sh
t0rr3sp3dr0: What do you think the label key should be? `sh.kaito.model`? `sh.kaito.model.name`? `sh.kaito.model-name`?

zhuangqh: `sh.kaito.model.name` this one works for me.

t0rr3sp3dr0: Done!

pkg/workspace/image/pusher.sh
t0rr3sp3dr0: Done!

pkg/workspace/image/pusher.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+
+#`}}
+
+wait
+mklayer
+mkconfig
+mkannotations
+mklayout
+push
```
Fei-Guo, line 80: Make sure if the push step fails, we have a way to copy the tuning outcome out of the container.

pkg/workspace/image/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+Kaito builds platform-independent images in a few steps described at a high level below. The implementation leverages [ORAS](https://github.com/oras-project/oras) to simplify the process, and its source code can be found in [pusher.sh](./pusher.sh). 
+
+1. Create the image layer by archiving `/data/adapter_config.json` and `/data/adapter_model.safetensors`.
+2. Compute the diff ID by hashing the image layer.
+3. Optionally compress the image layer.
+4. Create the image configuration with the diff ID.
+5. Optionally create the annotations file.
+6. Build the image layout with the image layer, image configuration, and annotations file.
+
```
Fei-Guo, line 37: We can keep the above, but we need to provide turn-key solution script and point to that script here.

test/e2e/preset_test.go
```diff
@@ -351,6 +376,11 @@ func validateResourceStatus(workspaceObj *kaitov1beta1.Workspace) {
 				return false
 			}
 
+			workspaceJSON, err := json.Marshal(workspaceObj)
```
zhuangqh, line 379: This line is inside a `Eventually` function. It means it will be printed multiple times.
I guess you only want the spec definition for debugging. Make sure it will be printed only once.
Otherwise, there are too much verbose logs.

zhuangqh: One coding suggestion, save it and print it after `Eventually`

test/e2e/preset_test.go
```diff
@@ -452,6 +496,11 @@ func validateRevision(workspaceObj *kaitov1beta1.Workspace, revisionStr string)
 					return false
 				}
 				isWorkloadAnnotationCorrect = dep.Annotations[WorkspaceRevisionAnnotation] == revisionStr
+
+				deploymentJSON, err := json.Marshal(dep)
+				Expect(err).NotTo(HaveOccurred())
+
+				GinkgoWriter.Printf("Deployment definition: %s\n", string(deploymentJSON))
```
zhuangqh, line 503: same here. Please check all of them.

test/e2e/preset_test.go
t0rr3sp3dr0: Fei asked me to extract those changes to another PR

test/e2e/preset_test.go
t0rr3sp3dr0: Removed from this PR

pkg/workspace/image/README.md
t0rr3sp3dr0: Done

pkg/workspace/tuning/preset-tuning.go
```diff
@@ -259,20 +200,14 @@ func CreatePresetTuning(ctx context.Context, workspaceObj *kaitov1beta1.Workspac
...
 
-	initContainer, imagePullSecrets, dataSourceVolume, dataSourceVolumeMount, err := prepareDataSource(ctx, workspaceObj)
-	if err != nil {
-		return nil, err
-	}
-	volumes = append(volumes, dataSourceVolume)
-	volumeMounts = append(volumeMounts, dataSourceVolumeMount)
+	initContainer, dataSourceVolumes, dataSourceVolumeMounts := prepareDataSource(ctx, workspaceObj)
+	volumes = append(volumes, dataSourceVolumes...)
+	volumeMounts = append(volumeMounts, dataSourceVolumeMounts...)
```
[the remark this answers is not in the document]
t0rr3sp3dr0: Done

pkg/workspace/manifests/manifests.go
```diff
@@ -194,7 +197,10 @@ func GenerateTuningJobManifest(ctx context.Context, wObj *kaitov1beta1.Workspace
 		kaitov1beta1.LabelWorkspaceName: wObj.Name,
 	}
 
-	// Add volume mounts to sidecar containers
+	for i := range initContainers {
```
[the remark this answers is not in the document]
t0rr3sp3dr0: Done

pkg/workspace/image/pusher.go
```diff
@@ -0,0 +1,80 @@
...
+	if annotationsData != nil {
+		annotationsData, err := json.Marshal(annotationsData)
+		if err != nil {
+			panic(err)
+		}
+
+		data["annotationsData"] = string(annotationsData)
+	}
+
+	if waitPath != nil {
```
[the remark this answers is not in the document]
t0rr3sp3dr0: Done

api/v1beta1/workspace_validation_test.go
```diff
@@ -1252,8 +1252,8 @@ func TestTuningSpecValidateCreate(t *testing.T) {
 		{
 			name: "Invalid Output Destination Casing",
 			tuningSpec: &TuningSpec{
-				Input:  &DataSource{Name: "valid-input", Image: "AZURE_ACR.azurecr.io/input:0.0.0"},
-				Output: &DataDestination{Image: "AZURE_ACR.azurecr.io/OUTPUT:0.0.0", ImagePushSecret: "secret"},
+				Input:  &DataSource{Name: "valid-input", Image: "kaito.azurecr.io/input:0.0.0"},
```
ishaansehgal99, line 1255: why rename

ishaansehgal99:
If you see ways, please try to split up PR.

api/v1beta1/workspace_validation_test.go
t0rr3sp3dr0: `AZURE_ACR.azurecr.io` is not a valid domain. You cannot have uppercase characters or underscores in a domain name. With the new validation method for image names, this was causing errors.

zhuangqh (approved):
Good job! Please make sure e2e test is always good before merging.

pkg/workspace/image/pusher.sh
t0rr3sp3dr0: Done! It's possible to attach an ephemeral container to the pod and access the tuning results.

image

Fei-Guo (approved):

PR #966, opened 2025-04-01
test: add more model preset testing config

zhuangqh:
**Reason for Change**:

cleanup some pipelines

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


.github/workflows/unit-tests.yml
```diff
@@ -55,6 +55,10 @@ jobs:
         run: |
           DEVICE=cpu make inference-api-e2e
 
+      - name: Run RAGEngine unit tests
+        run: |
+          make rag-service-test
+
```
bangqipropel, line 61: I think we are trying to split RAG engine and workspace instead of combining them.

.github/matrix-configs.json
zhuangqh, line 1: @ishaansehgal99 Can you confirm if this file is still needed?

.github/workflows/unit-tests.yml
zhuangqh: revert

Issue #976, opened 2025-04-03
Enable expandable_segments to reduce mem fragmentation


zhuangqh:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Enable this env `PYTORCH_CUDA_ALLOC_CONF=expandable_segments` by default.

**Describe alternatives you've considered**

**Additional context**

```
ERROR 04-01 23:46:56 [engine.py:160] torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 546.00 MiB. GPU 0 has a total capacity of 15.77 GiB of which 514.19 MiB is free. Process 562167 has 15.25 GiB memory in use. Of the allocated memory 13.73 GiB is allocated by PyTorch, with 46.00 MiB allocated in private pools (e.g., CUDA Graphs), and 736.86 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

PR #980, opened 2025-04-03
feat: Add Pytorch Expandable segments envvar

ishaansehgal99:
**Reason for Change**:
Set the PyTorch CUDA memory allocation configuration to use expandable segments for better memory management.


ishaansehgal99:
https://github.com/kaito-project/kaito/issues/976

PR #983, opened 2025-04-03
fix: Ensuring allowable SKU if SKUConfig not found

ishaansehgal99:
**Reason for Change**:
I noticed what I think is a slight logic bug here, I fixed it. But there is also the option to remove this check entirely, say someone wants to use a non-D or non-N series SKU.


ishaansehgal99:
Reminded of the original intent of this code

if the cloud-provider is not azure OR
if we are using azure but we have a sku which is not N_SERIES_PREFIX or D_SERIES_PREFIX

ishaansehgal99:
closing

PR #997, opened 2025-04-08
feat: check lora support availability

zhuangqh:
**Reason for Change**:

vllm doesn't support LoRA for falcon model.
```
[rank0]:     raise ValueError(
[rank0]: ValueError: Model FalconForCausalLM does not support LoRA, but LoRA is enabled. Support for this model may be added in the future. If this is important to you, please open an issue on github.
```

**Requirements**


pkg/model/interface.go
```diff
@@ -74,6 +74,9 @@ type VLLMParam struct {
 	DistributionParams map[string]string
 	// Parameters for running the model training/inference.
 	ModelRunParams map[string]string
+	// Indicates if vllm supports LoRA (Low-Rank Adaptation) for this model.
+	// doc: https://docs.vllm.ai/en/latest/models/supported_models.html#text-generation-task-generate
+	UnsupportsLoRA bool
```
Fei-Guo, line 79: UnsupportsLORA->DisallowLoRA?

ishaansehgal99 (approved):

pkg/model/interface.go
```diff
@@ -151,6 +154,9 @@ func (p *PresetParam) GetInferenceCommand(runtime RuntimeName, skuNumGPUs string
 		if !p.DisableTensorParallelism {
 			p.VLLM.ModelRunParams["tensor-parallel-size"] = skuNumGPUs
 		}
+		if !p.VLLM.UnsupportsLoRA {
```
zhuangqh, line 157: If users haven't provided an adapter, dont add this.

Fei-Guo (approved):
Nice change.

PR #1002, opened 2025-04-09
chore: refactor interface.go

Fei-Guo:
**Reason for Change**:

Code refactoring to improve readability. Following the suggestions provided by copilot. 

This change also fixes UT to initialize parameter maps.  

Fix a UT bug: we need to clear the adapters when the test does not need them.

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -109,18 +109,9 @@ func (h *HuggingfaceTransformersParam) DeepCopy() HuggingfaceTransformersParam {
 		return HuggingfaceTransformersParam{}
 	}
 	out := *h
-	out.TorchRunParams = make(map[string]string, len(h.TorchRunParams))
-	for k, v := range h.TorchRunParams {
-		out.TorchRunParams[k] = v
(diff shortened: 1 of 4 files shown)


pkg/model/interface.go
```diff
@@ -129,17 +121,18 @@ func (v *VLLMParam) DeepCopy() VLLMParam {
...
 	out := *v
-	out.DistributionParams = make(map[string]string, len(v.DistributionParams))
-	for k, v := range v.DistributionParams {
-		out.DistributionParams[k] = v
-	}
-	out.ModelRunParams = make(map[string]string, len(v.ModelRunParams))
-	for k, v := range v.ModelRunParams {
-		out.ModelRunParams[k] = v
-	}
+	out.DistributionParams = copyStringMap(v.DistributionParams)
```
zhuangqh, line 124:
```suggestion
	out.DistributionParams = maps.Clone(v.DistributionParams)
```

This is what you need.

zhuangqh (approved):

PR #1004, opened 2025-04-10
Update README.md

ttw2001:
Adding a flat list of Kaito usecases so users understand when to use Kaito and why clearly as part of README.md

**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo:
We don't update README.md for verbose info.  Please file an issue to describe the purpose first.

PR #1008, opened 2025-04-14
fix: allow backoff for tuning job

zhuangqh:


pkg/workspace/manifests/manifests.go
```diff
@@ -254,7 +253,6 @@ func GenerateTuningJobManifest(ctx context.Context, wObj *kaitov1beta1.Workspace
 			},
 		},
 		Spec: batchv1.JobSpec{
-			BackoffLimit: &numBackoff, // default is 6. A failed tuning job is unlikely to be self-recoverable, no need to recreate the pod.
```
ishaansehgal99, line 257: This removes backoff retries right? maybe want to retitle pr

ishaansehgal99 (approved):

ishaansehgal99 (changes requested):
Actually I remember now, if a tuning job is automatically restarted user doesn't have a chance to copy checkpoints/progress from the container. It's gone. I believe this is why we set it to 0. That and also as the comment indicates it is likely same issue arises again.

I think we should leave up to user if they want to manually restart the tuning job/investigate the errors.

zhuangqh:
> Actually I remember now, if a tuning job is automatically restarted user doesn't have a chance to copy checkpoints/progress from the container. It's gone. I believe this is why we set it to 0. That and also as the comment indicates it is likely same issue arises again.
> 
> I think we should leave up to user if they want to manually restart the tuning job/investigate the errors.

The initial idea of this pr was to let GRIT have a change to migrate this job. It requires job controller create a new pod.

pkg/workspace/manifests/manifests.go
ishaansehgal99, line 257: If we could save past jobs (including their data), we could enable backoff retries with or without GRIT.

ishaansehgal99 (approved):

Fei-Guo (approved):

pkg/workspace/manifests/manifests.go
Fei-Guo: This change makes sense in general. We should consider other approaches for saving intermittent data.

PR #1044, opened 2025-04-24
feat: parse supported_models.yaml in plugin pkg before starting manager

chewong:
**Reason for Change**:

Treats `presets/workspace/models/supported_models.yaml` as the source of truth for handling model-specific static information such as model versions, container image tags, and whether download is necessary or not.

**Requirements**

**Issue Fixed**:

N/A

**Notes for Reviewers**:

TODO in a follow-up PR: Eliminate any tags map in every `model.go` and treat the tag in `supported_models.yaml` as the source of truth.


pkg/utils/plugin/plugin.go
```diff
@@ -3,52 +3,141 @@
...
+	Runtime string `json:"runtime,omitempty"`
+
+	// Tag is the tag of the container image used to run the model.
+	Tag string `json:"tag,omitempty"`
+
+	// DownloadAtRuntime indicates whether the model should be downloaded
+	// at runtime. If set to true, the model will be downloaded when the
+	// model deployment is created. If set to false, it will use a container
+	// image that already contains the model weights.
+	DownloadAtRuntime bool `json:"download_at_runtime,omitempty"`
```
Fei-Guo, line 59: `json:"download_at_runtime,omitempty"` -> `json:"downloadAtRuntime,omitempty"`

Add // +optional

pkg/utils/plugin/plugin.go
```diff
@@ -3,52 +3,141 @@
...
 }
 
 var KaitoModelRegister ModelRegister
 
-// Register allows model to be added
-func (reg *ModelRegister) Register(r *Registration) {
+// Init initializes the KaitoModelRegister with a path to the supported_models.yaml
+// file. It reads the file, unmarshals the YAML data into a SupportedModels
+// struct, and registers each model instance using the RegisterInfo method.
+func (reg *ModelRegister) Init(supportedModelsFilePath string) error {
```
Fei-Guo, line 73: Rename to "InitModelInfo"

pkg/utils/plugin/plugin.go
```diff
@@ -3,52 +3,141 @@
...
+	reg.instances = make(map[string]*ModelInstance)
```
Fei-Guo, line 74: You don't need line 74, 75 because of line 102 and line 117

pkg/utils/plugin/plugin.go
```diff
@@ -3,52 +3,141 @@
...
+	// Currently, the only supported runtime is "tfs".
+	Runtime string `json:"runtime,omitempty"`
+
+	// Tag is the tag of the container image used to run the model.
+	Tag string `json:"tag,omitempty"`
+
+	// DownloadAtRuntime indicates whether the model should be downloaded
+	// at runtime. If set to true, the model will be downloaded when the
+	// model deployment is created. If set to false, it will use a container
+	// image that already contains the model weights.
```
Fei-Guo, line 58: "If set to true, the model will be downloaded when the model deployment is created, and the container image will always be the Kaito base image. If set to false, a container image whose name contains the model name will be used, in which the model weights are baked."

pkg/utils/plugin/plugin.go
```diff
@@ -3,52 +3,141 @@
...
+	// Version is the version of the model. It is a URL that points to the
+	// model's huggingface page, which contains the model's repository ID
+	// and revision ID, e.g. https://huggingface.co/mistralai/Mistral-7B-v0.3/commit/d8cadc02ac76bd617a919d50b092e59d2d110aff.
+	Version string `json:"version,omitempty"`
+
+	// Runtime is the runtime environment in which the model operates.
+	// Currently, the only supported runtime is "tfs".
+	Runtime string `json:"runtime,omitempty"`
+
+	// Tag is the tag of the container image used to run the model.
```
Fei-Guo, line 52: Add "if the model uses the Kaito base image, the tag field can be ignored"

add # +optional

pkg/utils/plugin/plugin.go
```diff
@@ -3,52 +3,141 @@
...
+// in the YAML file should not be hardcoded in the codebase.
+type SupportedModels struct {
+	// Models is a slice of ModelInfo structs that contains information about
+	// the supported models.
+	Models []ModelInfo `json:"models,omitempty"`
+}
+
+// ModelInfo is a struct that holds information about a model
+type ModelInfo struct {
+	// Name is the name of the model, which serves as a unique identifier.
```
Fei-Guo, line 34: The first 4 fields are required. Remove the json annotation "omitempty".

pkg/utils/plugin/plugin.go
zhuangqh, line 73: consider using go embed for this kind of static confile.

chewong: I can't embed files outside of `pkg/utils/plugin` folder and I don't want to move the file to there since it's a bigger change.

zhuangqh:
Could you share me the point to not reuse the old "&plugin.Registration" type？
We save all model related info there. (model name, type, tag)

pkg/utils/plugin/plugin.go
zhuangqh: add a pkg in presets/workspace/models/xxx.go to embed this file.  
ref this pkg where you use it.

chewong:
> Could you share me the point to not reuse the old "&plugin.Registration" type？ We save all model related info there. (model name, type, tag)

Good question. Let me see if there is a way to use embed and reuse `&plugin.Registration`

chewong:
@Fei-Guo @zhuangqh, I made further changes based on both of your comments:

- Used Go embed to read `supported_models.yaml` in a new file, `presets/workspace/models/metadata.go`, instead of calling `os.ReadFile` in `pkg/utils/plugin/plugin.go`, to keep the code cleaner.
- Reused the `&plugin.Registration` type for model registration by adding a new Metadata field. Overloaded `plugin.KaitoModelRegister.Register` to allow registration of both model instances and model metadata, eliminating the need to expose `RegisterInstance` and `RegisterInfo` by using a unified function for both.
- Added `_ "github.com/kaito-project/kaito/presets/workspace/models"` as the first import in `cmd/workspace/models.go`, ensuring metadata is loaded before model instances.
- Files changed went from 20 to 8.

ishaansehgal99:
Ideal if plugin register could be called once, otherwise approved

ishaansehgal99 (approved):

presets/workspace/models/metadata.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/kaito-project/kaito/pkg/model"
+)
+
+var (
+	//go:embed supported_models.yaml
+	supportedModelsYAML []byte
+
+	// SupportedModels is a global map that holds the source of truth
+	// for all supported models and their metadata.
+	SupportedModels map[string]*model.Metadata
```
Fei-Guo, line 18: You would need a sync.Mutex to protect the map. You can add it later though.

presets/workspace/models/metadata.go
```diff
@@ -0,0 +1,38 @@
+package models
```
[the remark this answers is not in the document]
chewong: Added a GitHub action job to check so I will remember to add it in the future

presets/workspace/models/metadata.go
Fei-Guo: Depends on whether go test -race would complain or not. Leave it for future change to validate.

Fei-Guo (approved):

zhuangqh (approved):

PR #1049, opened 2025-04-25
chore: avoid using goto by using retry.OnError

cvvz:
**Reason for Change**:

chore: avoid using goto by using retry.OnError
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo (approved):
Thanks for cleaning up the code!

PR #1052, opened 2025-04-26
fix: add node cache to avoid get/list node from apiserver

cvvz:
**Reason for Change**:

fix: add node cache to avoid get/list node from apiserver
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/ragengine/controllers/ragengine_controller.go
+++ b/pkg/ragengine/controllers/ragengine_controller.go
@@ -558,6 +558,14 @@ func (c *RAGEngineReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}); err != nil {
 		return err
 	}
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &corev1.Node{},
+		"metadata.name", func(rawObj client.Object) []string {
+			node := rawObj.(*corev1.Node)
+			return []string{node.Name}
+		}); err != nil {
+		return err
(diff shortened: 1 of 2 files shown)


Fei-Guo:
We intentionally avoid list node because 
1) it can be inefficient to watch nodes for large cluster.
2) We want to watch for nodeclaim instead. It is the delegate resource for every node used by kaito.

cvvz:
It just adds informer for node so the node resource can be cached locally, since no eventHandler added for it , node event won't trigger kaito reconcile.

cvvz:
I see. If we are struggling with the memory usage, we can put this behind for now, but I believe this should be a good optimization since in most case operator should not get/list from apiserver directly, I will raise an issue to track it.

Talk about memory usage, currently we cached all the pods in the cluster in [here](https://github.com/kaito-project/kaito/blob/main/pkg/workspace/controllers/workspace_controller.go#L744-L750), which will increase the memory usage as well. It was introduced from this [commit](https://github.com/kaito-project/kaito/commit/10c67bcc68a80398201254a43ecc511b0e990206
), and I think it is no use anymore and should be deleted. WDYT?

cvvz:
Add node informer will break https://github.com/kaito-project/kaito/pull/1070. 
We can not just add informer without adjust the node-related reconciliation logic. 
I will just close this PR for now.

PR #1056, opened 2025-04-26
chore: refactor unnecessary anonymous function

cvvz:
**Reason for Change**:

chore: refactor unnecessary anonymous function
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/api/v1alpha1/workspace_validation.go
+++ b/api/v1alpha1/workspace_validation.go
@@ -477,25 +477,23 @@ func (i *InferenceSpec) validateCreate(ctx context.Context, namespace string, in
 	// check if required fields are set
 	// this check only applies to vllm runtime
 	if runtime == model.RuntimeNameVLLM {
-		func() {
-			var (
-				cmName = i.Config
-				cmNS   = namespace
-				err    error
-			)
-			if cmName == "" {
-				klog.Infof("Inference config not specified. Using default: %q", DefaultInferenceConfigTemplate)
-				cmName = DefaultInferenceConfigTemplate
-				cmNS, err = utils.GetReleaseNamespace()
-				if err != nil {
-					errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to determine release namespace: %v", err), "namespace"))
-					return
-				}
+		var (
+			cmName = i.Config
+			cmNS   = namespace
+			err    error
+		)
+		if cmName == "" {
+			klog.Infof("Inference config not specified. Using default: %q", DefaultInferenceConfigTemplate)
+			cmName = DefaultInferenceConfigTemplate
+			cmNS, err = utils.GetReleaseNamespace()
+			if err != nil {
+				errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to determine release namespace: %v", err), "namespace"))
+				return errs
 			}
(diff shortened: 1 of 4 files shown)


zhuangqh:
The logic is different. We perform an Additive semantics here. It means check as much as possible errors.
I just don't want to immediately return in the parent function. Hence I add this block.

Fei-Guo:
For the changes in the webhook, I agree that the anonymous function seems to be unnecessary.  For the changes in the controller, I don't think the new code significantly improve the readability (the old code looks neater). What do you think?

Fei-Guo:
> The logic is different. We perform an Additive semantics here. It means check as much as possible errors. I just don't want to immediately return in the parent function. Hence I add this block.

For existing code, removing anonymous function has no impact. But usually, we don't add return call in the middle of webhook checks, which I agree.

cvvz:
The real problem is that we update conditions(status) too frequently, the anonymous function was introduced here to avoid too many `updateStatusConditionIfNotMatch` like this:
```
if updateErr := c.updateStatusConditionIfNotMatch(ctx, wObj, kaitov1beta1.WorkspaceConditionTypeSucceeded, metav1.ConditionFalse,
	"workspaceFailed", err.Error()); updateErr != nil {
	klog.ErrorS(updateErr, "failed to update workspace status", "workspace", klog.KObj(wObj))
	return reconcile.Result{}, updateErr
}
```
IMO, it should be improved from two aspects:
1. use `Recorder.Eventf` to track event rather than update condition/status during sync loop
2. Update crd status for only once at the end of sync loop.
I filed issue https://github.com/kaito-project/kaito/issues/1067 to track.

cvvz:
Talk to the way to handle error. 
IMO, the webhook can use `errs.Also` to collect and return as much as possible errors. I still can't get what's the goal of the anonymous function in webhook. Could you explain to me with some specific examples? Thanks.
As for the anonymous function in reconciler, it is used to avoid using too many `updateStatusConditionIfNotMatch`, as what I mentioned above, we should fix this from the root. I will revert my fix for this part.

zhuangqh:
Without anonymous function, the correct logic is 
```
		var (
			cmName = i.Config
			cmNS   = namespace
			err    error
		)
		if cmName == "" {
			klog.Infof("Inference config not specified. Using default: %q", DefaultInferenceConfigTemplate)
			cmName = DefaultInferenceConfigTemplate
			cmNS, err = utils.GetReleaseNamespace()
			if err != nil {
				errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to determine release namespace: %v", err), "namespace"))
			} else if err := i.validateConfigMap(ctx, cmNS, cmName, instanceType); err != nil {
				errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to evaluate validateConfigMap: %v", err), "Config"))
			}
		} else if err := i.validateConfigMap(ctx, cmNS, cmName, instanceType); err != nil {
			errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to evaluate validateConfigMap: %v", err), "Config"))
		}
```

just use anonymous function and simple return to make this logic clearer.

zhuangqh:
I don't hold strong point about this policy. If you persist in removing this anonymous function and return immediately. 
I'm ok.

PR #1057, opened 2025-04-26
fix: only add finalizer when the object is not in deletion state

cvvz:
**Reason for Change**:

Only add finalizer when the object is not in deletion state.
Otherwise the finalizer will be added/remove again and again until it is finally completely deleted. 
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo (approved):
Yes. If a deleted workspace without any finalizer triggers the reconcile, the ensureFinalizer call will most likely fails since the obj should be gone in apiserver.

PR #1058, opened 2025-04-26
fix: should not check preset model in sync loop

cvvz:
**Reason for Change**:

Should not check preset model in sync loop, this can lead to enqueue and reconcile endlessly.
Since there is already [the same check in validation webhook](https://github.com/kaito-project/kaito/blob/main/api/v1beta1/workspace_validation.go#L433-L436), so just delete the code should be fine.
Anyway, it's inappropriate to check the preset model in the sync loop.
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo (approved):

Fei-Guo (approved):
This makes sense. Thanks for the cleanup.

We have an extreme edge case which might trigger the problem. E.g., cx disables the webhook manually, but I think we can ignore it.

PR #1059, opened 2025-04-27
fix: add NodeClass informer and refactor CLOUD_PROVIDER env check

cvvz:
**Reason for Change**:

The PR fixed two things:
1. add NodeClass informer to avoid calling apiserver
2. avoid too many `os.Getenv("CLOUD_PROVIDER")` and env var check. Since this env is mandatory and should be set with proper value, we should check it before the controller start up, and pass the value in sync loop, rather than read and check it everywhere.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


zhuangqh (changes requested):
Check cloud_provider before the controller start up, this is good.

Pass `cloudProvider` env explicitly as a function parameter. This is meaningless.
The original code only has 7 calling of `os.getenv("CLOUD_PROVIDER").  Then you make this GLOBAL STATIC PARAMETER an explicit parameter of 17 functions. Do you really think it improve code readability?  And you still read this env in webhook in your latest code.

I suggest you just make "CLOUD_PROVIDER" a global const. Directly read from env is a good pattern for this kind of scenario. And it also convenient to make it an individual value in testing.

cvvz:
@zhuangqh  Thanks for your suggestion. I understand your concerns. This change causes the cloudprovider parameter to spread like a virus throughout the codebase. I also felt it was somewhat inelegant when I was making the changes, but I couldn't find a better solution.

Let me explain why I believe we should not fetch and validate environment variables inside the sync loop: this environment variable is mandatory, so our consensus is that its validity should be checked before the controller even starts. If it’s invalid, the controller should not start at all. My concern is that if we fetch and validate the environment variable inside the sync loop, then in the case of a misconfiguration, the controller would still start, but the reconciler would not function properly, leading to meaningless retries. Although I have now added a validation check before the controller runs, which should largely prevent this situation, fundamentally, validity checks like this should not exist inside the sync loop.However, as long as we use `os.Getenv`, we can't avoid such error handling. 

Passing the parameter explicitly isn't ideal, especially from a code review perspective, but I really don't have a better option. If we don't pass the value explicitly, we would either have to read it from a global variable, or wrap the related functions into methods of a struct—neither of which is feasible at the moment. Maybe in the future, we can refactor the functions into methods, so we won't have to pass around so many parameters.

Another point you raised is that I also read environment variables inside the webhook—you’re absolutely right. I actually intended to mention this. I don't want to read environment variables inside the webhook either, but because our webhook is implemented using the Knative library instead of the Kubebuilder scaffolding, I currently have no way to pass the environment variable value into it. I suggest that we use Kubebuilder to scaffold the webhook framework so that we can pass the environment variable value from the main function. I'll open a separate issue to discuss the webhook topic.

zhuangqh:
@cvvz Yes. Thanks for your contribution. I do think env var check before controller setup is necessary. I mean you can keep the original way to read this variable.

> we would either have to read it from a global variable

Could you explain to me why this solution don't work?  `cloudprovider` is a global and static variable, just read it from where you need it.

I suggest just replace `os.Getenv("CLOUD_PROVIDER")` with `os.Getenv(consts.CloudProviderEnv)`.

If we don't have a better solution, I suggest just keep it as it is.

cvvz:
I didn't mean that it could not work, it did work, especially after I added the check before the sync loop start.
What I mean is that we should avoid this kind of env value check(or any other similar validation) in sync loop, because when the env is misconfigured, the operator can never recover it by reconciling. However, we can't just read the value from env without checking it, so, we'd better not use `os.Getenv` in sync loop at all.

zhuangqh:
> I didn't mean that it could not work, it did work, especially after I added the check before the sync loop start. What I mean is that we should avoid this kind of env value check(or any other similar validation) in sync loop, because when the env is misconfigured and error returned, the operator will keep retrying but it can never recover it by reconciling, we should definitely avoid this situation. However, we can't just read the value from env without checking it, so, we'd better not use `os.Getenv` in sync loop at all.

You have added the check in `SetupWithManager`, right?  Then it's safe to do `os.Getenv` in sync loop.

OK. I come up with a simpler solution😊. Check this critical cloudProvider variable in main.go. If this provider is not supported, just crash. Then you don't need to validate this value.

cvvz:
> Check this critical cloudProvider variable in main.go. If this provider is not supported, just crash. Then you don't need to validate this value.

Yes, that's exactly what I did in this PR. https://github.com/kaito-project/kaito/pull/1059/files#diff-622797ccaf108961818663aa1074497d35dd2e5cc1356b1e2c5214a268a02b6e

> You have added the check in SetupWithManager, right? Then it's safe to do os.Getenv in sync loop.

Yes, it may be safe, if the check in main.go has no bug. But the kind of check should not exist in sync loop at all. Any error that can not be recovered by reconciliation should be avoid.

zhuangqh:
> > Check this critical cloudProvider variable in main.go. If this provider is not supported, just crash. Then you don't need to validate this value.
> 
> Yes, that's exactly what I did in this PR. https://github.com/kaito-project/kaito/pull/1059/files#diff-622797ccaf108961818663aa1074497d35dd2e5cc1356b1e2c5214a268a02b6e
> 
> > You have added the check in SetupWithManager, right? Then it's safe to do os.Getenv in sync loop.
> 
> Yes, it may be safe, if the check in main.go has no bug. But the kind of check should not exist in sync loop at all. Any error that can not be recovered by reconciliation should be avoid.

Nice. I don't know if I explained it clearly. I totally agree your idea.
Can we agree that we don't need to explicitly add parameters? Just use os.getEnv from where we need to read cloudprovider var.

cvvz:
Sorry, but I'm afraid we can't come to an agreement...  How about we discuss this further with more approvers? Thanks!

zhuangqh:
```go
var provider string
var providerVarLock sync.Mutex

func InitCloudProvider() error {
	cloud := os.Getenv("CLOUD_PROVIDER")
	unlock, err := SetCloudProviderWithLock(cloud)
	if err != nil {
		return err
	}
	defer unlock()
	return nil
}

func GetCloudProvider() string {
	if provider == "" {
		panic("cloud provider not initialized")
	}
	return provider
}

// use this in testing
func SetCloudProviderWithLock(cloud string) (func(), error) {
	// Lock the variable to prevent concurrent access
	providerVarLock.Lock()
	switch cloud {
	case consts.AzureCloudName, consts.AWSCloudName, consts.ArcCloudName:
		provider = cloud
		return func() {
			providerVarLock.Unlock()
		}, nil
	}
	providerVarLock.Unlock() // Unlock if the cloud provider is unsupported
	return nil, fmt.Errorf("unsupported cloud provider: %s", cloud)
}
```

An alternative solution.

cvvz:
Yes, global variable is a solution. 
I once thought about using global variables, but I still think passing parameters is better and safer, even though it makes more tedious modifications. 
In fact, if we are going to use global variable, I'd rather choose to keep it as is😄

zhuangqh:
> Yes, global variable is a solution. I once thought about using global variables, but I still think passing parameters is better and safer, even though it makes more tedious modifications. In fact, if we are going to use global variable, I'd rather choose to keep it as is😄

Interesting😊👍
I say you have checked this value when initializing controller, it's safe to use it (os.getenv) during sync loop. You say this simple yes or no check may have bug. If this check has bug, all of your code is meaningless, you still pass a buggy value everywhere.

I say just make this global static config a global variable. You say you still prefer to pass this variable through a deep calling chain.

I don't want to argue anymore. Then, just keep it as it is unless you give me a better solution.

Fei-Guo:
Do not pass extra parameter (which is a static value, never changed during the execution), it hurts my eyes. Sorry. 

I am even good with reading os.env. This pattern is commonly used in many projects. Again, we are not implementing a perfect controller that can cover all edge cases. It is just impractical.

cvvz:
Please take a close look at another piece of code in Kaito, and you'll understand.

In the `main.go` function, two environment variables, `WebhookServicePort` and `WebhookServiceName`, are passed. You can trace how these environment variables are used in the Knative code:

1. The values of the environment variables are retrieved through `os.Getenv()`, then stored in the context using `webhook.WithOptions` : [[kaito/cmd/workspace/main.go at main · kaito-project/kaito](https://github.com/kaito-project/kaito/blob/main/cmd/workspace/main.go#L147-L151)](https://github.com/kaito-project/kaito/blob/main/cmd/workspace/main.go#L147-L151)
2. The values are then retrieved from the context and stored in the `Options` struct through `webhook.GetOptions` : [[pkg/webhook/context.go at main · knative/pkg](https://github.com/knative/pkg/blob/main/webhook/context.go#L33-L39)](https://github.com/knative/pkg/blob/main/webhook/context.go#L33-L39)
3. The values in the `Options` struct are passed down further, either directly embedded into the `Webhook` struct ([[pkg/webhook/webhook.go at main · knative/pkg](https://github.com/knative/pkg/blob/main/webhook/webhook.go#L120-L121)](https://github.com/knative/pkg/blob/main/webhook/webhook.go#L120-L121)) or read from the `Options` and stored in the `reconciler` struct ([[pkg/webhook/certificates/controller.go at main · knative/pkg](https://github.com/knative/pkg/blob/main/webhook/certificates/controller.go#L62)](https://github.com/knative/pkg/blob/main/webhook/certificates/controller.go#L62))
4. In the subsequent reconcile loop, if it's a method of the `reconciler`, the values are directly fetched from the `reconciler` struct, or if it's a standalone function, the values are passed as function parameters ([[pkg/webhook/certificates/certificates.go at main · knative/pkg](https://github.com/knative/pkg/blob/main/webhook/certificates/certificates.go#L103)](https://github.com/knative/pkg/blob/main/webhook/certificates/certificates.go#L103)).

For setting environment variables and the reconcile logic, Knative's approach is to retrieve the environment variables from the top level using `os.Getenv()` and pass them down layer by layer, rather than calling `os.Getenv()` each time.

Back to this PR, as I mentioned, I can only pass `cloudprovider` as a parameter layer by layer because in some places we use functions instead of methods. If we could change them to methods, then we wouldn't need to pass parameters around, we could directly set the value in the struct instead, so the code change will be less tedious.

Using global variables or passing values through the context is not a great solution. The values should be set in the struct , just like Knative's reconciler ([[pkg/webhook/certificates/certificates.go at main · knative/pkg](https://github.com/knative/pkg/blob/main/webhook/certificates/certificates.go#L48)](https://github.com/knative/pkg/blob/main/webhook/certificates/certificates.go#L48)) and explicitly passed down. However, in the current situation, there are many functions rather than method, so the values have to be passed as function parameters. Other than making this PR look messy and hard to review, I don't see any major issues with this approach.

Fei-Guo:
"Other than making this PR look messy and hard to review" but this is the issue. I prioritize code readability over the potential problem that reading an env could bring.

cvvz:
> "Other than making this PR look messy and hard to review" but this is the issue. I prioritize code readability over the potential problem that reading an env could bring.

Agree. I'm ok to keep it as is only for this concern. 
The original purpose of this PR wasn't to modify the environment variable usage, I just made the change by the way. From a code review perspective, it is indeed unreasonable. Maybe we’ll address these subtle details in the future.

cvvz:
I filed a new PR https://github.com/kaito-project/kaito/pull/1069 without CLOUD_PROVIDER env related change and closed this one. PTAL, thanks.

cvvz:
I filed a new PR https://github.com/kaito-project/kaito/pull/1069 without CLOUD_PROVIDER env related change and closed this one. PTAL, thanks.

PR #1062, opened 2025-04-27
fix: move nodeclaim check to right place

cvvz:
**Reason for Change**:

The PR mainly did two things:
1. move nodeclaim check to right place. After creating nodeclaim, we should check the status in following reconcile, rather than polling to check the status. 
2. Fix a bug: This code `return reconcile.Result{Requeue: false}, err` actually cant stop reconciliation, because when error is not nil, the `reconcile.Result{Requeue: false}` will be ignored. We should use `reconcile.TerminalError`
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo:
Nice clean up. Fix UT please.

zhuangqh (approved):

cvvz:
IMO, since we have `CheckNodeClaimStatus` in `WaitForPendingNodeClaims`, there is no need to `CheckNodeClaimStatus` in `CreateNodeClaim`. Should we deleted the `CheckNodeClaimStatus` in `CreateNodeClaim`. cc @zhuangqh  @Fei-Guo

zhuangqh:
Oh I see. you mean `WorkspaceReconciler.CreateNodeClaim` not this [CreateNodeClaim](https://github.com/kaito-project/kaito/blob/main/pkg/utils/nodeclaim/nodeclaim.go#L244).

If I understand correctly, `CheckNodeClaimStatus` in `CreateNodeClaim` is the first time you create nodeclaim and pending waiting for it. `WaitForPendingNodeClaims` is something when this obj requeue or controller restarts.

If you remove that code, do you want to let it fail immediately at https://github.com/kaito-project/kaito/blob/main/pkg/workspace/controllers/workspace_controller.go#L487 and requeue for pending waiting?

Fei-Guo:
WaitForPendingNodeClaims is mainly used to protect a case where controller is restarted for various reasons, When the reconciliation loop restarts, we want it to wait for the nodeclaim created by previous reconciler loop to be ready before continuing.

cvvz:
No need to fail. Just return nil after creating nodeclaim. 
Since we watched nodeclaim, if it status changed another reconciliation will be triggered.

cvvz:
Actually, check status after creating can only be performed one time, because if no nodeclaim need to be created it won't be performed,  but check status in `WaitForPendingNodeClaims` can always be performed.

zhuangqh:
> No need to fail. Just return nil after creating nodeclaim. Since we watched nodeclaim, if it status changed another reconciliation will be triggered.

Yes. This way works. I agree with you.

Fei-Guo:
Let us keep all sync loop for now.

cvvz:
Yes, I'm not planning to make the changes in this PR. Just raised for discussion.
We need to first enhance the e2e testing capabilities before making major changes or refactoring.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -243,34 +244,17 @@ func GenerateEC2NodeClassManifest(ctx context.Context) *awsv1beta1.EC2NodeClass
...
-			klog.Error(consts.ErrorInstanceTypesUnavailable, "reconcile will not continue")
-			return fmt.Errorf(consts.ErrorInstanceTypesUnavailable)
-		}
+	err = kubeClient.Create(ctx, nodeClaimObj.DeepCopy(), &client.CreateOptions{})
+	if err != nil {
 		return err
-	})
+	}
+
+	return nil
```
rambohe-ch, line 257: how about optimize as following:

```
return kubeClient.Create(ctx, nodeClaimObj, &client.CreateOptions{})
```

I think there is no need to deepcopy object, because Create function will not modify object.

rambohe-ch:
> Yes, I'm not planning to make the changes in this PR. Just raised for discussion. We need to first enhance the e2e testing capabilities before making major changes or refactoring.

@cvvz It's a good idea to improve E2E tests at first before improving sync loops.

pkg/utils/nodeclaim/nodeclaim.go
cvvz: You are right, fixed.

rambohe-ch:
@Fei-Guo @zhuangqh This pull request has removed a nodeclaim status check, and make code more readable.  maybe we can include it in the 0.4.6 release.

rambohe-ch (approved):

PR #1069, opened 2025-04-28
fix: add nodeclass informer

cvvz:
**Reason for Change**:

add NodeClass informer to avoid calling apiserver
This PR is to replace https://github.com/kaito-project/kaito/pull/1059, which removed CLOUD_PROVIDER env related change according to the suggestion.
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/workspace/controllers/workspace_controller.go
+++ b/pkg/workspace/controllers/workspace_controller.go
@@ -50,6 +50,7 @@ import (
 	"github.com/kaito-project/kaito/pkg/workspace/inference"
 	"github.com/kaito-project/kaito/pkg/workspace/manifests"
 	"github.com/kaito-project/kaito/pkg/workspace/tuning"
+	"github.com/kaito-project/kaito/pkg/featuregates"
 )
 
 const (
@@ -745,6 +746,7 @@ func (c *WorkspaceReconciler) applyInference(ctx context.Context, wObj *kaitov1b
 // SetupWithManager sets up the controller with the Manager.
 func (c *WorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	c.Recorder = mgr.GetEventRecorderFor("Workspace")
+
 	if err := c.setupInformers(mgr); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


cvvz:
Abandon https://github.com/kaito-project/kaito/pull/1059 and file this one.

zhuangqh:
> ## PR Reviewer Guide 🔍
> #### (Review updated until commit [7b008bd](https://github.com/kaito-project/kaito/commit/7b008bd8a20388623650e12124a81976afb1a5dd))
> Here are some key observations to aid the review process:
> 
> ⏱️ **Estimated effort to review**: 3 🔵🔵🔵⚪⚪
> 🧪 **No relevant tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Possible Issue**](https://github.com/kaito-project/kaito/pull/1069/files#diff-6bb654738e06d041fd6cf471085650b66d73b5f3175f4a5a6320580c857acf48R185-R194)
> The `validCloudProvider` function is duplicated in both `main.go` files. It would be better to move this function to a shared utility file to avoid code duplication.
> 
> [**Possible Issue**](https://github.com/kaito-project/kaito/pull/1069/files#diff-622797ccaf108961818663aa1074497d35dd2e5cc1356b1e2c5214a268a02b6eR199-R208)
> The `validCloudProvider` function is duplicated in both `main.go` files. It would be better to move this function to a shared utility file to avoid code duplication.
> 
> [**Error Handling**](https://github.com/kaito-project/kaito/pull/1069/files#diff-48b5462b98109860e3cfdedcf776300516aa4d25678b4e008e829482b034a43eR780-R781)
> The `setupInformers` method returns an error if the `CLOUD_PROVIDER` environment variable is not set. However, it does not log the error before returning. Consider adding a log statement to provide more context.

@cvvz Can you adopt this AI suggestion? `The validCloudProvider function is duplicated in both main.go files. It would be better to move this function to a shared utility file to avoid code duplication.`  I think pkg/utils/consts is a good place for this.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -763,6 +763,37 @@ func (c *WorkspaceReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		}); err != nil {
+		return err
+	}
+
+	// Informer for NodeClass
+	provider := os.Getenv("CLOUD_PROVIDER")
+	if provider == "" {
+		return apis.ErrMissingField("CLOUD_PROVIDER environment variable must be set")
+	}
+	if provider == consts.AzureCloudName {
```
zhuangqh, line 782: add a guard here. 
```go
	if featuregates.FeatureGates[consts.FeatureFlagEnsureNodeClass] {
```
Not all node provisioner require the nodeclass.

Fei-Guo:
TBH, I don't think we need an informer for nodeclass for now. It is almost a static configuration and there is no value to build a wacher for it. And it is only used when we integrate with Karpenter. However, the "FeatureFlagEnsureNodeClass" feature flag is off by default. Even if you want to add an informer. This code needs to be guarded by the feature flag.

cvvz:
> we just read once during controller startup

The nodeclass get request is sent as long as it needs to create nodeclaim. 
The code path is CreateNodeClaim -> CheckNodeClass -> IsNodeClassAvailable -> kubeClient.Get

cvvz:
Also, I agree to apply nodeclass automatically once enable Karpenter rather than check it before create nodeclaim

cvvz:
Should not check nodeclass everytime before create nodeclaim, instead, check and create node class before reconciliation start.
No need for nodeclass informer. I will file another pr for better solution.

cvvz:
Check nodeclass before controller startup to avoid keep checking nodeclass before creating nodeclaim.
https://github.com/kaito-project/kaito/pull/1088

PR #1070, opened 2025-04-28
fix: ensureNodePlugins does not work

cvvz:
**Reason for Change**:

`resources.CheckNvidiaPlugin(ctx, nodeObj)` checked `nodeObj`, which is a cached value. Retry is no use.

Another problem is that `return nil` in `default` condition will always be executed no matter found or not found. I think this is not what we want. This fix assume that we only `return nil` when found. 

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


cvvz:
If we add node informer(https://github.com/kaito-project/kaito/pull/1052), `resources.GetNode` will get node from cache rather than apiserver. But this fix should still work, because node informer will make sure the cache always sync with apiserver.

Fei-Guo (approved):

Issue #1050, opened 2025-04-25
RAGEngine Index Document API returns wrong doc_id's


bfoley13:
**Describe the bug**
When calling `/index` with multiple documents you can hit a race condition where doc_id's [attached to documents on the response](https://github.com/kaito-project/kaito/blob/bb4635cfc63a78529bf35653e0e051f48b739aad/presets/ragengine/main.py#L99) are not the correct id's of the documents. Likely due to [this async gather](https://github.com/kaito-project/kaito/blob/main/presets/ragengine/vector_store/base.py#L77)

**Steps To Reproduce**
Debug the service and hit:

```
curl -X POST -H "Content-Type: application/json" -d '{"index_name":"test_index","documents":[{"text": "This is a test document"},{"text": "Another test document"}]}' localhost:5000/index
```
**Expected behavior**

Document Id's are for correct document

**Logs**

**Environment**

- Kubernetes version (use `kubectl version`): N/A
- OS (e.g: `cat /etc/os-release`): MAC OS 
- Install tools:
- Others:

**Additional context**
In this attached example, you an see the response from the server vs the docstore containing the documents show different ID's

![Image](https://github.com/user-attachments/assets/d4d91e1c-278a-4cab-8772-53bca6c3c8b5)

PR #1071, opened 2025-04-28
fix: ragengine index document returns wrong ids

bfoley13:
**Reason for Change**:
Fixes the `/index` api responses to have correct ID's for documents. This is important for update/delete API's to come.

**Requirements**

**Issue Fixed**:
Fixes: #1050 

**Notes for Reviewers**:


presets/ragengine/tests/api/test_main.py
```diff
@@ -48,11 +48,11 @@ async def test_index_documents_success(async_client):
     response = await async_client.post("/index", json=request_data)
     assert response.status_code == 200
     doc1, doc2 = response.json()
-    assert (doc1["text"] == "This is a test document")
+    assert (doc1["text"] == "This is a test document" or
```
bfoley13, line 51: running `make rag-service-test` locally multiple times causes the original to fail. seems like the indexed arent deleted between tests?

presets/ragengine/vector_store_manager/manager.py
```diff
@@ -10,9 +10,10 @@ class VectorStoreManager:
     def __init__(self, vector_store: BaseVectorStore):
         self.vector_store = vector_store
 
-    async def index(self, index_name: str, documents: List[Document]) -> List[str]:
+    async def index(self, index_name: str, documents: List[Document]) -> List[Dict[str, Any]]:
         """Index new documents."""
-        return await self.vector_store.index_documents(index_name, documents)
+        ids = await self.vector_store.index_documents(index_name, documents)
+        return await self.vector_store._get_index_documents_by_ids(index_name, ids)
```
bfoley13, line 16: This addition fetches docs from the store based on ID's, which should be keyed for O(1) lookup time but will add some latency

Fei-Guo (approved):

Fei-Guo (approved):

PR #1074, opened 2025-04-28
refactor: remove fields from PresetParam in favor of embedded Metadata

chewong:
**Reason for Change**:

Fully integrates supported_models.yaml as the source of truth for model metadata, replacing hardcoded ModelFamilyName and tag maps in model presets.

**Changes:**
- Updated `PresetParam` in `pkg/model/interface.go` to embed `Metadata` struct, removing `ModelFamilyName` and `Tag`.
- Modified model files (e.g., `deepseek`, `falcon`, `llama2`, `mistral`, etc.) to use `metadata.MustGet` for metadata retrieval.
- Added constants for model names and removed tag maps in model files.
- Enhanced `presets/workspace/models/metadata.go` with `MustGet` function and mutex for thread-safe metadata access.
- Updated `api/v1alpha1` and `api/v1beta1` validation logic to use `params` variable for inference parameters.
- Adjusted `preset-tuning_test.go` to reflect `Metadata` embedding in `PresetParam`.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/workspace/models/metadata.go
+++ b/presets/workspace/models/metadata.go
@@ -16,12 +16,9 @@ var (
 	//go:embed supported_models.yaml
 	supportedModelsYAML []byte
 
-	// mutex is used to synchronize access to the SupportedModels map.
-	mutex sync.Mutex
-
-	// SupportedModels is a global map that holds the source of truth
+	// supportedModels is a map that holds the source of truth
 	// for all supported models and their metadata.
-	SupportedModels map[string]*model.Metadata
+	supportedModels sync.Map
 )
 
 // Catalog is a struct that holds a list of supported models parsed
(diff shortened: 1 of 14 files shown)


chewong:
Before changing Phi 3.5's tag in `supported_models.yaml`:

image

After bumping it to 0.1.1:

image

presets/workspace/models/metadata.go
```diff
@@ -15,6 +16,9 @@ var (
 	//go:embed supported_models.yaml
 	supportedModelsYAML []byte
 
+	// mutex is used to synchronize access to the SupportedModels map.
+	mutex sync.Mutex
+
 	// SupportedModels is a global map that holds the source of truth
 	// for all supported models and their metadata.
 	SupportedModels map[string]*model.Metadata
```
zhuangqh, line 24: Consider making `SupportedModels` a `sync.Map`. It's designed for this use case.
> (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow,

ref:  https://pkg.go.dev/sync#Map

zhuangqh (approved):

PR #1080, opened 2025-04-29
fix: document id's on list document responses are the same as document id's on create document responses

bfoley13:
**Reason for Change**:
LlamaIndex has some internal node id referencing it uses within its document stores that differ from the [document ID's we create documents with](https://github.com/kaito-project/kaito/blob/10af1f42eddbacb9e335b06bebbb4088cad7b629/presets/ragengine/vector_store/base.py#L44C1-L45C1).

This change updates the response ID's from the internal LlamaIndex nodeIds to the document ID's we create for the documents. I chose the `ref_doc_id` because it is what is stored within the vector store on document additions, so when users look into the vector store(cosmos/postgres/etc.) they will see the same doc id as is returned from the ragengine service.

I also found that the id list build on document create calls with new indexes was also returning a random ordered list by calling list() on a set. Changed this fill out lists by document indexes in the input list.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/ragengine/tests/vector_store/test_base_store.py
+++ b/presets/ragengine/tests/vector_store/test_base_store.py
@@ -124,11 +124,8 @@ class BaseVectorStoreTest(ABC):
         all_docs = await vector_store_manager.list_documents_in_index("test_add_index", limit=10, offset=1)
         print(all_docs)
 
-        for idx, doc_id in enumerate(ids):
-            # list_documents_in_index returns documents with doc_id's that are not the same as the id's returned from index_documents
-            response_doc = [doc for doc in all_docs if doc['doc_id'] == vector_store_manager.index_map['test_add_index'].docstore.get_ref_doc_info(doc_id).node_ids[0]][0] or None
-            assert response_doc is not None
-            assert documents[idx].text == response_doc['text']
+        # Validate id's from index_documents match the expected ids
+        assert all(doc['doc_id'] == ids[idx] for idx, doc in enumerate(all_docs))
 
     @pytest.mark.asyncio
     async def test_persist_index(self, vector_store_manager):
--- a/presets/ragengine/vector_store/base.py
+++ b/presets/ragengine/vector_store/base.py
@@ -214,17 +214,17 @@ class BaseVectorStore(ABC):
         """
         try:
             if isinstance(doc_store, SimpleDocumentStore):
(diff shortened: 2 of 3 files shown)


presets/ragengine/tests/vector_store/test_base_store.py
```diff
@@ -124,11 +124,8 @@ async def test_add_document_on_existing_index(self, vector_store_manager):
         all_docs = await vector_store_manager.list_documents_in_index("test_add_index", limit=10, offset=1)
         print(all_docs)
 
-        for idx, doc_id in enumerate(ids):
-            # list_documents_in_index returns documents with doc_id's that are not the same as the id's returned from index_documents
-            response_doc = [doc for doc in all_docs if doc['doc_id'] == vector_store_manager.index_map['test_add_index'].docstore.get_ref_doc_info(doc_id).node_ids[0]][0] or None
-            assert response_doc is not None
-            assert documents[idx].text == response_doc['text']
+        # Validate id's from index_documents match the expected ids
+        assert all(doc['doc_id'] == ids[idx] for idx, doc in enumerate(all_docs))
```
ishaansehgal99, line 128: this check looks good - why remove old checks?

presets/ragengine/vector_store/base.py
```diff
@@ -214,17 +214,17 @@ async def _process_document(self, doc_id: str, doc_stub, doc_store, max_text_len
...
-                text, hash_value = doc_info.text, doc_info.hash
+                text, hash_value, ref_doc_id = doc_info.text, doc_info.hash, doc_stub.ref_doc_id
 
             # Truncate if needed
             is_truncated = bool(max_text_length and len(text) > max_text_length)
             truncated_text = text[:max_text_length] if is_truncated else text
 
             return {
-                "doc_id": doc_id,
+                "doc_id": ref_doc_id,
```
ishaansehgal99, line 227: Is ref_doc_id same as doc_id?

ishaansehgal99 (approved):

presets/ragengine/tests/vector_store/test_base_store.py
bfoley13: These are equivalent checks, just with the changes made simplifies the logic. I wrote the old checks in the last PR

PR #1088, opened 2025-04-30
fix: check nodeclass before controller start

cvvz:
**Reason for Change**:

Check `nodeclass` before controller start, rather than check before creating `nodeclaim`

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo:
The old code was trying to avoid a case where user deletes the nodeclass accidentally. Since we are not adding nodeclass watcher, if we don't check nodeclass every time before creating a node and user deletes the nodeclass, the only remediation is to ask the user to delete the controller manually to force it to restart. 

So ideally, we may need to add a separate controller for nodeclass only to strictly recover it whenever it is deleted. My thought right now is that the churn rate of creating nodeclaim by workspace controller is low. Hence, the overhead of performing a nodeclass check before creating every nodeclaim is low. I would suggest leaving current behavior unchanged. It is fair to add a TODO to optimize this behavior later.

PR #1090, opened 2025-04-30
feat: add update/delete document api's and handling for faiss

bfoley13:
**Reason for Change**:
This adds update/delete document api's and handling for FAISS. This required a wrapper around the LlamaIndex FaissVectorStore and wrapping our initially used FlatL2 faiss index with a IDMap index to support delete operations.  Base functionality is still the same with the wrapped index, we just now have support for document mapping to id's in the IDMap index.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/ragengine/main.py
+++ b/presets/ragengine/main.py
@@ -6,7 +6,7 @@ from vector_store_manager.manager import VectorStoreManager
 from embedding.huggingface_local_embedding import LocalHuggingFaceEmbedding
 from embedding.remote_embedding import RemoteEmbeddingModel
 from fastapi import FastAPI, HTTPException, Query
-from models import (IndexRequest, ListDocumentsResponse,
+from models import (IndexRequest, ListDocumentsResponse, UpdateDocumentRequest,
                     QueryRequest, QueryResponse, DocumentResponse, HealthStatus)
 from vector_store.faiss_store import FaissVectorStoreHandler
 
@@ -250,6 +250,83 @@ async def list_documents_in_index(
             documents=documents,
             count=len(documents)
         )
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+
+@app.post(
+    "/indexes/{index_name}/documents/{doc_id}",
+    response_model=DocumentResponse,
+    summary="Update document in an Index by its ID",
+    description="""
+    Update document in an Index by its ID.
+
+    ## Request Example:
+    ```json
+    POST /indexes/test_index/documents/test_doc_id
+    {"text": "Sample document text.", "metadata": {"author": "John Doe", "category": "example"}}
+    ```
+
+    ## Response Example:
+    ```json
+    {
+        "doc_id": "123456",
+        "text": "Sample document text.",
+        "metadata": {"author": "John Doe"},
+        "is_truncated": true
+    },
+    ```
+    """,
+)
+async def update_document_in_index(
+    index_name: str,
+    doc_id: str,
+    request: UpdateDocumentRequest,
+):
+    try:
+        decoded_index_name = unquote(index_name)
+        decoded_doc_id = unquote(doc_id)
+        await rag_ops.update_document(
+            index_name=decoded_index_name,
+            doc_id=decoded_doc_id,
+            new_doc=request
+        )
+
+        return request
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+
+@app.delete(
+    "/indexes/{index_name}/documents/{doc_id}",
+    summary="Delete Document in an Index by its ID",
+    description="""
+    Delete Document in an Index by its ID.
+
+    ## Request Example:
+    ```
+    DELETE /indexes/test_index/documents/test_doc_id
+    ```
+
+    ## Response Example:
+    ```json
+    {
+      "message": "Successfully deleted document test_doc_id from index test_index."
+    }
+    ```
+    """,
+)
+async def delete_document_in_index(
+    index_name: str,
+    doc_id: str,
+):
+    try:
+        decoded_index_name = unquote(index_name)
+        decoded_doc_id = unquote(doc_id)
+        await rag_ops.delete_document(
+            index_name=decoded_index_name,
+            doc_id=decoded_doc_id
+        )
+
+        return {"message": f"Successfully deleted document {decoded_doc_id} from index {decoded_index_name}."}
     except Exception as e:
         raise HTTPException(status_code=500, detail=str(e))
 
--- a/presets/ragengine/models.py
+++ b/presets/ragengine/models.py
@@ -25,6 +25,13 @@ class IndexRequest(BaseModel):
     index_name: str
     documents: List[Document]
 
+class UpdateDocumentRequest(BaseModel):
+    doc_id: str
+    text: str
+    hash_value: Optional[str] = None
+    metadata: Optional[dict] = Field(default_factory=dict)
+    is_truncated: bool = False
+
 class QueryRequest(BaseModel):
     index_name: str
     query: str
--- a/presets/ragengine/tests/api/test_main.py
+++ b/presets/ragengine/tests/api/test_main.py
@@ -107,6 +107,86 @@ async def test_query_index_success(mock_get, async_client):
 @pytest.mark.asyncio
 @respx.mock
 @patch("requests.get")  # Mock the requests.get call for fetching model metadata
+async def test_document_update_success(mock_get, async_client):
+    #Mock the response for the default model fetch
+    mock_get.return_value.status_code = 200
+    mock_get.return_value.json.return_value = {
+        "data": [{"id": "mock-model", "max_model_len": 2048}]
+    }
+
+    # Mock HTTPX response for Custom Inference API
+    mock_response = {"result": "This is the completion from the API"}
+    respx.post("http://localhost:5000/v1/completions").mock(return_value=httpx.Response(200, json=mock_response))
+
+    # Index Request
+    request_data = {
+        "index_name": "test_update_index",
+        "documents": [
+            {"text": "This is a test document"},
+            {"text": "Another test document"}
+        ]
+    }
+
+    response = await async_client.post("/index", json=request_data)
+    assert response.status_code == 200
+    doc1, doc2 = response.json()
+    update_doc = doc1 if doc1["text"] == "Another test document" else doc2
+    assert update_doc["doc_id"] != ""
+
+    update_doc_id = update_doc["doc_id"]
+    update_doc["text"] = "This is an updated test document"
+    response = await async_client.post(f"/indexes/test_update_index/documents/{update_doc_id}", json=update_doc)
+    assert response.status_code == 200
+    assert response.json()["doc_id"] == update_doc_id
+    assert response.json()["text"] == "This is an updated test document"
+
+    # Query Request
+    request_data = {
+        "index_name": "test_update_index",
+        "query": "updates test query",
+        "top_k": 1,
+        "llm_params": {"temperature": 0.7}
+    }
+
+    response = await async_client.post("/query", json=request_data)
+    assert response.status_code == 200
+    assert response.json()["response"] == "{'result': 'This is the completion from the API'}"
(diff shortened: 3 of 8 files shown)


presets/ragengine/main.py
```diff
@@ -253,6 +253,83 @@ async def list_documents_in_index(
...
+
+    ## Response Example:
+    ```json
+    {
+      "message": "Successfully deleted document test_doc_id from index test_index."
+    }
+    ```
+    """,
+)
+async def delete_document_in_index(
```
bfoley13, line 317: do we want update/delete to fail on non existent document ids?

presets/ragengine/tests/api/test_main.py
```diff
@@ -104,6 +104,86 @@ async def test_query_index_success(mock_get, async_client):
     # Ensure the model fetch was called once
     mock_get.assert_called_once_with("http://localhost:5000/v1/models", headers=ANY)
 
+@pytest.mark.asyncio
+@respx.mock
+@patch("requests.get")  # Mock the requests.get call for fetching model metadata
+async def test_document_update_success(mock_get, async_client):
```
bfoley13, line 110: Working on adding more tests

presets/ragengine/main.py
```diff
@@ -253,6 +253,83 @@ async def list_documents_in_index(
...
+        decoded_doc_id = unquote(doc_id)
+        await rag_ops.update_document(
+            index_name=decoded_index_name,
+            doc_id=decoded_doc_id,
+            new_doc=request
+        )
+
+        return request
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
```
ishaansehgal99, line 296: Likely updating one or more documents at a time so something like

```suggestion
@app.put("/indexes/{index_name}/documents", response_model=Dict[str, List[str]])
async def update_documents(index_name: str, request: UpdateDocumentsRequest):
    """Batch update documents in an index."""
    try:
        return await rag_ops.update(index_name, request.documents)
    except Exception as e:
        raise HTTPException(status_code=500, detail=str(e))
```

FastAPI automatically decodes path parameters so don't need unquote

presets/ragengine/main.py
ishaansehgal99: this happens. we dont want to fail entire request, we return something like the ones we could delete and/or print error message

{
  "deleted": ["doc1", "doc2"],
  "not_found": ["doc3"]
}

ishaansehgal99, line 317: same thing one endpoint that can handle one or more ("batch") documents deletion

presets/ragengine/models.py
```diff
@@ -25,6 +25,13 @@ class IndexRequest(BaseModel):
     index_name: str
     documents: List[Document]
 
+class UpdateDocumentRequest(BaseModel):
+    doc_id: str
+    text: str
+    hash_value: Optional[str] = None
+    metadata: Optional[dict] = Field(default_factory=dict)
+    is_truncated: bool = False
```
ishaansehgal99, line 33: I'm thinking we have doc_id, text, etc. in the document class, what do you think? 

Then its 
class UpdateRequest(BaseModel):
    documents: List[Document]
    ...maybe couple other fields need be....

presets/ragengine/main.py
Fei-Guo: I think we can just pass whatever err msg returned by the storage server if it implements the delete api.

presets/ragengine/vector_store/base.py
```diff
@@ -208,6 +208,33 @@ async def add_document_to_index(self, index_name: str, document: Document, doc_i
...
+        """Common logic for deleting a document."""
+        if index_name not in self.index_map:
+            raise ValueError(f"No such index: '{index_name}' exists.")
+        if self.use_rwlock:
+            async with self.rwlock.writer_lock:
+                self.index_map[index_name].delete_ref_doc(doc_id, delete_from_docstore=True)
+                self.index_store.add_index_struct(self.index_map[index_name].index_struct)
+        else:
+            self.index_map[index_name].delete_ref_doc(doc_id, delete_from_docstore=True)
+            self.index_store.add_index_struct(self.index_map[index_name].index_struct)
```
ishaansehgal99, line 221: 1. Not sure if we have this, but could be useful to have a `document_exists` function. 

  if not self.document_exists(doc_id, index_name):
            print(f"Document with ID {doc_id} not found in index '{index_name}'. Skipping.")
            return

 2. Also iirc I don't think you need the `self.index_store.add_index_struct(self.index_map[index_name].index_struct)` isn't that just for index? but i could be wrong.
  
  3. Also need to wrap in try-catch.

    async def delete_document(self, index_name: str, doc_id: str):
        """Common logic for deleting a document."""
        if index_name not in self.index_map:
            raise ValueError(f"No such index: '{index_name}' exists.")
        
        if not self.document_exists(doc_id, index_name):
            print(f"Document with ID {doc_id} not found in index '{index_name}'. Skipping.")
            return

         try
	        if self.use_rwlock:
	            async with self.rwlock.writer_lock:
	                self.index_map[index_name].delete_ref_doc(doc_id, delete_from_docstore=True)
	                self.index_store.add_index_struct(self.index_map[index_name].index_struct)
	        else:
	            self.index_map[index_name].delete_ref_doc(doc_id, delete_from_docstore=True)
	            self.index_store.add_index_struct(self.index_map[index_name].index_struct)
          
       except NotImplementedError as e: 
	       "you will need this as iirc faiss and other dbs may not implement delete natively"
		print(f"Delete not yet implemented for Faiss index. Skipping document {doc_id}.")

        except Exception as e:
            print(f"Unable to Delete Document from the VectorStoreIndex. Skipping. Error: {e}")
	
	Maybe a finally persist? You can leave this commented for now
        finally:
            self._persist(index_name)

presets/ragengine/vector_store/faiss_store.py
```diff
@@ -17,5 +20,65 @@ def __init__(self, embed_model: BaseEmbeddingModel):
...
+        """
+        if index_name not in self.index_map:
+            raise ValueError(f"No such index: '{index_name}' exists.")
+        retrieved_doc = await self.index_map[index_name].docstore.aget_ref_doc_info(doc_id)
+        if not retrieved_doc:
+            return
+        faiss_doc_id = self.index_map[index_name].storage_context.vector_store.ref_doc_id_map.get(doc_id)
+        if faiss_doc_id is not None:
+            if self.use_rwlock:
+                async with self.rwlock.writer_lock:
```
Fei-Guo, line 47: Do we still need to lock since line 79 already takes the lock?

presets/ragengine/vector_store/base.py
```diff
@@ -208,6 +208,33 @@ async def add_document_to_index(self, index_name: str, document: Document, doc_i
...
+        
+        llama_doc = LlamaDocument(id_=doc_id, text=document.text, metadata=document.metadata, excluded_llm_metadata_keys=[key for key in document.metadata.keys()])
+
+        if self.use_rwlock:
+            async with self.rwlock.writer_lock:
+                await self.index_map[index_name].update_ref_doc(llama_doc)
+                self.index_store.add_index_struct(self.index_map[index_name].index_struct)
+        else:
+            await self.index_map[index_name].update_ref_doc(llama_doc)
+            self.index_store.add_index_struct(self.index_map[index_name].index_struct)
```
ishaansehgal99, line 236: Awesome. Same comments as above - "update is equivalent to deleting the document and then inserting it again" (https://docs.llamaindex.ai/en/stable/api_reference/indices/) so if delete for example isn't implemented neither with update.
    
mini example
        try:
            self.index_map[index_name].update_ref_doc(llama_doc)
        except NotImplementedError as e:
            print("Update is equivalent to deleting the document and then inserting it again.")
            print(f"Update not yet fully implemented for index. Skipping document {document.doc_id}. Error: {e}")
        except Exception as e:
            print(f"Unable to Update Document in the VectorStoreIndex. Skipping. Error: {e}")

presets/ragengine/vector_store/faiss_map_store.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+from typing import Any, List, cast
+
+from llama_index.core.schema import BaseNode
+from llama_index.core.bridge.pydantic import PrivateAttr
+from llama_index.vector_stores.faiss import FaissVectorStore
+
+class FaissVectorMapStore(FaissVectorStore):
```
ishaansehgal99, line 8: Oh wow this seems cool, looks like you found a way to actually get deletion/update to work with faiss! Legend. If this is experiencing issues, I would say though we can leave it unimplemented for faiss - faiss is a very simple little example other integrations may support this better

presets/ragengine/models.py
bfoley13: Love it! Will make the change

presets/ragengine/main.py
bfoley13: updated!

presets/ragengine/main.py
bfoley13: for update I went with {"updated_documents":[], "skipped_documents":[]} but in reality the api handles creating documents if they dont exist in the update call so we likely wont see skipped documents

presets/ragengine/vector_store/faiss_map_store.py
```diff
@@ -0,0 +1,137 @@
...
+from llama_index.core.vector_stores.types import (
+    VectorStoreQuery,
+    VectorStoreQueryResult,
+)
+
+class FaissVectorMapStore(FaissVectorStore):
+    # ref_doc_id_map is used to map the ref_doc_id to fiass index id
+    ref_doc_id_map: dict = {}
+    # node_id_map is used to map the faiss index id to llama_index node id
+    node_id_map: dict = {}
```
bfoley13, line 19: to get around the bug in llamaindex we can use a lookup map to keep track of their node_id mappings to faiss id's

presets/ragengine/main.py
bfoley13: updated the api to what we discussed!

bfoley13: moved to batch handling

presets/ragengine/vector_store/faiss_map_store.py
bfoley13: I'm hopeful to get that going after this gets in!

ishaansehgal99 (approved):

presets/ragengine/vector_store/faiss_map_store.py
ishaansehgal99: should these go in init so its specific to each object instance?

presets/ragengine/vector_store/faiss_map_store.py
```diff
@@ -0,0 +1,137 @@
...
+        faiss_index: Any,
+    ) -> None:
+        """Initialize params."""
+        import_err_msg = """
+            `faiss` package not found. For instructions on
+            how to install `faiss` please visit
+            https://github.com/facebookresearch/faiss/wiki/Installing-Faiss
+        """
+        try:
+            import faiss
```
ishaansehgal99, line 32: import faiss at top?

bfoley13: removed it from init

presets/ragengine/vector_store/base.py
```diff
@@ -208,6 +206,69 @@ async def add_document_to_index(self, index_name: str, document: Document, doc_i
...
+        for doc_id in doc_ids:
+            if doc_id in self.index_map[index_name].ref_doc_info:
+                found_docs.append(doc_id)
+            else:
+                not_found_docs.append(doc_id)
+
+        try:
+            if self.use_rwlock:
+                async with self.rwlock.writer_lock:
+                    await asyncio.gather(*(self.index_map[index_name].adelete_ref_doc(doc_id, delete_from_docstore=True) for doc_id in doc_ids))
```
bfoley13, line 225: checked through other vector_store implementations in llama_index and most dont implement `delete_nodes` so went for `adelete_ref_doc` so later options will have support out of the gate

ishaansehgal99 (approved):

PR #1091, opened 2025-04-30
feat: Deprecate Llama 2 Models and Add Llama3

ishaansehgal99:
**Reason for Change**:
This PR marks AccessMode field as deprecated and removes references to our now deprecated llama-2 models, as well as adds our now downloadable llama-3 model.


bangqipropel:
One question, do we need to consider if someone is running llama2 now?

presets/workspace/models/supported_models.yaml
```diff
@@ -6,34 +6,15 @@ models:
...
-    type: llama2-chat
-    runtime: llama-2
-    tag: 0.0.4
+  # TODO: Uncomment when ready to trigger llama3 build process
+  # - name: llama3.1-8b-instruct
+  #   type: text-generation
+  #   version: https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct/commit/0e9e39f249a16976918f6564b8830bc894c89659
+  #   runtime: tfs
+  #   downloadAtRuntime: true
+  #   tag: 0.0.1
```
chewong, line 15: nit: The `tag` field will not be used when `downloadAtRuntime` is enabled, as the kaito-base image will serve as the image for the inference workload.

ishaansehgal99: Yeah I realize, I'd like us to maintain tag for our own metadata sake. In next PR I'll add little check to our build pipeline to skip over building models with "downloadAtRuntime: true"

chewong (approved):

Fei-Guo (approved):

PR #1103, opened 2025-05-07
test: Add E2E test for new Llama Model

ishaansehgal99:
**Reason for Change**:
Add E2E test for new Llama Model

### DIFF
--- a/test/e2e/preset_test.go
+++ b/test/e2e/preset_test.go
@@ -867,7 +867,7 @@ var _ = Describe("Workspace Preset", func() {
 		validateWorkspaceReadiness(workspaceObj)
 	})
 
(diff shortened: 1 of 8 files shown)


test/e2e/preset_test.go
```diff
@@ -846,6 +845,26 @@ var _ = Describe("Workspace Preset", func() {
 		validateWorkspaceReadiness(workspaceObj)
 	})
 
+	It("should create a llama-3.1-8b-instruct workspace with preset public mode successfully", func() {
```
zhuangqh, line 848: The default e2e test in pr will only run these fastcheck cases. You can add a temporary label here to force run it once.

ishaansehgal99:
failed as expected due to: 
```
This preset requires a modelAccessSecret with HF_TOKEN key under presetOptions to download the model
```
Adding

PR #1105, opened 2025-05-07
docs: expand on container probes in 20250325-distributed-inference.md

chewong:
**Reason for Change**:

- Removed mentions of HuggingFace Transformer
- Updated warning when `workspace.resource.count` is larger than the required node count
- Expanded on container probes and added sequential diagrams

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/docs/proposals/20250325-distributed-inference.md
+++ b/docs/proposals/20250325-distributed-inference.md
@@ -155,7 +155,7 @@ To address these issues, the following probing strategy is proposed:
 
 The following sequential diagram summarizes the expected behavior of the liveness and readiness probes during different failure scenarios:
 
-#### Leader Pod Crash
+#### Leader Pod Crash Before and After Inference Service is Ready
 
 ```mermaid
 sequenceDiagram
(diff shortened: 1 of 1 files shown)


docs/proposals/20250325-distributed-inference.md
```diff
@@ -133,35 +132,135 @@ env:
...
+    W->>L: Reconnects to new leader (1–2 min delay)
+    Note over W: May heartbeat to old leader
+    L->>W: Waits for all workers to join
+    W->>L: All workers join
+    Note over V: Inference service becomes available again
+```
+
+#### Worker Pod Crash
+
+```mermaid
```
ishaansehgal99, line 203: Wow nice diagrams. So based on this multinode is completely fault tolerant?

ishaansehgal99 (approved):

docs/proposals/20250325-distributed-inference.md
chewong: In theory, yes. A mechanism to restart the leader container on error should be sufficient to cover most failure scenarios, but there might be other unknown scenarios I didn’t account for, which I’d need to identify during testing.

docs/proposals/20250325-distributed-inference.md
```diff
@@ -133,35 +132,135 @@ env:
...
 ...
 ```
 
 ### Liveness and Readiness Probes
 
-Standard HTTP probes are insufficient for multi-node vLLM, as only the leader pod serves the `/health` endpoint at port 5000. Probes must be adapted to verify both Ray cluster integrity (all nodes) and the leader's vLLM server readiness.
+Standard HTTP probes are insufficient for multi-node vLLM, as only the leader pod serves the /health endpoint at port 5000, and Ray cluster dynamics in Kubernetes introduce significant complexities. Key challenges include:
+
+- **Leader Dependency and Initialization**: The leader waits for all n-1 worker pods to join the Ray cluster (where n is the StatefulSet replica count) before starting tasks like model downloading and loading it to GPU memory. If a worker pod fails permanently during initialization or restarts later, the leader does not reinitialize the new worker, rendering the vLLM service unusable until the leader is restarted to recreate the Ray cluster from a clean state.
+- **Worker Rejoining Delay:** In the case of a leader restart, existing worker pods may take 1–2 minutes to recognize a new leader, as they may continue sending heartbeats to the old leader.
```
zhuangqh, line 144: From my test, it cost only 10seconds to let worker rejoin to the ray cluster after leader restarting. 
Let's make thing clearer. There are two phases of initialization. 1. ray cluster initialization. It cost about 10seconds 2. vllm application initialization. It cost couple minutes.

worker rejoin in the phase1 works, it cost couple of seconds.
worker rejoin in the phase2 don't work.

zhuangqh (approved):

zhuangqh (approved):

docs/proposals/20250325-distributed-inference.md
```diff
@@ -133,38 +132,125 @@ env:
...
+| **Leader** | Checks Ray cluster health via `python3 -c 'import ray; ray.init(); from ray._private.state import actors as get_all_actors; get_all_actors()'`, filtering by the correct job id to exclude duplicates. Ensures zero dead actors and alive actors equal StatefulSet replica count. Excludes `/health` endpoint.                                                                                                                                                                                                                                                                                                            | Checks `/health` endpoint at `$VLLM_LEADER_SVC:5000` after some initial delay.          |
```
zhuangqh, line 152: please update the health check script here as we discussed offline. We don't need to write the exact code here.
Just say we list actor through dashboard API.

docs/proposals/20250325-distributed-inference.md
```diff
@@ -133,38 +132,125 @@ env:
...
+  initialDelaySeconds: 60
   periodSeconds: 15
 readinessProbe:
   exec:
     command:
       - /bin/sh
       - -c
-      - /workspace/vllm/multi-node-health-check.sh --host $(KAITO_LEADER_SERVICE_HOST) --ray_port 6379 --vllm_port 5000
-  initialDelaySeconds: 60 # Allow time for distributed setup
+      - /workspace/vllm/multi-node-health-check.sh readiness --host $(VLLM_LEADER_SVC) --ray_port 6379 --vllm_port 5000
```
zhuangqh, line 247: update to the py code version.

Fei-Guo (approved):

PR #1114, opened 2025-05-08
feat: adding metadata filtering on list documents api

bfoley13:
**Reason for Change**:
Adding optional metadata filtering on the list documents API

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo (approved):

ishaansehgal99 (approved):

presets/ragengine/vector_store/base.py
```diff
@@ -330,6 +335,17 @@ async def list_documents_in_index(self,
...
+    async def _filter_documents(self, doc_items, metadata_filter, offset, limit):
+        """
+        Filter documents based on metadata.
+        """
+        filtered_docs = []
+        for doc_id, doc_stub in doc_items:
+            doc_metadata = getattr(doc_stub, "metadata", {})
+            if all(doc_metadata.get(k) == v for k, v in metadata_filter.items()):
+                filtered_docs.append((doc_id, doc_stub))
+        return islice(filtered_docs, offset, offset + limit)
```
ishaansehgal99, line 347: I think you may need this. iirc islice returns an iterator, not a list
```suggestion
        return list(islice(filtered_docs, offset, offset + limit))
```

ishaansehgal99: actually nvm

PR #1118, opened 2025-05-12
feat: support volume for tuning

cvvz:
**Reason for Change**:

feat: support volume for tuning

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/workspace/tuning/preset-tuning.go
```diff
@@ -256,16 +267,17 @@ func CreatePresetTuning(ctx context.Context, workspaceObj *kaitov1beta1.Workspac
...
 
-	outputImage := output.Image
-	if outputImage == "" {
-		return nil, nil, corev1.Volume{}, corev1.VolumeMount{}
+	if output.Volume != nil {
+		return nil, nil, nil, nil
 	}
 
+	outputImage := output.Image
+
```
umagnus, line 280: why we delete `if outputImage == ""` here, if image is empty, since we don't need to add a sidecar image in L202 but only need a pause image like code before

cvvz: resolved

api/v1alpha1/workspace_validation.go
```diff
@@ -234,7 +234,7 @@ func (r *DataSource) validateUpdate(old *DataSource, isTuning bool) (errs *apis.
 		errs = errs.Also(apis.ErrInvalidValue("During tuning Name field cannot be changed once set", "Name"))
 	}
 	if r.Volume != nil {
-		errs = errs.Also(apis.ErrInvalidValue("Volume support is not implemented yet", "Volume"))
+		// errs = errs.Also(apis.ErrInvalidValue("Volume support is not implemented yet", "Volume"))
```
umagnus, line 237: should delete this if for empty branch

pkg/workspace/tuning/preset-tuning.go
```diff
@@ -182,35 +182,34 @@ func CreatePresetTuning(ctx context.Context, workspaceObj *kaitov1beta1.Workspac
...
 	if imagePushSecret != nil {
 		imagePullSecrets = append(imagePullSecrets, *imagePushSecret)
 	}
 
-	pauseContainer := corev1.Container{
-		Name:            "pause",
-		Image:           "registry.k8s.io/pause:latest",
-		ImagePullPolicy: corev1.PullAlways,
-	}
-	sidecarContainers = append(sidecarContainers, pauseContainer)
```
zhuangqh, line 212: These lines of code are hard to understand. Together with the 4 outputs params of `prepareDataDestination`.

What about make them two explicit cases? for example

```
if workspaceObj.Tuning.Output.Volume != nil {
volumes = xx
volumeMount = xx
sidecarContainers  = xx
else if workspaceObj.Tuning.Output.Image != "" {
volumes = xx
volumeMount = xx
sidecarContainers  = xx
}
```

cvvz: The code change here is not very large, I tends to keep the original logic as it is, maybe we could make the refactor in another standalone pr.

cvvz: The code change here is not very large, I tends to keep the original logic as it is, maybe we could make the refactor in another standalone pr.

zhuangqh (approved):

presets/workspace/tuning/text-generation/fine_tuning.py
```diff
@@ -124,7 +125,8 @@ def on_step_end(self, args, state: TrainerState, control: TrainerControl, **kwar
 ))
 trainer.train()
 os.makedirs(ta_args.output_dir, exist_ok=True)
-trainer.save_model(ta_args.output_dir)
+# only save the adapter weights
```
ishaansehgal99, line 128: Problem with this is that if the tuning crashes there is no way to resume from where it crashed from

ishaansehgal99: However I agree that saving an entire checkpoints is cumbersome. Can you still do checkpoint but do adapter-only checkpointing? @cvvz

api/v1alpha1/workspace_validation.go
```diff
@@ -209,10 +209,6 @@ func (r *DataSource) validateCreate() (errs *apis.FieldError) {
 	if len(r.URLs) > 0 {
 		sourcesSpecified++
 	}
-	if r.Volume != nil {
-		errs = errs.Also(apis.ErrInvalidValue("Volume support is not implemented yet", "Volume"))
-		sourcesSpecified++
-	}
```
ishaansehgal99, line 215: @Fei-Guo We want this API to be changed in alpha version as well I assume? Just confirming

presets/workspace/tuning/text-generation/fine_tuning.py
cvvz: Actually, the check point wasn't used before, because it saved checkpoints and other model files in empty dir and only push the adapter to image registry, once the tuning pod completed or failed, the empty dir will be cleaned. 

If customer want to save the checkpoint in PV for resuming, they can  set `save_strategy` as "epoch", but now the behaviour is the same as before.

zhuangqh (approved):

Fei-Guo (approved):

PR #1140, opened 2025-05-20
chore: Run nvidia-device-plugin only on gpu node.

MartinForReal:
**Reason for Change**:

Run nvidia-device-plugin only on gpu node.
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
+++ b/charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
@@ -20,18 +20,7 @@ spec:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
-            - matchExpressions:
-              {{- if eq .Values.cloudProviderName "azure" }}
-              - key: kubernetes.azure.com/cluster
-                operator: Exists
-              - key: type
-                operator: NotIn
-                values:
-                - virtual-kubelet
-              {{- else if eq .Values.cloudProviderName "aws" }}
-              - key: "k8s.io/cloud-provider-aws"
-                operator: Exists
-              {{- end }}
+{{ toYaml .Values.nvDevicePluginNodeSelectorTerm | indent 12 }}
       tolerations:
         # Allow this pod to be rescheduled while the node is in "critical add-ons only" mode.
         # This, along with the annotation above marks this pod as a critical add-on.
--- a/charts/kaito/workspace/values-aws.yaml
+++ b/charts/kaito/workspace/values-aws.yaml
@@ -36,9 +36,5 @@ clusterName: "kaito"
 nvDevicePluginNodeSelectorTerm:
 # gpu node selection term works for azure gpu node
   - matchExpressions:
-    - key: kubernetes.azure.com/cluster
+    - key: "k8s.io/cloud-provider-aws"
       operator: Exists
-    - key: type
-      operator: NotIn
-      values:
-      - virtual-kubelet


charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
```diff
@@ -28,6 +28,9 @@ spec:
                 operator: NotIn
                 values:
                 - virtual-kubelet
+              - key: accelerator
```
MartinForReal, line 31:
```suggestion
 +              - key: accelerator
-+                operator: "Equal"
-+                value: "nvidia"
++                operator: In
++                values: ["nvidia"]
```

zhuangqh:
Is this behavior limited to aks node or does it apply to all of cloud vendors? 
When I search from github, https://github.com/search?q=accelerator%3A+nvidia&type=code
Possible selector may be "accelerator: nvidia-tesla-p100", "accelerator: nvidia-node", accelerator: nvidia"

charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
```diff
@@ -28,6 +28,10 @@ spec:
...
```
ishaansehgal99, line 31: I wouldn’t lean on a custom label like this unless you know that every sku, cloud (and every future cluster) will carry exactly that key/value.

MartinForReal: moved this setting to value.yaml

MartinForReal:
> Is this behavior limited to aks node or does it apply to all of cloud vendors? When I search from github, https://github.com/search?q=accelerator%3A+nvidia&type=code Possible selector may be "accelerator: nvidia-tesla-p100", "accelerator: nvidia-node", accelerator: nvidia"

changed the label selector operator to Exists

zhuangqh:
> @ishaansehgal99 @zhuangqh Do you think it is a good idea to move this config to value? Should I also add a value.yaml for aws?

I think the original implementation is more concise. There is a value named "cloudProviderName".

Also, the question is about "do every nv gpu nodes have this label?"

MartinForReal:
> > @ishaansehgal99 @zhuangqh Do you think it is a good idea to move this config to value? Should I also add a value.yaml for aws?
> 
> I think the original implementation is more concise. There is a value named "cloudProviderName".
> 
> Also, the question is about "do every nv gpu nodes have this label?"

The customer needs to manually edit ds to add the node selector. With this change applied, the node selector can be applied without additional modification. And we don't need to figure out whether the accelerator label will be applied to node.

zhuangqh:
> > > @ishaansehgal99 @zhuangqh Do you think it is a good idea to move this config to value? Should I also add a value.yaml for aws?
> > 
> > 
> > I think the original implementation is more concise. There is a value named "cloudProviderName".
> > Also, the question is about "do every nv gpu nodes have this label?"
> 
> The customer needs to manually edit ds to add the node selector. With this change applied, the node selector can be applied without additional modification. And we don't need to figure out whether the accelerator label will be applied to node.

I see. The original charts provide two out-of-box options for aws and azure. 
For more flexibility, what about in this way. 
Still keep these two options, if there is a user provided "nvDevicePluginNodeSelectorTerm". Use it.

pseudocode:
```
if "nvDevicePluginNodeSelectorTerm" is not empty:
  ".value.nvDevicePluginNodeSelectorTerm"
else if cloudprovider is aws:
  xxxxx
else if cloudprovider is azure:
  xxxxx
fi
```

zhuangqh (approved):

MartinForReal:
remove label as suggested because node which is configured by gpu-operator may not have the accelerator label.
Here the config is moved to value and customer can change the default value. no need to delete device plugin

ishaansehgal99:
I don't really get why do we need this? Current solution works no? Is there something broken.

ishaansehgal99 (approved):

PR #1149, opened 2025-05-26
[WIP] feat: enable prefix caching in RAG generation

### LINKED ISSUE #920: Enable Prefix Caching during RAG Generation
During RAG queries, repeated prefixes are common. Enable the prefix caching flag in vLLM to optimize performance: [vLLM Prefix Caching](https://docs.vllm.ai/en/latest/features/automatic_prefix_caching.html).

_Note: We don't have a reliable way to confirm if queries use vLLM. This can either be a user-configurable parameter or enabled by default with a fallback if it fails._

umagnus:
**Reason for Change**:

Set `enable-prefix-caching` to true in model.py as default so if cx use model trained by kaito, it use this feature by default
In presets, set `enable_prefix_caching` to true if cx set parameter `prefixCaching` as true, add this parameter to huggingface or custom model when cx use vLLM
WIP for more test

**Requirements**

**Issue Fixed**:

Fixes #920 

**Notes for Reviewers**:

### DIFF
--- a/api/v1alpha1/ragengine_types.go
+++ b/api/v1alpha1/ragengine_types.go
@@ -52,6 +52,9 @@ type InferenceServiceSpec struct {
 	// AccessSecret is the name of the secret that contains the service access token.
 	// +optional
 	AccessSecret string `json:"accessSecret,omitempty"`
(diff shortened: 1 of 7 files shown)


zhuangqh:
Per offline discussion, this pr is no longer needed. enable_prefix_caching is true by default.

PR #1150, opened 2025-05-27
chore: Rename model name to keep aligned with huggingface model card

MartinForReal:
**Reason for Change**:

chore: Rename model name to keep aligned with huggingface model card

It's easier for benchmark suites to fetch model config from huggingface . https://github.com/vllm-project/vllm/issues/16574
**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/workspace/models/deepseek/model.go
+++ b/presets/workspace/models/deepseek/model.go
@@ -69,7 +69,7 @@ func (*llama8b) GetInferenceParameters() *model.PresetParam {
 			},
 			VLLM: model.VLLMParam{
 				BaseCommand:    inference.DefaultVLLMCommand,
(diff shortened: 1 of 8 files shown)


MartinForReal:
image

MartinForReal:
hi @ishaansehgal99, I agree the model id is managed by kaito(it is not changed in this pr), but thmodel's name in vllm model catalog can be renamed to something that can be fetched from huggingface. As mentioned in vllm issue, benchmark suite in vllm will fetch model config and change request based on the model config. Can we rename it so that other tools can fetch model config from huggingface?

MartinForReal (commented):
the model name is not changed in workspace yaml. here only model name in vllm is changed.

PR #1155, opened 2025-05-28
test: Exclude gpu-provisioner check in case of running e2e on managed mode

helayoty:
**Reason for Change**:
E2E tests needs to pass the gpu-provisioner pod/deployment checks when running on managed mode. To add flexibility  to the e2e framework, a new go test flag `skip-gpu-provisioner-check` is introduced. default value is false which will check for gpu-provisioner. 

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -24,6 +24,7 @@ var (
 	ctx                 = context.Background()
 	namespaceName       = fmt.Sprint(utils.E2eNamespace, rand.Intn(100))
 	nodeProvisionerName = os.Getenv("TEST_SUITE")
+	deploymentMode      = os.Getenv("DEPLOYMENT_MODE")
 )
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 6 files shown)


ishaansehgal99 (approved):

test/e2e/e2e_test.go
```diff
@@ -53,7 +54,8 @@ var _ = BeforeSuite(func() {
 			Should(Succeed(), "Failed to wait for	karpenter deployment")
 	}
 
-	if nodeProvisionerName == "gpuprovisioner" {
+	if deploymentMode != "managed" &&
```
zhuangqh, line 57: what about make this "managed" a constant value

Fei-Guo:
Can we introduce a go test flag to disable the check for gpu-provisioner pod? It does not make sense to do mode check here since no one understands the environment variable.

test/e2e/e2e_test.go
helayoty: The new approach by introducing a go test flag will solve this.

Fei-Guo (approved):

PR #1175, opened 2025-06-10
feat: update rag engine image url to be build from env

bfoley13:
**Reason for Change**:
This makes it so the RAGEngine image can be built from env variables. This will allow for E2E tests to pull the correct image as well as easier image bumps in the future.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo:
Address pr-reviewer's suggestion?

bangqipropel (commented):
I see a lot of duplicate 0.3.2, maybe reuse them? Since this is not a stable one, we need to change it soon

Fei-Guo (approved):

PR #1203, opened 2025-06-20
feat: add local cache for model files

zhuangqh:
Leverage local PV to utilize GPU SKU-attached
high-performance local NVMe disks for model file caching.

- leverage csi driver https://github.com/Azure/local-csi-driver
- add script to sync charts from local-csi-driver
- add garbagecollect controller for orphan pv deletion

Signed-off-by: zhuangqh <zhuangqhc@gmail.com>


pkg/sku/azure_sku_handler.go
```diff
@@ -13,20 +13,20 @@ func NewAzureSKUHandler() CloudSKUHandler {
 		{SKU: "Standard_NC8as_T4_v3", GPUCount: 1, GPUMemGB: 16, GPUModel: "NVIDIA T4"},
 		{SKU: "Standard_NC16as_T4_v3", GPUCount: 1, GPUMemGB: 16, GPUModel: "NVIDIA T4"},
 		{SKU: "Standard_NC64as_T4_v3", GPUCount: 4, GPUMemGB: 64, GPUModel: "NVIDIA T4"},
-		{SKU: "Standard_NC24ads_A100_v4", GPUCount: 1, GPUMemGB: 80, GPUModel: "NVIDIA A100"},
-		{SKU: "Standard_NC48ads_A100_v4", GPUCount: 2, GPUMemGB: 160, GPUModel: "NVIDIA A100"},
-		{SKU: "Standard_NC96ads_A100_v4", GPUCount: 4, GPUMemGB: 320, GPUModel: "NVIDIA A100"},
+		{SKU: "Standard_NC24ads_A100_v4", GPUCount: 1, GPUMemGB: 80, GPUModel: "NVIDIA A100", NVMeDiskEnabled: true},
+		{SKU: "Standard_NC48ads_A100_v4", GPUCount: 2, GPUMemGB: 160, GPUModel: "NVIDIA A100", NVMeDiskEnabled: true},
+		{SKU: "Standard_NC96ads_A100_v4", GPUCount: 4, GPUMemGB: 320, GPUModel: "NVIDIA A100", NVMeDiskEnabled: true},
 		{SKU: "Standard_ND96asr_A100_v4", GPUCount: 8, GPUMemGB: 320, GPUModel: "NVIDIA A100"},
```
chewong, line 19: I believe this sku has NVMe disks too

pkg/workspace/controllers/workspace_gc_finalizer.go
```diff
@@ -5,7 +5,10 @@ package controllers
 
 import (
 	"context"
+	"time"
 
+	"go.uber.org/multierr"
```
chewong: consider using a similar `errors` package from Kubernetes instead - `k8s.io/apimachinery/pkg/util/errors`

presets/workspace/models/llama3/model.go
```diff
@@ -93,7 +93,7 @@ type llama3_3_70Binstruct struct{}
 func (*llama3_3_70Binstruct) GetInferenceParameters() *model.PresetParam {
 	return &model.PresetParam{
 		Metadata:                  metadata.MustGet(PresetLlama3_3_70BInstructModel),
-		DiskStorageRequirement:    "140Gi",
+		DiskStorageRequirement:    "220Gi",
```
chewong, line 96: Why 80GB more storage required?

pkg/utils/consts/consts.go
```diff
@@ -40,3 +40,7 @@ const (
 
 	NodePluginInstallTimeout = 60 * time.Second
 )
+
+var (
```
chewong, line 44:
```suggestion
const (
```

pkg/workspace/manifests/manifests.go
```diff
@@ -98,7 +98,7 @@ func GenerateServiceManifest(workspaceObj *kaitov1beta1.Workspace, serviceType c
 func GenerateStatefulSetManifest(workspaceObj *kaitov1beta1.Workspace, revisionNum string, imageName string,
```
chewong, line 98: This feature only supports StatefulSets for now?

pkg/sku/aws_sku_handler.go
```diff
@@ -13,55 +13,55 @@ func NewAwsSKUHandler() CloudSKUHandler {
...
-		{SKU: "g6.2xlarge", GPUCount: 1, GPUMemGB: 24, GPUModel: "NVIDIA L4"},
-		{SKU: "g6.4xlarge", GPUCount: 1, GPUMemGB: 24, GPUModel: "NVIDIA L4"},
-		{SKU: "g6.8xlarge", GPUCount: 1, GPUMemGB: 24, GPUModel: "NVIDIA L4"},
-		{SKU: "g6.16xlarge", GPUCount: 1, GPUMemGB: 24, GPUModel: "NVIDIA L4"},
-		{SKU: "gr6.4xlarge", GPUCount: 1, GPUMemGB: 24, GPUModel: "NVIDIA L4"},
-		{SKU: "gr6.8xlarge", GPUCount: 1, GPUMemGB: 24, GPUModel: "NVIDIA L4"},
-		{SKU: "g6.12xlarge", GPUCount: 4, GPUMemGB: 96, GPUModel: "NVIDIA L4"},
-		{SKU: "g6.24xlarge", GPUCount: 4, GPUMemGB: 96, GPUModel: "NVIDIA L4"},
-		{SKU: "g6.48xlarge", GPUCount: 8, GPUMemGB: 192, GPUModel: "NVIDIA L4"},
+		{SKU: "p4d.24xlarge", GPUCount: 8, GPUMemGB: 320, GPUModel: "NVIDIA A100", NVMeDiskEnabled: true},
```
chewong, line 16: Do we know if the CSI driver works on non-AKS clusters?

charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -58,3 +58,9 @@ rules:
     resources: ["validatingwebhookconfigurations"]
     verbs: ["update"]
     resourceNames: ["validation.workspace.kaito.sh"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["list", "get", "watch"]
```
chewong: Looks like we are missing `delete`

pkg/workspace/controllers/workspace_gc_finalizer.go
```diff
@@ -48,3 +59,101 @@ func (c *WorkspaceReconciler) garbageCollectWorkspace(ctx context.Context, wObj
...
+		return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil
+	}
+	return ctrl.Result{}, nil
+}
+
+// When a node with local volumes gets removed from a cluster before deleting those volumes,
+// the PV and PVC objects may still exist. Force deleting objects in this case.
+// Ref: https://github.com/kubernetes-csi/external-provisioner?tab=readme-ov-file#deleting-local-volumes-after-a-node-failure-or-removal
+func deleteOrphanedLocalPV(ctx context.Context, c client.Client, pvc *corev1.PersistentVolumeClaim) error {
+	nodeName := pvc.Annotations["volume.kubernetes.io/selected-node"]
```
chewong: nit: import the constant from https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-helpers/storage/volume/pv_helpers.go#L51

pkg/workspace/controllers/workspace_gc_finalizer.go
```diff
@@ -48,3 +59,101 @@ func (c *WorkspaceReconciler) garbageCollectWorkspace(ctx context.Context, wObj
...
+		pv.Finalizers = nil
+		if err := c.Update(ctx, pv); err != nil {
+			klog.ErrorS(err, "Failed to update PV finalizers", "pv", klog.KObj(pv))
+			return err
+		}
+		klog.InfoS("Removed finalizers from PV", "pv", klog.KObj(pv))
+	} else if client.IgnoreNotFound(err) != nil {
+		klog.ErrorS(err, "Failed to get PV", "pv", pvName)
+		return err
+	}
```
chewong: Can we simply delete the PVC and let the CSI driver controller clean up the PV for us?

pkg/sku/aws_sku_handler.go
zhuangqh: Yes. It works. no vendor locked in.

pkg/workspace/controllers/workspace_gc_finalizer.go
zhuangqh: No, it can't. The main reason to add this logic is that

> When a node with local volumes gets removed from a cluster before deleting those volumes, the PV and PVC objects may still exist. normal deletion of the volume and thus deletion of the PV is not possible anymore because the CSI driver instance on the node is not available

It's a special local pv csi driver which deploy in daemonset. We need to clean them manually in edge cases.
https://github.com/kubernetes-csi/external-provisioner?tab=readme-ov-file#deleting-local-volumes-after-a-node-failure-or-removal

pkg/workspace/manifests/manifests.go
zhuangqh: Yes. Because it's local pv, cannot be shared across nodes. 
If I add this to deployment, i think it would fail when scaling that deployment.

I am considering migrating all of them to statefulset.  WDTY

presets/workspace/models/llama3/model.go
zhuangqh: This size is the os disk size of the node.
It needs extra disk to place container images.

pkg/utils/consts/consts.go
zhuangqh: It's a trick here. We will get the address of this variable. Thus, it should be a var instead of const. lol

ishaansehgal99 (approved):

pkg/workspace/controllers/workspace_gc_finalizer.go
zhuangqh: done

zhuangqh:
image

zhuangqh:
comments are addressed. PTAL @chewong

pkg/model/interface.go
```diff
@@ -268,6 +268,7 @@ func (p *PresetParam) buildVLLMInferenceCommand(rc RuntimeContext) []string {
 		if revision != "" {
 			p.VLLM.ModelRunParams["code-revision"] = revision
 		}
+		p.VLLM.ModelRunParams["download_dir"] = utils.DefaultWeightsVolumePath
```
chewong:
```suggestion
		p.VLLM.ModelRunParams["download-dir"] = utils.DefaultWeightsVolumePath
```

zhuangqh:
test is good. https://github.com/kaito-project/kaito/actions/runs/15824343108/job/44600664548?pr=1203

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		return err
+	}
+	klog.InfoS("Removed finalizers from PV", "pv", klog.KObj(pv))
+
+	// Delete the PV
+	if err := c.Delete(ctx, pv); err != nil {
+		klog.ErrorS(err, "Failed to delete PV", "pv", klog.KObj(pv))
+		return client.IgnoreNotFound(err)
+	}
+	klog.InfoS("Successfully deleted PV", "pv", klog.KObj(pv))
```
chewong: Potential race condition between kube-controller-manager adding the finalizer and GC controller removing it. Would it make more sense to delete the PV first, and then remove the finalizer?

go.mod
```diff
@@ -15,12 +15,14 @@ require (
 	github.com/prometheus/client_golang v1.20.5
 	github.com/samber/lo v1.49.1
 	github.com/stretchr/testify v1.10.0
+	go.uber.org/multierr v1.11.0
```
chewong: nit: `go mod tidy` to remove unused dependency

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+	if client.IgnoreNotFound(err) != nil {
+		klog.ErrorS(err, "Failed to get node", "node", nodeName, "pvc", klog.KObj(pvc))
+		return ctrl.Result{}, err
+	}
+
+	// Node does not exist, force delete pv
+	klog.InfoS("Node does not exist, deleting orphaned PV", "pv", klog.KObj(pv), "pvc", klog.KObj(pvc), "node", nodeName)
+
+	return ctrl.Result{}, forceDeletePV(ctx, c.Client, pv)
```
chewong, line 106: We only force deleted PV here, but not PVC. Do we need to force delete PVC too?

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+			Name: pvc.Spec.VolumeName,
+		},
+	}}
+})
+
+// SetupWithManager sets up the controller with the Manager.
+func (c *PersistentVolumeGCReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	c.Recorder = mgr.GetEventRecorderFor("PersistentVolumeGCController")
+
+	builder := ctrl.NewControllerManagedBy(mgr).
```
chewong, line 170: Maybe watch for node/workspace deletion too? Without it, the controller won't go ahead and delete PV and PVC right away

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/kaito-project/kaito/pkg/utils/consts"
+)
+
+type PersistentVolumeGCReconciler struct {
+	client.Client
+	Log      logr.Logger
+	Scheme   *runtime.Scheme
```
rambohe-ch: The field `Log` and `Scheme` is unused in controller, how about remove them?

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	return result, nil
+}
+
+// When a node with local volumes gets removed from a cluster before deleting those volumes,
+// the PV and PVC objects may still exist. Force deleting objects in this case.
+// Ref: https://github.com/kubernetes-csi/external-provisioner?tab=readme-ov-file#deleting-local-volumes-after-a-node-failure-or-removal
+func (c *PersistentVolumeGCReconciler) deleteOrphanedLocalPV(ctx context.Context, pv *corev1.PersistentVolume, pvc *corev1.PersistentVolumeClaim) (ctrl.Result, error) {
+	if pvc.DeletionTimestamp.IsZero() {
```
rambohe-ch, line 83: `if pvc.DeletionTimestamp.IsZero()` --> `if !pvc.DeletionTimestamp.IsZero()`

I think we should return when pvc is terminating.

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		return result, err
+	}
+
+	return result, nil
+}
+
+// When a node with local volumes gets removed from a cluster before deleting those volumes,
+// the PV and PVC objects may still exist. Force deleting objects in this case.
+// Ref: https://github.com/kubernetes-csi/external-provisioner?tab=readme-ov-file#deleting-local-volumes-after-a-node-failure-or-removal
+func (c *PersistentVolumeGCReconciler) deleteOrphanedLocalPV(ctx context.Context, pv *corev1.PersistentVolume, pvc *corev1.PersistentVolumeClaim) (ctrl.Result, error) {
```
rambohe-ch: If we use `deleteOrphanedLocalPV` as function name, we need to move `check node exist` routine out of this function. In this function, we will delete orphaned local pv only.

pkg/workspace/controllers/garbagecollect/pv_controller.go
rambohe-ch: There is a requeue configuration for ensuring all pvs will be checked every 10 seconds.

rambohe-ch: `ctrl.Result{RequeueAfter: 10 * time.Second}`

pkg/workspace/controllers/garbagecollect/pv_controller.go
zhuangqh: Yes. you are right.

pkg/workspace/controllers/garbagecollect/pv_controller.go
zhuangqh: If pv is deleted, kcm will mark this pvc as Lost state, it will be deleted by kcm later.

rambohe-ch: agree +1, maybe delete pvc by this gc controller, and related pv will be removed by the native pv gc controller.

pkg/workspace/controllers/garbagecollect/pv_controller.go
zhuangqh: workspace deletion will add a deletiontimestamp to statefulset.
kcm will try to delete pvc as well.

When gc controller witness this pvc event,  it will keep reconcile this item until pv is well deleted.

pkg/workspace/controllers/garbagecollect/pv_controller.go
zhuangqh: np

charts/kaito/workspace/templates/local-csi-storageclass.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: local-nvme-disk
```
zhuangqh: kaito-local-nvme-disk

presets/workspace/models/llama3/model.go
```diff
@@ -93,7 +93,7 @@ type llama3_3_70Binstruct struct{}
 func (*llama3_3_70Binstruct) GetInferenceParameters() *model.PresetParam {
 	return &model.PresetParam{
 		Metadata:                  metadata.MustGet(PresetLlama3_3_70BInstructModel),
-		DiskStorageRequirement:    "140Gi",
```
zhuangqh, line 96: 220GB = 140G(model weights) + base image( 10GB) + other container in the nodes (overhead)

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	if nodeName == "" {
+		return ctrl.Result{}, nil
+	}
+
+	// Check if the node exists
+	node := &corev1.Node{}
+	err := c.Client.Get(ctx, client.ObjectKey{Name: nodeName}, node)
+	if err == nil {
+		// Node exists, waiting for csi driver to clean up
+		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
```
ishaansehgal99, line 96: consider make 10 sec constant var

ishaansehgal99 (approved):

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+	// Node does not exist, force delete pv
+	klog.InfoS("Node does not exist, deleting orphaned PV", "pv", klog.KObj(pv), "pvc", klog.KObj(pvc), "node", nodeName)
+
+	return ctrl.Result{}, forceDeletePV(ctx, c.Client, pv)
+}
+
+func forceDeletePV(ctx context.Context, c client.Client, pv *corev1.PersistentVolume) error {
+	// Remove finalizers to allow deletion
+	pv.Finalizers = nil
+	if err := c.Update(ctx, pv); err != nil {
```
rambohe-ch: we should use patch to remove finalizer. A reference for removing finalizer is here:
https://github.com/Azure/gpu-provisioner/blob/407dc0e89a92131aafcfee8851dceade109805ae/vendor/sigs.k8s.io/karpenter/pkg/controllers/node/termination/controller.go#L237

pkg/workspace/controllers/garbagecollect/pv_controller.go
```diff
@@ -0,0 +1,173 @@
...
+		klog.ErrorS(err, "Failed to update PV finalizers", "pv", klog.KObj(pv))
+		return err
+	}
+	klog.InfoS("Removed finalizers from PV", "pv", klog.KObj(pv))
+
+	// Delete the PV
+	if err := c.Delete(ctx, pv); err != nil {
```
rambohe-ch: we should delete the pv by using a force option?

chewong (approved):

charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -58,3 +58,12 @@ rules:
     resources: ["validatingwebhookconfigurations"]
     verbs: ["update"]
     resourceNames: ["validation.workspace.kaito.sh"]
+  - apiGroups: [""]
+    resources: ["persistentvolumeclaims"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["persistentvolumes"]
+    verbs: ["get", "list", "watch", "update", "delete"]
```
rambohe-ch: update --> patch?

rambohe-ch (approved):

PR #1232, opened 2025-07-02
release: update manifest and helm charts for v0.5.0

zhuangqh:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/charts/kaito/ragengine/Chart.yaml
+++ b/charts/kaito/ragengine/Chart.yaml
@@ -6,13 +6,13 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.1
+version: 0.5.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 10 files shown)


bfoley13 (approved):

charts/kaito/ragengine/Chart.yaml
```diff
@@ -6,13 +6,13 @@ type: application
...
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.1
+version: 0.5.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
-appVersion: 0.0.1
+appVersion: 0.5.0
```
helayoty, line 15: This shouldn't be updated to workspace version as it's different application with different version.

bangqipropel:
after the discussion, let's do the RAG tag as the same as others

charts/kaito/ragengine/Chart.yaml
zhuangqh: I tend to use the same tag for ragengine and workspace. Because they are in the same repo. Then we can easily find codebase based on image tag. (image tag is consistent with repo tag).

PR #1247, opened 2025-07-09
docs: update installation docs to support different cloud providers

### LINKED ISSUE #1211: docs: generic kaito installation flow
primary onboarding flow that is cloud agnostic, and not tied to azure/aks. move azure specific flow to cloud providers category

Jont828:
**Reason for Change**:

Update docs to have generalized installation flow with steps on using BYO GPU nodes and setting up GPU provisioner.

**Requirements**

**Issue Fixed**:

Fixes #1211
**Notes for Reviewers**:


website/docs/installation.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+For each provider, these are additional prerequisites:
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs groupId="provider">
+<TabItem value="azure" label="Azure" default>
+
+- [Azure CLI](https://learn.microsoft.com/cli/azure/install-azure-cli) to provision Azure resources
```
sozercan: we should not have any cloud provider here. you can think of it as installing on a kind cluster. however, in the cloud providers section we can have cloud provider specific instructions for supporting provisioning etc for specific cloud providers like azure, aws, gke, etc

sozercan (changes requested):

website/docs/installation.mdx
```diff
@@ -0,0 +1,237 @@
...
+aws eks update-kubeconfig --name $CLUSTER_NAME --region $AWS_REGION
+```
+
+</TabItem>
+</Tabs>
+
+## Install Kaito
+
+<Tabs groupId="provider">
+<TabItem value="azure" label="Azure" default>
```
chewong: These tabs are really cool 😄

Makefile
```diff
@@ -377,7 +379,6 @@ az-patch-install-ragengine-helm-e2e:
 aws-patch-install-helm:
 	yq -i '(.image.repository)                                              = "$(REGISTRY)/workspace"'                    	./charts/kaito/workspace/values.yaml
 	yq -i '(.image.tag)                                                     = "$(IMG_TAG)"'                               	./charts/kaito/workspace/values.yaml
-	yq -i '(.featureGates.Karpenter)                                    	= "true"'                                       ./charts/kaito/workspace/values.yaml
```
sozercan, line 380: fyi - I went through this with @Jont828, this feature gate does not exist. that's why we are removing it
https://github.com/kaito-project/kaito/blob/cb21ff3f967a3d5e73bc9c110856a6318fc5614f/pkg/featuregates/featuregates.go#L17

website/docs/quick-start.md
```diff
@@ -8,27 +8,67 @@ After installing KAITO, you can quickly deploy a phi-3.5-mini-instruct inference
 
 - A Kubernetes cluster with KAITO installed (see [Installation](installation))
 - `kubectl` configured to access your cluster
+- **GPU nodes available in your cluster** - You have two options:
+  - **Existing GPU nodes**: If you already have GPU nodes in your cluster, you can use the preferred nodes approach (shown below)
+  - **Auto-provisioning**: Set up automatic GPU node provisioning for your cloud provider ([Azure](azure-setup) | [AWS](aws))
+
+<!-- ## Understanding Auto-Provisioning vs Instance Types
```
sozercan: @Jont828 can you remove the leftover comments?

website/docs/installation.mdx
sozercan, line 1: this file can be removed now

Makefile
```diff
@@ -31,8 +31,10 @@ TEST_SUITE ?= gpuprovisioner
 AZURE_SUBSCRIPTION_ID ?= $(AZURE_SUBSCRIPTION_ID)
 AZURE_LOCATION ?= eastus
 AKS_K8S_VERSION ?= 1.30.0
-AZURE_RESOURCE_GROUP ?= demo
-AZURE_CLUSTER_NAME ?= kaito-demo
+CLUSTER_NAME ?= kaito-demo
```
sozercan: do we still need the changes to makefile?

website/docs/quick-start.md
```diff
@@ -8,27 +8,67 @@ After installing KAITO, you can quickly deploy a phi-3.5-mini-instruct inference
 
 - A Kubernetes cluster with KAITO installed (see [Installation](installation))
 - `kubectl` configured to access your cluster
+- **GPU nodes available in your cluster** - You have two options:
+  - **Existing GPU nodes**: If you already have GPU nodes in your cluster, you can use the preferred nodes approach (shown below)
+  - **Auto-provisioning**: Set up automatic GPU node provisioning for your cloud provider ([Azure](azure-setup) | [AWS](aws))
```
sozercan:
```suggestion
  - **Auto-provisioning**: Set up automatic GPU node provisioning for your cloud provider ([Azure](azure) | [AWS](aws))
```

website/docs/quick-start.md
```diff
@@ -100,6 +144,17 @@ kubectl run -it --rm --restart=Never curl --image=curlimages/curl -- curl -X POS
...
 🎉 Congratulations! You've successfully deployed and tested your first model with KAITO.
 
-- **Learn More**: Explore the full range of supported models in the [presets documentation](https://github.com/kaito-project/kaito/tree/main/presets)
-- **Advanced Usage**: Learn about [workspace configurations](https://github.com/kaito-project/kaito/blob/main/api/v1alpha1/workspace_types.go)
-- **Contributing**: See how to [contribute new models](https://github.com/kaito-project/kaito/blob/main/docs/How-to-add-new-models.md)
+**What's Next:**
+
+- **Explore More Models**: Check out the full range of supported models in the [presets documentation](https://github.com/kaito-project/kaito/tree/main/presets)
+- **Set Up Auto-Provisioning**: Configure automatic GPU node provisioning for your cloud provider:
+  - [Azure Setup](azure-setup) - Azure GPU Provisioner
```
sozercan:
```suggestion
  - [Azure Setup](azure) - Azure GPU Provisioner
```

website/docs/azure.md
```diff
@@ -0,0 +1,172 @@
...
+- Integrates with Azure's managed identity system for secure access
+
+### When to Use Auto-Provisioning
+
+Choose auto-provisioning when:
+- You want KAITO to manage GPU node creation automatically
+- Your workloads have varying GPU requirements
+- You prefer to specify exact Azure instance types in your workspaces
+
+Alternative: If you already have GPU nodes or manage them separately, use the [preferred nodes approach](quick-start#option-1-using-preferred-nodes-existing-gpu-nodes) instead.
```
sozercan, line 32:
```suggestion
:::note
 Alternative: If you already have GPU nodes or manage them separately, use the preferred nodes approach in the [Quick Start](quick-start) instead.
:::
```

website/docs/aws.md
```diff
@@ -1,105 +1,72 @@
...
-Check status of the Helm chart installations.
+## Create EKS Cluster and install Karpenter
 
-```bash
-helm list -n default
-```
+Follow the instructions [here](https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/) to create an EKS cluster and install Karpenter.
 
-Check status of `workspace`.
+Then update the Kaito workspace controller to the latest version that supports Karpenter integration:
```
chewong:
```suggestion
Then update the KAITO workspace controller to the latest version that supports Karpenter integration:
```

website/docs/aws.md
```diff
@@ -1,105 +1,72 @@
...
   labelSelector:
     matchLabels:
-      apps: falcon-7b
+      apps: phi-3.5-mini
 inference:
   preset:
-    name: "falcon-7b"
-
-$ kubectl apply -f $kaito_workspace_aws
+    name: phi-3.5-mini-instruct
```
chewong: let's use phi-4-mini-instruct, a newer model

```suggestion
    name: phi-4-mini-instruct
```

website/docs/azure.md
```diff
@@ -0,0 +1,172 @@
...
+metadata:
+  name: auto-provisioned-workspace
+resource:
+  instanceType: "Standard_NC6s_v3"  # Will trigger node creation
+  labelSelector:
+    matchLabels:
+      apps: phi-3.5-mini
+inference:
+  preset:
+    name: phi-3.5-mini-instruct
```
chewong: ditto

website/sidebars.js
```diff
@@ -45,6 +45,7 @@ const sidebars = {
             collapsed: false,
             items: [
                 'aws',
+                'azure',
```
chewong: Could you put the cloud providers category after `Getting Started`

website/docs/quick-start.md
```diff
@@ -8,27 +8,67 @@ After installing KAITO, you can quickly deploy a phi-3.5-mini-instruct inference
...
 
 ```yaml title="phi-3.5-workspace.yaml"
 apiVersion: kaito.sh/v1beta1
 kind: Workspace
 metadata:
   name: workspace-phi-3-5-mini
 resource:
-  instanceType: "Standard_NC24ads_A100_v4"
+  preferredNodes:
+    - gpunp-26695285-vmss000000
```
chewong: I believe the preferred node(s) needs to both:
1. Match `.resource.instanceType`
2. Contain the selector label below

website/docs/quick-start.md
```diff
@@ -8,27 +8,67 @@ After installing KAITO, you can quickly deploy a phi-3.5-mini-instruct inference
...
+Create a YAML file named `phi-3.5-workspace.yaml` with the following content:
+
 
 ```yaml title="phi-3.5-workspace.yaml"
 apiVersion: kaito.sh/v1beta1
 kind: Workspace
 metadata:
   name: workspace-phi-3-5-mini
 resource:
-  instanceType: "Standard_NC24ads_A100_v4"
```
chewong, line 22: `.resource.instanceType` is a required field, so let's not remove it

website/docs/quick-start.md
sozercan: @chewong I believe it just needs to match the name and the label. @ishaansehgal99 mentioned it uses instanceType as a fallback if not found. @ishaansehgal99 can you confirm?

website/docs/quick-start.md
sozercan: see my comment above

website/docs/quick-start.md
chewong: I can help double check

Jont828:
@sozercan @chewong Updated the PR except for the last comment on the `instanceType` field. Any word on that or if we need to change it here?

website/docs/quick-start.md
chewong: Just checked. It defaults to `Standard_NC24ads_A100_v4` if it's left empty but it might mess up the calculation done by the admission webhook where it ensures the available GPU memory from the instance type exceeds the model memory requirement.

For small models, it should be fine. For large models that require more GPU memory than `Standard_NC24ads_A100_v4`, the admission webhook will block users from creating the Workspace unless `.resource.instanceType` uses a large enough instance type.

chewong: @Jont828 could you add a note saying `instanceType` should be set to the same value as the preferred nodes (maybe advise the user to use `kubectl get node <node-name> -ojson | jq ...` to extract the value)? Other than that, the PR looks good to me

sozercan: @chewong isn't instance type a cloud provider specific thing? say, if we are running this on a kind cluster, that won't exist?

chewong: Yes it is cloud-specific, with a heavy emphasis on Azure instance types. KAITO uses it for various calculations within its webhook (https://github.com/kaito-project/kaito/blob/main/api/v1beta1/workspace_validation.go#L311-L320). Giving it the wrong value or using the default value might lead to the webhook denying requests.

For kind cluster, the only way to bypass this check is to choose an instance type from this [list](https://github.com/kaito-project/kaito/blob/8f6f312d4f6fe7ededd33b40025dcb5019c58060/pkg/sku/azure_sku_handler.go) as a fake instance type.

website/docs/azure.md
```diff
@@ -0,0 +1,174 @@
...
+  --scope /subscriptions/$SUBSCRIPTION/resourceGroups/$RESOURCE_GROUP/providers/Microsoft.ContainerService/managedClusters/$CLUSTER_NAME \
+  --role "Contributor"
+```
+
+### Step 3: Install GPU Provisioner
+
+Install the Azure GPU Provisioner using Helm:
+
+```bash
+export GPU_PROVISIONER_VERSION=0.4
```
zhuangqh: 0.3.5?

website/docs/aws.md
```diff
@@ -1,105 +1,72 @@
...
-helm list -n default
-```
+Follow the instructions [here](https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/) to create an EKS cluster and install Karpenter.
 
-Check status of `workspace`.
+Then update the KAITO workspace controller to the latest version that supports Karpenter integration:
 
 ```bash
-kubectl describe deploy kaito-workspace -n kaito-workspace
+helm update kaito-workspace --namespace kaito-workspace --set cloudProviderName=aws
```
zhuangqh: hmmm. this part is strange. Don't user already have KAITO installed in their cluster following previous steps.
KAITO supports karpenter from a very early release. We don't need this notice.

website/docs/quick-start.md
zhuangqh: https://github.com/kaito-project/kaito/tree/main/examples#gpu-resource-requirement-bypass
we have an annotation to bypass the resource check for nodes. It's good for BYO nodes.

And there is a topic about BYO nodes. https://kaito-project.github.io/kaito/docs/kaito-on-byo-gpu-nodes
what about bringing this preferrenode configuration doc there?

zhuangqh: why do you change the quickstart example here from provisioning node to preferred node?
provisioning mode is the recommended usage.

If you want to reduce cloud-specific items here, I suggest you to leave a place holder for instanceType field. And say you can find the sku type from xxxx.

website/docs/aws.md
sozercan: this is updating to add the cloudProviderName. by default, it's set to azure. @zhuangqh is that a requirement for aws, this can be removed if it's not

website/docs/quick-start.md
Fei-Guo: We 100% should not recommend preferred nodes here, it directly defeats the purpose of having a node provisioner in Kaito. If you want it to be vendor neural, add a comment: 
```
# for azure, use sku xxx
# for aws, use sku yyy
```

We already mentioned Karpenter API compatible in the architecture, hence it is ok to list the current vendors that support Karpenter for now (AWS, azure).

Jont828: @zhuangqh I changed the example to use `preferredNode` when collaborating with @sozercan.

This PR addresses #1211 which is intended to make the install/quick start steps agnostic to any cloud provider and move Azure/AWS specifics to their own page. The quick start would include the lowest common denominator between cloud providers that might not support GPU node provisioning and so we don't want to block users from trying Kaito b/c they can't install GPU provisioning.

Jont828: Based on @chewong's comment, there are some issues with the `instanceType` field with how sets a default value specific to Azure and possibly being blocked by the webhook. It seems like we might need to rework the behavior of that field, which is out of the scope of the docs. Can we open an issue to discuss that and leave a note to follow up and update the docs accordingly?

website/docs/aws.md
Jont828: I can update line 33 at least to make the sentence match what the command is doing.

website/docs/quick-start.md
chewong: To compromise, can we have three [tabs](https://docusaurus.io/docs/markdown-features/tabs) in the quickstart sections to showcase three different flavors of Workspaces:
1) Azure with sku xxx -> also add a link to /docs/azure
2) AWS with sku yyy -> also add a link to /docs/aws
3) Preferred Nodes -> also add a link to /docs/kaito-on-byo-gpu-nodes

website/docs/aws.md
Jont828: @zhuangqh Do you mean to say that it is not a requirement and can be removed then?

website/docs/quick-start.md
Jont828: It's done! I added steps detailing the option to BYO nodes or set up auto-provisioning and the pros/cons. Let me know if you have any additional concerns.

website/docs/quick-start.mdx
```diff
@@ -0,0 +1,237 @@
...
+    - gpunp-26695285-vmss000001
+  labelSelector:
+    matchLabels:
+      apps: llm-inference
+inference:
+  preset:
+    name: phi-4-mini-instruct
+```
+
+:::tip
```
Fei-Guo: I would suggest removing lines 63-78, which are exactly the same compared to line 108-119. This may confuse the reader. This just happens to work based on current implementation and is not intended from the API perspective.

website/docs/quick-start.mdx
```diff
@@ -0,0 +1,237 @@
...
+```
+
+### Option 2: Auto-provision GPU nodes
+
+The following cloud providers support auto-provisioning GPU nodes in addition to BYO nodes.
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+<Tabs groupId="provider">
```
zhuangqh, line 76: cool! I love the tab here.

website/docs/aws.md
zhuangqh: I think it's still required. let's keep it here.

zhuangqh (approved):

website/docs/quick-start.mdx
```diff
@@ -0,0 +1,237 @@
...
+
+```bash
+kubectl get workspace workspace-phi-4-mini
+```
+
+When the `WORKSPACEREADY` column becomes `True`, the model has been deployed successfully:
+
+```bash
+NAME                   INSTANCE                   RESOURCEREADY   INFERENCEREADY   JOBSTARTED   WORKSPACESUCCEEDED   AGE
+workspace-phi-4-mini   Standard_NC24ads_A100_v4   True            True                          True                 4h15m
```
sozercan, line 169:
```suggestion
workspace-phi-4-mini   Standard_NC6s_v3   True            True                          True                 4h15m
```

to keep it consistent

website/docs/quick-start.mdx
sozercan: Agreed, we can remove this tip since it's duplicated below

sozercan:
LGTM after addressing https://github.com/kaito-project/kaito/pull/1247/files#r2224227592

PR #1252, opened 2025-07-10
feat: Add Gateway API Inference Extension support

### LINKED ISSUE #1202: Inference-Aware Routing Layer
Provide an inference-aware routing layer capable of making dynamic decisions based on real-time metrics and model states and factors like KV cache, prefix-cache, workload distribution, or latency requirements.

Status: In progress
Owner: Ernest Wong

chewong:
**Reason for Change**:

Add Gateway API Inference Extension support to KAITO's preset models:
- Automatically creates an InferencePool per KAITO Workspace.
- This feature is behind a feature gate called `GatewayAPIInferenceExtension`, which is set to false by default.

**Requirements**

**Issue Fixed**:

Fixes #1202 

**Notes for Reviewers**:


pkg/utils/crd.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	gvrCache = make(map[string]*metav1.APIResourceList)
```
zhuangqh: I am feeling this cache is unnecessary. The caller wants to know whether it exists in server side now, not the outdated cache. And this call is not expensive? discoveryClient.ServerResourcesForGroupVersion
We can call it directly each time.

pkg/workspace/manifests/manifests.go
```diff
@@ -486,3 +490,290 @@ func GenerateModelPullerContainer(ctx context.Context, workspaceObj *v1beta1.Wor
...
+			Namespace: workspaceObj.Namespace,
+			Labels: map[string]string{
+				kaitov1beta1.LabelWorkspaceName: workspaceObj.Name,
+			},
+			OwnerReferences: []v1.OwnerReference{
+				*v1.NewControllerRef(workspaceObj, v1beta1.GroupVersion.WithKind("Workspace")),
+			},
+		},
+		Spec: gaiev1alpha2.InferencePoolSpec{
+			TargetPortNumber: 5000,
```
zhuangqh: inference.PortInferenceServer

chewong:
> Very detailed code, no obvious problems found.
> 
> I want to discuss about the ux here. Creating an epp for all workspaces when featuregate is enabled would be too wide. Do we need to add an annotation for workspace to explicitly enable the creation of an epp? @Fei-Guo @chewong

I am open to adding an annotation in addition to the feature gate to gate this feature

chewong:
I'll hold until https://github.com/kaito-project/kaito/pull/1284 is merged.

chewong:
Converted back to draft. Waiting for https://github.com/kaito-project/kaito/issues/1346 to be implemented.

charts/kaito/workspace/values.yaml
```diff
@@ -16,7 +16,8 @@ securityContext:
     drop:
       - "ALL"
 featureGates:
-  vLLM: "true"
+  vLLM: true
+  gatewayAPIInferenceExtension: false
```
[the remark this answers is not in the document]
chewong: It should default to false for now.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -826,6 +841,67 @@ func updateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+			errs = append(errs, err)
+		}
+
+		inferenceParam := model.GetInferenceParameters()
+		if err := resources.CheckResourceStatus(obj, c.Client, inferenceParam.ReadinessTimeout); err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return kerrors.NewAggregate(errs)
```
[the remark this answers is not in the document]
chewong: It's an anti-pattern to log and return the error at the same time

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -826,6 +841,67 @@ func updateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+				if err := c.Update(ctx, obj); err != nil {
+					errs = append(errs, err)
+				}
+			}
+			// Existing resource already matches the desired spec (oldObj); continue to next object
+			continue
+		}
+
+		if !apierrors.IsNotFound(err) {
+			errs = append(errs, err)
```
Fei-Guo: For apiserver access error, we may need to return err and stop the following checks (who are most likely to fail as well).

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -826,6 +841,67 @@ func updateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+
+	ociRepository := manifests.GenerateInferencePoolOCIRepository(wObj)
+	helmRelease, err := manifests.GenerateInferencePoolHelmRelease(wObj, isStatefulSet)
+	if err != nil {
+		return err
+	}
+
+	errs := []error{}
+	for _, obj := range []client.Object{ociRepository, helmRelease} {
+		oldObj := obj.DeepCopyObject()
```
Fei-Guo: For line 879, the obj main contains meta data such as uid. You may need to clear all unrelated meta data before performing line 880.  Otherwise, line 880 check will always fail.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -826,6 +841,67 @@ func updateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+
+		if !apierrors.IsNotFound(err) {
+			errs = append(errs, err)
+		}
+		if err := resources.CreateResource(ctx, obj, c.Client); client.IgnoreAlreadyExists(err) != nil {
+			errs = append(errs, err)
+		}
+
+		inferenceParam := model.GetInferenceParameters()
+		if err := resources.CheckResourceStatus(obj, c.Client, inferenceParam.ReadinessTimeout); err != nil {
```
Fei-Guo: ReadinessTimeout, this timeout is way too long, like 20 - 40 minutes for large model.  We should set a reasonable timeout for gateway extension deployment.

Fei-Guo: oh, I see. The endpointpicker needs to wait for the workspace pod to be ready? Is this the reason you pick the model timeout here? If so, please add a comment.

pkg/workspace/manifests/manifests.go
```diff
@@ -374,3 +378,79 @@ func GenerateModelPullerContainer(ctx context.Context, workspaceObj *v1beta1.Wor
...
+		},
+	}
+}
+
+// GenerateInferencePoolHelmRelease generates a Flux HelmRelease for the inference pool.
+func GenerateInferencePoolHelmRelease(workspaceObj *kaitov1beta1.Workspace, isStatefulSet bool) (*helmv2.HelmRelease, error) {
+	matchLabels := map[string]string{
+		kaitov1beta1.LabelWorkspaceName: workspaceObj.Name,
+	}
+	if isStatefulSet {
```
Fei-Guo, line 408: Is there a clear doc explaining this label needs to be added?

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -826,6 +841,67 @@ func updateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+//
+// How it works:
+// 1) Dry-runs preset inference generation to determine if the target workload is a StatefulSet.
+// 2) Renders a Flux OCIRepository and a HelmRelease for the InferencePool chart.
+// 3) Creates the resources if absent; updates them if the desired spec differs.
+// 4) Waits for resources to become ready using the model's inference readiness timeout.
+// 5) Aggregates and returns any errors.
+//
+// Idempotent and safe to call on every reconcile; no-op if preconditions are not met.
+func (c *WorkspaceReconciler) ensureGatewayAPIInferenceExtension(ctx context.Context, wObj *kaitov1beta1.Workspace) error {
```
Fei-Guo, line 888: Question:  I notice that the chart include the epp deployment. I assume the helmrelease cr status readiness does not depends on the readiness of this deployment. If so, you need to call checkresourcestatus for this specific deployment at the end of this call. 

A more general question, how should we deal with the case that this deployment is never ready (due to code bug etc, note that we don't own this code), should we always mark this workspace not ready?

pkg/workspace/controllers/workspace_controller.go
Fei-Guo: But ideally, the epp should depends on a service, instead of a real live endpoint.

pkg/utils/consts/consts.go
```diff
@@ -49,6 +50,22 @@ const (
...
 	NodePluginInstallTimeout = 60 * time.Second
+
+	// PortInferenceServer is the default port for the inference server.
+	PortInferenceServer = int32(5000)
+
+	// InferencePoolChartURL is the OCI registry URL for the Gateway API Inference Extension inferencepool chart.
+	InferencePoolChartURL = "oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool"
+
+	// InferencePoolChartVersion is the tag/version of the inferencepool chart to deploy.
+	// MUST KEEP IN SYNC with the version in go.mod.
```
Fei-Guo, line 62: How do we check if they are in sync or not?

pkg/utils/consts/consts.go
```diff
@@ -49,6 +50,22 @@ const (
 	GpuSkuPrefix = "Standard_N"
 
 	NodePluginInstallTimeout = 60 * time.Second
+
+	// PortInferenceServer is the default port for the inference server.
+	PortInferenceServer = int32(5000)
```
Fei-Guo, line 56: I think we must have defined this port somewhere in the code base, can you double check?

Fei-Guo (commented):
Overall a nice change! Thanks for working on this awesome feature. Detailed comments are added below.

pkg/utils/consts/consts.go
chewong: We are relying on manual check for now, but in a follow-up PR, I can make `InferencePoolChartVersion` a runtime variable set via go build's [`ldflags`](https://stackoverflow.com/questions/47509272/how-to-set-package-variable-using-ldflags-x-in-golang-build) to keep the version in sync with the one in `go.mod`

pkg/utils/consts/consts.go
chewong: There is one defined `pkg/workspace/tuning/preset_tuning.go`. Will remove it in the latest commit

pkg/workspace/controllers/workspace_controller.go
chewong: >  assume the helmrelease cr status readiness does not depends on the readiness of this deployment

Flux HelmRelease's readiness actually does depend on the resources it installs via Helm. For example, I specified a non-existing image in the HelmRelease and it would eventually timeout and the controller would mark HelmRelease as not ready.

image

> should we always mark this workspace not ready?

Since the HelmRelease is owned by Workspace, I'd say it should be treated like any other resources owned by Workspace. We should only mark Workspace as Ready when all the child resources are ready.

pkg/workspace/controllers/workspace_controller.go
chewong: The timeout is probably too huge. I can set it to be the same as the default HelmRelease readiness timeout, which is 5 minutes (https://fluxcd.io/flux/components/helm/helmreleases/#timeout)

pkg/workspace/manifests/manifests.go
chewong: Updated the comment

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -826,6 +839,93 @@ func updateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+	}
+
+	// Create or update OCIRepository
+	existingOCIRepo := &sourcev1.OCIRepository{}
+	err = resources.GetResource(ctx, ociRepository.Name, ociRepository.Namespace, c.Client, existingOCIRepo)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		if err := resources.CreateResource(ctx, ociRepository, c.Client); err != nil {
```
zhuangqh: ignore exists?

chewong: Fixed

PR #1265, opened 2025-07-15
feat: expose kaito supported models via configmap

Fei-Guo:
This will help any client to know the list of models that are supported by Kaito, instead of reading from gh repo.

### DIFF
--- a/hack/compare_model_configs.sh
+++ b/hack/compare_model_configs.sh
@@ -24,7 +24,9 @@
 #   -d, --debug       Keep temporary files for debugging
 #   -h, --help        Show this help message
 
-set -e
+set -o errexit
+set -o nounset
+set -o pipefail
 
 # Parse command line arguments
 VERBOSE=false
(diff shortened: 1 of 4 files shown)


hack/compare_model_configs.sh
```diff
@@ -0,0 +1,160 @@
+#!/bin/bash
+
```
Fei-Guo, line 2: This script is generated by copilot.

hack/compare_model_configs.sh
```diff
@@ -0,0 +1,160 @@
...
+# key in the Helm ConfigMap template, ignoring comment lines.
+#
+# Usage: ./compare_model_configs.sh [OPTIONS]
+#
+# Options:
+#   -v, --verbose     Show verbose output including normalized files
+#   -d, --debug       Keep temporary files for debugging
+#   -h, --help        Show this help message
+
+set -e
```
chewong:
```suggestion
set -o errexit
set -o nounset
set -o pipefail
```

chewong (approved):

PR #1277, opened 2025-07-16
feat: add /v1/chat/completions API

### LINKED ISSUE #1257: Add support for chat/completions API in RAGEngine
The API model of the RAGEngine should more closely resemble the [chat/completion](https://platform.openai.com/docs/api-reference/chat) spec.

Update the RAGEngine `/query` API to support chat message formats. 

**Additional context**

bfoley13:
Add OpenAI-Compatible Chat Completions API with RAG Support

  This PR implements a new /v1/chat/completions endpoint that provides OpenAI-compatible chat completion functionality with integrated RAG (Retrieval-Augmented Generation) capabilities.

  Key Features

  - OpenAI-Compatible API: Implements /v1/chat/completions endpoint matching OpenAI's chat completions specification
  - Smart RAG Integration: Automatically performs RAG when index_name is provided, otherwise passes through to the underlying LLM
  - Message Format Support: Handles system, user, and developer role messages with proper content extraction
  - Passthrough Logic: Automatically detects when to bypass RAG (tools, complex content, unsupported roles) and forwards requests directly to the LLM
  - Comprehensive Testing: Extensive test coverage including success/failure scenarios, different message types, and edge cases
  - Metrics Integration: Added Prometheus metrics for chat completion latency and request counting

  Changes

  - New API Endpoint: /v1/chat/completions with full OpenAI compatibility
  - Enhanced Models: Added ChatCompletionResponse model and message parsing utilities
  - Inference Updates: Added chat_completions_passthrough method for direct LLM forwarding
  - Metrics: New rag_chat_* metrics for monitoring chat completion performance
  - Dependencies: Added openai package for type compatibility
  - Testing: 25+ new test cases covering various scenarios and edge cases

  Usage

  {
    "index_name": "my_index",
    "model": "gpt-4",
    "messages": [
      {"role": "system", "content": "You are a helpful assistant."},
      {"role": "user", "content": "What is RAG?"}
    ],
    "temperature": 0.7,
    "max_tokens": 2048
  }

  The implementation intelligently routes requests: RAG-enabled queries when an index is specified, direct LLM passthrough for tool usage, complex content, or when no index is provided.
  
  ---
  
  Fixes #1257

### DIFF
--- a/test/rage2e/rag_test.go
+++ b/test/rage2e/rag_test.go
@@ -739,7 +739,7 @@ func createAndValidateIndexPod(ragengineObj *kaitov1alpha1.RAGEngine) (map[strin
     ]
 }'`
 	opts := PodValidationOptions{
-		PodName:            "index-pod",
+		PodName:            fmt.Sprintf("index-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: "Kaito is an operator that automates the AI/ML model inference or tuning workload in a Kubernetes cluster",
@@ -764,7 +764,7 @@ func createAndValidateUpdateDocumentPod(ragengineObj *kaitov1alpha1.RAGEngine, d
     ]
 }'`
 	opts := PodValidationOptions{
-		PodName:            "update-document-pod",
+		PodName:            fmt.Sprintf("update-document-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: `"updated_documents":[{"doc_id":"` + docID + `"`,
@@ -780,7 +780,7 @@ func createAndValidateDeleteDocumentPod(ragengineObj *kaitov1alpha1.RAGEngine, d
 -H "Content-Type: application/json" \
 -d '{"doc_ids": ["` + docID + `"]}'`
 	opts := PodValidationOptions{
-		PodName:            "delete-document-pod",
+		PodName:            fmt.Sprintf("delete-document-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: `"deleted_doc_ids":["` + docID + `"]`,
@@ -795,7 +795,7 @@ func createAndValidateDeleteIndexPod(ragengineObj *kaitov1alpha1.RAGEngine) erro
 	curlCommand := `curl -X DELETE ` + ragengineObj.ObjectMeta.Name + `:80/indexes/kaito \
 -H "Content-Type: application/json"`
 	opts := PodValidationOptions{
-		PodName:            "delete-index-pod",
+		PodName:            fmt.Sprintf("delete-index-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: "Successfully deleted index kaito",
@@ -835,7 +835,7 @@ func createAndValidateQueryPod(ragengineObj *kaitov1alpha1.RAGEngine, expectedSe
 }'`
 	}
 	opts := PodValidationOptions{
-		PodName:            "query-pod",
+		PodName:            fmt.Sprintf("query-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: expectedSearchQueries,
@@ -881,7 +881,7 @@ func createAndValidateQueryChatMessagesPod(ragengineObj *kaitov1alpha1.RAGEngine
 }'`
 	}
 	opts := PodValidationOptions{
-		PodName:            "chat-completions-pod",
+		PodName:            fmt.Sprintf("chat-completions-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: expectedSearchQueries,
@@ -895,7 +895,7 @@ func createAndValidateQueryChatMessagesPod(ragengineObj *kaitov1alpha1.RAGEngine
 func createAndValidatePersistPod(ragengineObj *kaitov1alpha1.RAGEngine, expectedPersistResult string) error {
 	curlCommand := `curl -X POST ` + ragengineObj.ObjectMeta.Name + `:80/persist/kaito`
 	opts := PodValidationOptions{
-		PodName:            "persist-pod",
+		PodName:            fmt.Sprintf("persist-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: expectedPersistResult,
@@ -909,7 +909,7 @@ func createAndValidatePersistPod(ragengineObj *kaitov1alpha1.RAGEngine, expected
 func createAndValidateLoadPod(ragengineObj *kaitov1alpha1.RAGEngine, expectedLoadResult string) error {
 	curlCommand := `curl -X POST ` + ragengineObj.ObjectMeta.Name + `:80/load/kaito?overwrite=True`
 	opts := PodValidationOptions{
-		PodName:            "load-pod",
+		PodName:            fmt.Sprintf("load-pod-%s", utils.GenerateRandomString()),
 		CurlCommand:        curlCommand,
 		Namespace:          ragengineObj.ObjectMeta.Namespace,
 		ExpectedLogContent: expectedLoadResult,
@@ -939,7 +939,12 @@ func createAndValidateAPIPod(ragengineObj *kaitov1alpha1.RAGEngine, opts PodVali
 	By(fmt.Sprintf("Creating %s", opts.PodName), func() {
 		pod := GenerateCURLPodManifest(opts.PodName, opts.CurlCommand, opts.Namespace)
 		Eventually(func() error {
-			return utils.TestingCluster.KubeClient.Create(ctx, pod, &client.CreateOptions{})
+			err := utils.TestingCluster.KubeClient.Create(ctx, pod, &client.CreateOptions{})
+			if err != nil {
+				GinkgoWriter.Printf("Failed to create pod %s: %v\n", opts.PodName, err)
+				return err
+			}
+			return nil
 		}, utils.PollTimeout, utils.PollInterval).
 			Should(Succeed(), fmt.Sprintf("Failed to create %s", opts.PodName))
 	})
@@ -953,6 +958,7 @@ func createAndValidateAPIPod(ragengineObj *kaitov1alpha1.RAGEngine, opts PodVali
 					Name:      opts.PodName,
 				}, pod)
 				if err != nil {
+					GinkgoWriter.Printf("Failed to get pod %s: %v\n", opts.PodName, err)
 					return false
 				}
 				return pod.Status.Phase == v1.PodRunning || pod.Status.Phase == v1.PodSucceeded


presets/ragengine/models.py
```diff
@@ -93,4 +99,12 @@ class QueryResponse(BaseModel):
 
 class HealthStatus(BaseModel):
     status: str
-    detail: Optional[str] = None 
+    detail: Optional[str] = None
+
+
+def messages_to_prompt(messages: List[ChatMessage]) -> str:
```
zhuangqh: This part looks strange, and I doubt whether it will work properly.

do you need to call the chat completion inference api instead of converting it to prompt string?

presets/ragengine/inference/inference.py
```diff
@@ -102,6 +108,28 @@ async def acomplete(self, prompt: str, formatted: bool = False, **kwargs: Any) -
         finally:
             # Clear params after the completion is done
             self.params = {}
+    
+    async def chat_completions_passthrough(self, chatCompletionsRequest: CompletionCreateParams, **kwargs: Any) -> ChatCompletionResponse:
+        try:
+            client = await self._get_httpx_client()
+            # Use v1/chat/completions endpoint for chat completions passthrough
```
bfoley13: Have this overwrite for the tests currently.  Need to understand what the expected behavior here is. If the LLM_INFERENCE_URL doesnt contain `/v1/chat/completions` should we error out?

bangqipropel: Azure OAI is like https://<application>.openai.azure.com/openai/deployments/<model>/chat/completions?api-version=
no v1/ here

presets/ragengine/models.py
```diff
@@ -93,4 +103,26 @@ class QueryResponse(BaseModel):
...
+    return "\n".join(string_messages)
+
+def get_message_content(message: Dict) -> str:
+    """Extract content from a ChatCompletionMessageParam."""
+    if message.get("role") == "user":
+        return message.get("content", {}).get("text", "") if isinstance(message.get("content"), dict) else message.get("content", "")
+    elif message.get("role") == "system":
+        return message.get("content", {}).get("text", "") if isinstance(message.get("content"), dict) else message.get("content", "")
+    elif message.get("role") == "developer":
+        return message.get("content", {}).get("text", "") if isinstance(message.get("content"), dict) else message.get("content", "")
```
bangqipropel: Any differences between these 3?

presets/ragengine/requirements.txt
```diff
@@ -16,6 +16,7 @@ uvicorn==0.34.2
 asyncio==3.4.3
 aiorwlock==1.5.0
 httpx==0.27.0
+openai
```
bangqipropel: version number?

bangqipropel (commented):
it is a little weird if we support "query" and "chat/completions"

since they are in totally different style in name.

And "query" can fetch the model automatically but "chat/completions" cannot.

I think maybe these should be discussed next PR

presets/ragengine/vector_store/base.py
```diff
@@ -201,6 +207,117 @@ async def query(self,
...
+
+        if request.get("tools") or request.get("functions"):
+            should_passthrough_request = True
+        
+        if request.get("index_name") and request.get("index_name") not in self.index_map:
+            raise HTTPException(status_code=404, detail=f"No such index: '{request.get('index_name')}' exists.")
+        
+        if not request.get("index_name"):
+            # If no index is specified, we cannot use the chat engine and should pass through to the LLM directly.
+            should_passthrough_request = True
```
Fei-Guo: From lines 235 > 270, there are multiple checks for passthrough the query. From efficiency perspective, if any check is true, the rest of the checks can be skipped, and we should do cheaper checks first (e.g, check for index_name).

presets/ragengine/vector_store/base.py
```diff
@@ -201,6 +207,117 @@ async def query(self,
...
+        chat_engine = self.index_map[request.get("index_name")].as_chat_engine(
+            llm=self.llm,
+            chat_mode=ChatMode.CONDENSE_PLUS_CONTEXT,
+            similarity_top_k=request.get("top_k", 5),
+            node_postprocessors=node_postprocessors,
+        )
+
+        if self.use_rwlock:
+            async with self.rwlock.reader_lock:
+                chat_result = await chat_engine.achat(prompt)
```
zhuangqh, line 310: we haven't implemented this method? achat

zhuangqh (commented):
The class `CustomLLM` we inherit in our `Inference` class have an achat api and the openai client provided by llamaIndex support that as well. Is it possible to reuse these clients instead of building from raw http request.

presets/ragengine/vector_store/base.py
bfoley13: We actually do: https://github.com/run-llama/llama_index/blob/2cfce4ef6bdb61ce11c507cb6876b318fd630e19/llama-index-core/llama_index/core/llms/custom.py#L52

The achat function on CustomLLM just calls the complete functions on our Inference class so we're covered!

presets/ragengine/vector_store/base.py
bfoley13: Swapped the cheaper checks to be first, but lines 235-261 here also perform validation, are we okay with making the call to the llm before validating all messages? At the end of the day, I believe the LLM should give us the right error to send back

presets/ragengine/requirements.txt
bfoley13: added

presets/ragengine/models.py
bfoley13: updated to add user content handling and combine the rest

presets/ragengine/inference/inference.py
bfoley13: Alright I used `/chat/completions` as a validation in the LLM_INFERENCE_URL for now then

bfoley13:
> The class `CustomLLM` we inherit in our `Inference` class have an achat api and the openai client provided by llamaIndex support that as well. Is it possible to reuse these clients instead of building from raw http request.

@zhuangqh 

The `chat` and `achat` methods on the LLM classes only reference the `messages` class within llamaindex. I did try and create a `class ChatCompletionsRequest(TypedDict)` that had the messages as a field, but sadly when using the ChatMessage class and sending the request to the backing llm I get `Object of type ChatMessage is not JSON serializable`.

The OpenAI chat completions create request is very large and there is no easy way to wrap their request class as the only thing exported is a union: 
What we need: https://github.com/openai/openai-python/blob/e6c6757553bbdb777c31d0daf5916fb9e2b47ff8/src/openai/types/chat/completion_create_params.py#L36

Whats actually exported: https://github.com/openai/openai-python/blob/e6c6757553bbdb777c31d0daf5916fb9e2b47ff8/src/openai/types/chat/completion_create_params.py#L410

The expot in init: https://github.com/openai/openai-python/blob/e6c6757553bbdb777c31d0daf5916fb9e2b47ff8/src/openai/types/chat/__init__.py#L19C1-L20C1

presets/ragengine/vector_store/base.py
```diff
@@ -201,6 +207,116 @@ async def query(self,
...
+
+        Returns:
+            ChatCompletionResponse: The response containing the generated chat completion.
+        """
+        if request.get("index_name") and request.get("index_name") not in self.index_map:
+            raise HTTPException(status_code=404, detail=f"No such index: '{request.get('index_name')}' exists.")
+
+        openai_request = None
+        last_error = None
+        for model_cls in CompletionCreateParams.__args__:
```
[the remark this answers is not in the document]
bfoley13: bad suggestion as this is a Union type not a pydantic model

test/rage2e/rag_test.go
```diff
@@ -725,131 +628,188 @@ func createAndValidateQueryPod(ragengineObj *kaitov1alpha1.RAGEngine, expectedSe
...
+			"role": "user",
+			"content": "what is kaito?"
+		}
+	],
+    "max_tokens": 50,
+    "temperature": 0
+}'`
+	}
+	opts := PodValidationOptions{
+		PodName:            "query-pod",
```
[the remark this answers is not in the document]
bfoley13: changed to chat-completions-pod

presets/ragengine/vector_store/base.py
zhuangqh: gotcha, Thank you.

presets/ragengine/vector_store/base.py
```diff
@@ -176,16 +180,21 @@ async def query(self,
                 )
             )
 
-        query_engine = self.index_map[index_name].as_query_engine(
+        query_engine = self.index_map[index_name].as_chat_engine(
```
bfoley13, line 182: The chat engine internals use validation on the LLM to check if chat/completions or completions API should be used based off the metadata.  I added handling to the metadata creation to check the URL to validate it can handle chat/completions or not. This simplified our CustomLLM code a bit

Fei-Guo (approved):

presets/ragengine/main.py
```diff
@@ -278,6 +282,81 @@ async def query_index(request: QueryRequest):
...
+    "/v1/chat/completions",
+    response_model=ChatCompletionResponse,
+    summary="OpenAI-Compatible Chat Completions API",
+    description="""
+    OpenAI-compatible chat completions endpoint with RAG capabilities.
+
+    ## Request Example:
+    ```json
+    {
+      "model": "example_index",
```
Fei-Guo: In this api document, do you need to add index_name here? This is the key for user to use this api.

presets/ragengine/vector_store/base.py
```diff
@@ -201,6 +210,134 @@ async def query(self,
...
+                model=request.get("model"),
+                choices=[{
+                    "message": {
+                        "role": "assistant",
+                        "content": chat_result.response,
+                    },
+                    "finish_reason": "stop",
+                    "index": 0
+                }],
+                source_nodes=[
```
zhuangqh, line 325: That's so cool

presets/ragengine/main.py
```diff
@@ -278,6 +282,81 @@ async def query_index(request: QueryRequest):
...
+      },
+      "source_documents": [...]
+    }
+    ```
+    """,
+)
+async def chat_completions(request: dict):
+    start_time = time.perf_counter()
+    status = STATUS_FAILURE  # Default status
+    print("trying to handle chat completion request")
```
zhuangqh: should use logger if they are not debugging only during development.

presets/ragengine/inference/inference.py
```diff
@@ -102,14 +102,91 @@ async def acomplete(self, prompt: str, formatted: bool = False, **kwargs: Any) -
...
+        except Exception as e:
+            logger.error(f"Unexpected exception in achat(): {e}")
+            raise HTTPException(status_code=500, detail=f"An unexpected error occurred: {str(e)}")
+        finally:
+            # Clear params after the completion is done
+            self.params = {}
+    
+    async def chat_completions_passthrough(self, chatCompletionsRequest: CompletionCreateParams, **kwargs: Any) -> ChatCompletionResponse:
+        try:
+            if "/chat/completions" not in LLM_INFERENCE_URL:
```
zhuangqh, line 166: This logic is a bit too coarse; we can refine it later. cc @bangqipropel

presets/ragengine/inference/inference.py
```diff
@@ -220,7 +305,7 @@ def _debug_curl_command(self, data: dict) -> None:
     @property
     def metadata(self) -> LLMMetadata:
         """Get LLM metadata."""
-        return LLMMetadata()
+        return LLMMetadata(is_chat_model=LLM_INFERENCE_URL.endswith("/chat/completions"))
```
zhuangqh: I couldn't find this field in another place. `is_chat_model`

zhuangqh: context:
azure openai: https://xxxxxx.openai.azure.com/openai/deployments/gpt-4o/chat/completions?api-version=2025-01-01-preview
vllm:  xxxxx/v1/chat/completions
huggingface: https://router.huggingface.co/featherless-ai/v1/chat/completions

if you want to check the url, you need check the `path` field in url.

presets/ragengine/inference/inference.py
bfoley13: Yeah sadly the LlamaIndex library doesnt give us all these options and response fields so I wanted the passthrough to be pretty straightforward. Whatever inference errors we get we pass back directly

presets/ragengine/inference/inference.py
bfoley13: Its deep within the `achat` call stack here: https://github.com/run-llama/llama_index/blob/83a0deceb7eb2a436f85c8483bf21d761718265f/llama-index-core/llama_index/core/llms/llm.py#L639

based off the metadata it calls chat or complete!

presets/ragengine/main.py
bfoley13: Good catch! Will add here and also follow up PR with docs to come

presets/ragengine/main.py
bfoley13: removed the print statements

presets/ragengine/inference/inference.py
bfoley13: agreed, changed to parse the url and validate against the path field

zhuangqh:
```
 Failed to create index-pod
  Expected success, but got an error:
      <*errors.StatusError | 0xc00089f5e0>: 
      pods "index-pod" already exists
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "pods \"index-pod\" already exists",
              Reason: "AlreadyExists",
              Details: {Name: "index-pod", Group: "", Kind: "pods", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 409,
          },
      }
````
@bfoley13

zhuangqh (approved):

PR #1302, opened 2025-07-23
docs: add search to website using Algolia

chewong:
**Reason for Change**:

image

image

Added the new secrets as repository secrets

image

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -57,6 +57,9 @@ jobs:
 
     - run: yarn install --frozen-lockfile
     - run: yarn build
+      env:
+        ALGOLIA_APP_ID: ${{ secrets.ALGOLIA_APP_ID }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-docs.yml
```diff
@@ -57,6 +57,9 @@ jobs:
 
     - run: yarn install --frozen-lockfile
     - run: yarn build
+      env:
```
zhuangqh, line 60: can you confirm whether these two keys will be encrypted on the frontend?

chewong: It's not encrypted but per Algolia's docs, the search API key is safe for production frontend usage: https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key

image

zhuangqh (approved):

PR #1303, opened 2025-07-23
docs: add aikit to integrations

sozercan:
**Reason for Change**:

Adds AIKit as part of integrations to extend KAITO with CPU-based inference, and increased custom model support with a variety of different formats.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/website/docs/aikit.md
+++ b/website/docs/aikit.md
@@ -1,0 +1,117 @@
+# AIKit Integration with KAITO
+
+[AIKit](https://github.com/sozercan/aikit/) provides a streamlined way to package and deploy large language models (LLMs) as container images.
+
+This document demonstrates how to integrate AIKit-built models with KAITO workspaces for efficient AI model deployment on Kubernetes, including CPU-based inference and custom model creation with a variety of supported formats, such as GGUF, GPTQ, EXL2, and more.
+
+For more detailed information about AIKit, please refer to the [AIKit documentation](https://sozercan.github.io/aikit/docs/). For any AIKit-related issues, please open an issue in the [AIKit repository](https://github.com/sozercan/aikit/issues).
+
+## Overview
+
+AIKit enables you to:
+
+- 📦 [Package AI models](https://sozercan.github.io/aikit/docs/create-images) as OCI container images with minimal configuration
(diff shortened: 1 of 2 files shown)


zhuangqh:
This feature is awesome! Let's add a e2e test for it.

website/docs/aikit.md
```diff
@@ -0,0 +1,117 @@
...
+  template:
+    spec:
+      containers:
+        - name: llama-3point2-3b
+          image: ghcr.io/sozercan/llama3.2:3b
+          args:
+            - "run"
+            - "--address=:5000"
+```
+
```
zhuangqh, line 68: could you add a hint here? users may need to check the memory requirement before they deploy a cpu model and add these requirements to the spec.

OOM is the most common problem we met during deploying a llm model.

website/docs/aikit.md
```diff
@@ -0,0 +1,117 @@
...
+```bash
+export IMAGE_NAME="your-registry/your-model:latest"
+
+docker buildx build -t $IMAGE_NAME --push \
+    --build-arg="model=huggingface://TheBloke/Llama-2-7B-Chat-GGUF/llama-2-7b-chat.Q4_K_M.gguf" \
+    "https://raw.githubusercontent.com/sozercan/aikit/main/models/aikitfile.yaml"
+```
+
+After building the image, you can use it in your KAITO workspace configuration by updating the `image` field.
+
```
Fei-Guo, line 128: Can you also call out the limitations here. What are supported runtimes or anything that works in local AI but not included in this integration.

website/docs/aikit.md
sozercan: thanks, added a note. it is hard to get exact requirements, but i added general guidelines, and a way to get more accurate estimation.

website/docs/aikit.md
sozercan: thanks, added a note at the end.

sozercan:
@zhuangqh thanks! we can follow up with e2e after docs pr is merged, wdyt?

chewong (approved):

zhuangqh (approved):
LGTM

Fei-Guo (approved):

Issue #1197, opened 2025-06-17
docs: add search to website


sozercan:
website needs search integration setup through algolia

https://docusaurus.io/docs/search#using-algolia-docsearch

example: https://github.com/project-copacetic/copacetic/pull/1082

blazethunderstorm:
@sozercan can I work on this issue

sozercan:
@blazethunderstorm thanks for your interest! this might be good for one of the maintainers to do it since it requires credentials.

PR #1328, opened 2025-07-29
docs: verify docs site with Algolia

chewong:
**Reason for Change**:

Currently, the search function in the docs site is not working because Algolia wasn't able to verify the ownership of the site domain.

image

Per https://www.algolia.com/doc/tools/crawler/getting-started/create-crawler/#verify-your-domain, we can verify the domain by adding a `<meta>` HTML tag with specific content. I've added the site verification payload as a repository secret:

image

**Testing**

image

**Issue Fixed**:

Follow-up for #1197

**Notes for Reviewers**:

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -60,6 +60,7 @@ jobs:
       env:
         ALGOLIA_APP_ID: ${{ secrets.ALGOLIA_APP_ID }}
         ALGOLIA_API_KEY: ${{ secrets.ALGOLIA_API_KEY }}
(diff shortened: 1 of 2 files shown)


zhuangqh (approved):

Fei-Guo:
A side note, do you think if it is a good idea to provide a search functionality using a Kaito inference server?

chewong:
Successfully verified the site locally:

image



--- 

> A side note, do you think if it is a good idea to provide a search functionality using a Kaito inference server?

@Fei-Guo Yes I think it's a good idea. If we have an expert on frontend development, we can try to do it with https://docusaurus.io/docs/markdown-features/react

PR #1333, opened 2025-07-30
ci: add pipeline to generate versioned docs for each minor version

### LINKED ISSUE #1288: Automate versioned docs generation as part of release pipeline

Jont828:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

Fixes #1288 
**Notes for Reviewers**:

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -6,8 +6,6 @@ on:
     paths:
     - '.github/workflows/deploy-docs.yaml'
     - 'website/**'
-    tags:
-    - 'v[0-9]+.[0-9]+.0'
   pull_request:
     branches:
     - main
@@ -58,40 +56,6 @@ jobs:
           ${{ runner.os }}-website-
 
     - run: yarn install --frozen-lockfile
-    
-    - name: Create docs version for release
-      if: startsWith(github.ref, 'refs/tags/v') && github.event_name == 'push'
-      run: |
-        # Extract version from git tag (e.g., v0.5.0 -> 0.5)
-        TAG_VERSION=${GITHUB_REF#refs/tags/v}
-        MINOR_VERSION=$(echo $TAG_VERSION | cut -d. -f1-2)
-        VERSION_NAME="v${MINOR_VERSION}.x"
-        
-        echo "Detected new minor release tag: $TAG_VERSION"
-        echo "Creating docs version: $VERSION_NAME"
-        
-        # Check if this version already exists
-        if [ -d "versioned_docs/version-$VERSION_NAME" ]; then
-          echo "Version $VERSION_NAME already exists, skipping creation"
-        else
-          echo "Creating new docs version: $VERSION_NAME"
-          npm run docusaurus docs:version $VERSION_NAME
-          echo "Created new version $VERSION_NAME"
-          
-          # Configure git for committing
-          git config --local user.email "action@github.com"
-          git config --local user.name "GitHub Action"
-          
-          # Add and commit the versioned docs
-          git add versioned_docs/ versioned_sidebars/ versions.json
-          if git diff --staged --quiet; then
-            echo "No changes to commit"
-          else
-            git commit -m "docs: create version $VERSION_NAME for release $TAG_VERSION"
-            git push origin main
-          fi
-        fi
-    
     - run: yarn build
       env:
         ALGOLIA_APP_ID: ${{ secrets.ALGOLIA_APP_ID }}


.github/workflows/deploy-docs.yml
```diff
@@ -6,6 +6,8 @@ on:
     paths:
     - '.github/workflows/deploy-docs.yml'
     - 'website/**'
+    tags:
+    - 'v[0-9]+.[0-9]+.[0-9]+*'
```
chewong: What's the reason for triggering this action for any tag creation? Assuming the versioned doc already exists (created by `docs-versioning.yml`) and people commit doc changes to the right place, we shouldn't need to remove and regenerate the docs right?

.github/workflows/docs-versioning.yml
```diff
@@ -0,0 +1,171 @@
...
+
+      - name: Extract version from tag
+        id: get-version
+        run: |
+          # Extract version from tag (e.g., v0.6.0 -> 0.6.x)
+          VERSION_TAG=${GITHUB_REF#refs/tags/}
+          echo "VERSION_TAG=$VERSION_TAG" >> $GITHUB_OUTPUT
+          
+          # Convert v0.6.0 to 0.6.x format for documentation versioning
+          DOC_VERSION=$(echo $VERSION_TAG | sed -E 's/^v([0-9]+)\.([0-9]+)\.0$/\1.\2.x/')
```
chewong: nit: let's keep the v prefix


```suggestion
          DOC_VERSION=$(echo $VERSION_TAG | sed -E 's/^v([0-9]+)\.([0-9]+)\.0$/v\1.\2.x/')
```

.github/workflows/docs-versioning.yml
```diff
@@ -0,0 +1,171 @@
...
+          Release tag: \`${{ steps.get-version.outputs.VERSION_TAG }}\`
+          
+          ## What's next?
+          After merging this PR, the documentation for version ${{ steps.get-version.outputs.DOC_VERSION }} will be available at:
+          - Latest: \`/docs/\` (if this is the latest version)
+          - Versioned: \`/docs/${{ steps.get-version.outputs.DOC_VERSION }}/\`
+          
+          The version will also appear in the documentation version dropdown." \
+            --head "$BRANCH_NAME" \
+            --base main
```
chewong: Consider using an existing Action to perform PR creation: https://github.com/peter-evans/create-pull-request.

Example PR it generated: https://github.com/open-policy-agent/gatekeeper/pull/4068

chewong:
Looks promising! https://github.com/chewong/kaito/pull/1/files

.github/workflows/deploy-docs.yml
chewong: Resolved offline

chewong (changes requested):
Could you resolve the merge conflicts?

.github/workflows/deploy-docs.yaml
chewong, line 1: I don't think we need to trigger this for every patch version anymore

.github/workflows/docs-versioning.yaml
```diff
@@ -0,0 +1,229 @@
+name: Create Versioned Documentation
+
+on:
+  push:
+    tags:
+      # Trigger on new minor version tags (e.g., v0.6.0, v0.7.0, v1.0.0) and patch releases (e.g., v0.6.1, v0.6.2)
+      # This regex matches tags like v0.6.0, v1.0.0, v0.5.1, v0.5.2
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
chewong: Let's only trigger this action with new major or minor versions only

.github/workflows/docs-versioning.yaml
```diff
@@ -0,0 +1,229 @@
+name: Create Versioned Documentation
+
+on:
+  push:
+    tags:
+      # Trigger on new minor version tags (e.g., v0.6.0, v0.7.0, v1.0.0) and patch releases (e.g., v0.6.1, v0.6.2)
```
chewong:
```suggestion
      # Trigger on new minor version tags (e.g., v0.6.0, v0.7.0, v1.0.0)
```

.github/workflows/docs-versioning.yaml
```diff
@@ -0,0 +1,229 @@
...
+      # This regex matches tags like v0.6.0, v1.0.0, v0.5.1, v0.5.2
```
chewong:
```suggestion
      # This regex matches tags like v0.6.0, v1.0.0
```

PR #1337, opened 2025-07-30
feat: add a feature gate to disable node auto-provisioning and validate the preferredNodes

### LINKED ISSUE #1171: Why do we need to label nodes when we already provide the node name in preferredNode?
This feature request addresses a user experience issue. While I understand this relates to Kubernetes' native orchestration, the current behavior creates a confusing user experience.

The current implementation also leads to poor error handling: if a customer mislabels a node, the controller still creates a nodeclaim but eventually times out because no controller handles the nodeclaim. This timeout error is disconnected from the actual root cause (the labeling error), making debugging difficult.
1. Have the workspace controller automatically label the node when a preferredNode is specified.
2. If option 1 is not feasible, at minimum return a "no node found" error immediately when a preferredNode is set but no matching labeled node exists.

**Additional context**

Jont828:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

Fixes #1171

**Notes for Reviewers**:

### DIFF
--- a/pkg/workspace/controllers/workspace_controller.go
+++ b/pkg/workspace/controllers/workspace_controller.go
@@ -493,10 +493,8 @@ func (c *WorkspaceReconciler) getAllQualifiedNodes(ctx context.Context, wObj *ka
 				}
 
 				// match the instanceType
-				if len(wObj.Resource.PreferredNodes) == 0 { // don't match in perferred nodes mode
-					if node.Labels[corev1.LabelInstanceTypeStable] == wObj.Resource.InstanceType {
-						qualifiedNodes = append(qualifiedNodes, lo.ToPtr(node))
-					}
+				if node.Labels[corev1.LabelInstanceTypeStable] == wObj.Resource.InstanceType {
+					qualifiedNodes = append(qualifiedNodes, lo.ToPtr(node))
 				}
 			}
 		}


Jont828:
I managed to get PR #1337 :)

api/v1beta1/workspace_validation.go
```diff
@@ -292,13 +293,31 @@ func (r *DataDestination) validateUpdate() (errs *apis.FieldError) {
...
 
 func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, bypassResourceChecks bool, runtime model.RuntimeName) (errs *apis.FieldError) {
+	// Check if node auto-provisioning is disabled and validate preferred nodes
+	if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+		if len(r.PreferredNodes) != *r.Count {
+			errs = errs.Also(apis.ErrInvalidValue(
+				fmt.Sprintf("When node auto-provisioning is disabled, the number of preferred nodes (%d) must match the count (%d)",
+					len(r.PreferredNodes), *r.Count), "preferredNodes"))
+		}
+	}
```
chewong: Could you add a unit test case that covers this?

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -360,6 +361,11 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
 	newNodesCount := lo.FromPtr(wObj.Resource.Count) - len(selectedNodes)
 
 	if newNodesCount > 0 {
+		// Check if node auto-provisioning is disabled
+		if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+			return fmt.Errorf("node auto-provisioning is disabled but insufficient nodes available: need %d nodes, have %d selected nodes", lo.FromPtr(wObj.Resource.Count), len(selectedNodes))
+		}
```
chewong, line 367: Unit test 😄

chewong (commented):
LGTM, pending unit tests

api/v1beta1/workspace_validation.go
```diff
@@ -292,13 +293,31 @@ func (r *DataDestination) validateUpdate() (errs *apis.FieldError) {
...
 
 func (r *ResourceSpec) validateCreateWithTuning(tuning *TuningSpec) (errs *apis.FieldError) {
+	// Check if node auto-provisioning is disabled and validate preferred nodes
+	if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+		if len(r.PreferredNodes) != *r.Count {
+			errs = errs.Also(apis.ErrInvalidValue(
+				fmt.Sprintf("When node auto-provisioning is disabled, the number of preferred nodes (%d) must match the count (%d)",
+					len(r.PreferredNodes), *r.Count), "preferredNodes"))
+		}
+	}
```
chewong, line 303: We should also have the same logic to check `len(r.PreferredNodes)` in `func (r *ResourceSpec) validateUpdate` because preferred nodes field is not immutable. Might be worth creating a dedicated function so we don't have to duplicate this block of code 3 times

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -463,6 +482,15 @@ func (c *WorkspaceReconciler) getAllQualifiedNodes(ctx context.Context, wObj *ka
 		}
 	}
 
+	// After looping, check for missing preferred nodes if feature gate is enabled
+	if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+		// Since we remove a preferredNode from the preferredNodeSet if we've seen it in the list of nodes matching the label,
+		// anything node left in the set means it does not match the label selector and we can surface an error.
+		if len(deletingPreferredNodes)+len(notReadyNodes)+preferredNodeSet.Len() > 0 {
+			return nil, fmt.Errorf("when node auto-provisioning is disabled, all preferred nodes must be ready, running, and match the label selector. The following nodes do not meet the required conditions: deleting nodes: %+v, not ready nodes: %+v, nodes missing label: %+v", deletingPreferredNodes, notReadyNodes, preferredNodeSet.UnsortedList())
```
chewong: nit: preferredNodeSet could contain both nodes with no labels or nodes that don't exist at all

```suggestion
			return nil, fmt.Errorf("when node auto-provisioning is disabled, all preferred nodes must be ready, running, and match the label selector. The following nodes do not meet the required conditions: deleting nodes: %+v, not ready nodes: %+v, nodes missing label or do not exist: %+v", deletingPreferredNodes, notReadyNodes, preferredNodeSet.UnsortedList())
```

chewong (changes requested):

api/v1beta1/workspace_validation.go
chewong: We should also have this same check in `func (r *ResourceSpec) validateUpdate` because preferred nodes field is not immutable. Might be worth creating a dedicated function so we don't have to duplicate this block of code 3 times

Jont828: I think we could just move this to the top of the validate function actually if they're all referencing the same field.

chewong:
Unit test is still failing

chewong (approved):

chewong:
@Fei-Guo @zhuangqh could you take a look?

pkg/workspace/controllers/workspace_controller_test.go
```diff
@@ -853,9 +1049,9 @@ func TestApplyWorkspaceResource(t *testing.T) {
 			}
 			ctx := context.Background()
 
-			err := reconciler.applyWorkspaceResource(ctx, &tc.workspace)
+			err := reconciler.applyWorkspaceResource(ctx, &tc.workspace) // TODO: fix this nil pointer error
```
chewong: Extra comment?

pkg/workspace/controllers/workspace_controller_test.go
```diff
@@ -696,14 +886,20 @@ func TestGetAllQualifiedNodes(t *testing.T) {
 			nodes, err := reconciler.getAllQualifiedNodes(ctx, tc.workspace)
 
 			if tc.expectedError != nil {
-				assert.Equal(t, tc.expectedError.Error(), err.Error())
-				assert.Check(t, nodes == nil, "Response node array should be nil")
+				fmt.Printf("Nodes are %v\n", nodes)
```
chewong: Ditto. Let's remove all debug statements

api/v1alpha1/workspace_validation.go
```diff
@@ -57,6 +58,15 @@ func (w *Workspace) Validate(ctx context.Context) (errs *apis.FieldError) {
 	if len(errmsgs) > 0 {
 		errs = errs.Also(apis.ErrInvalidValue(strings.Join(errmsgs, ", "), "name"))
 	}
+	// Check if node auto-provisioning is disabled and validate preferred nodes
+	if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+		if len(w.Resource.PreferredNodes) != *w.Resource.Count {
```
chewong:
```suggestion
		if len(w.Resource.PreferredNodes) < *w.Resource.Count {
```

pkg/workspace/controllers/workspace_controller_test.go
```diff
@@ -676,14 +688,192 @@ func TestGetAllQualifiedNodes(t *testing.T) {
...
+							}},
+						},
+					},
+				}}
+				// relevantMap := c.CreateMapWithType(nodeList)
+				// for _, obj := range nodeList.Items {
+				// 	n := obj
+				// 	objKey := client.ObjectKeyFromObject(&n)
+				// 	relevantMap[objKey] = &n
+				// }
```
chewong: nit: Remove commented code

api/v1alpha1/workspace_validation.go
chewong: I think we also stopped modifying anything in `api/v1alpha1` so it should be fine to discard any changes in this file

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -419,46 +425,79 @@ func (c *WorkspaceReconciler) applyWorkspaceResource(ctx context.Context, wObj *
...
+			node := nodeList.Items[index]
+			if nodeIsReadyAndNotDeleting(node) {
+				// match the preferred node
+				if preferredNodeSet.Has(node.Name) {
+					qualifiedNodes = append(qualifiedNodes, lo.ToPtr(node))
+					continue
+				}
+
+				// match the instanceType
+				if len(wObj.Resource.PreferredNodes) == 0 { // don't match in perferred nodes mode
```
Fei-Guo: I think this check is not need, I understand it comes from the existing code, but the logic is wrong. We already exclude preferred node for sku check in line 496. Can you remove it?

Jont828:
@chewong @Fei-Guo are we good to merge this?

zhuangqh (changes requested):
When NAP is disabled, we can still use nodes listed from `wObj.Resource.LabelSelector.MatchLabels` ?

Jont828:
@zhuangqh Can we make that a follow up? It seems like there's interest in rethinking how the `preferredNodes` field works, so I'd like to consolidate that disucssion in a separate issue or thread. The goal of this PR is to surface an error when NAP isn't set up and you don't have enough preferred nodes that are ready or match the label selector.

zhuangqh:
> @zhuangqh Can we make that a follow up? It seems like there's interest in rethinking how the `preferredNodes` field works, so I'd like to consolidate that disucssion in a separate issue or thread. The goal of this PR is to surface an error when NAP isn't set up and you don't have enough preferred nodes that are ready or match the label selector.

Sounds good. Thank you.

Issue #1346, opened 2025-08-04
Create HelmReleases as part of Workspace Reconciliation Loop


chewong:
**Is your feature request related to a problem? Please describe.**

As we add more integration to KAITO's Workspace to improve performance of the inference workload, we find it hard to manage all the additional Kubernetes resources that comes with it. We need a generic way to bundle a group of YAMLs and the ability to perform simple customizations for each Workspace.

**Describe the solution you'd like**

1. Install Flux's Source Controller and Helm Controller (don't create other components to reduce footprint) from https://github.com/fluxcd-community/helm-charts/tree/main/charts/flux2 as KAITO Helm chart's dependency. See https://helm.sh/docs/chart_best_practices/dependencies/ on how to define Helm chart dependency.
2. Use `.values.featureGates.GatewayAPIInferenceExtension` from https://github.com/kaito-project/kaito/pull/1252 as a boolean [condition](https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags) to determine whether to install Flux or not.
3. Add Flux's scheme to https://github.com/kaito-project/kaito/blob/main/cmd/workspace/main.go#L66-L75

**Describe alternatives you've considered**

https://docs.k3s.io/helm

**Additional context**

Jont828:
I can work on this!

chewong:
Fixed by #1363

PR #1363, opened 2025-08-06
feat: add Flux Helm controller as optional dependency in Helm chart

Jont828:
**Reason for Change**:

Adds a dependency on the Flux Helm controller in the Kaito Helm chart that can be enabled/disabled with a feature flag. This helps unblock #1252 by allowing us to leverage the HelmRelease CRD to install a large number of resources through a Helm chart instead of manually creating them individually with client calls.

**Requirements**

**Issue Fixed**:

#1346
**Notes for Reviewers**:

### DIFF
--- a/charts/kaito/workspace/Chart.yaml
+++ b/charts/kaito/workspace/Chart.yaml
@@ -36,4 +36,4 @@ dependencies:
 - name: flux2
   version: "2.16.3"
   repository: https://fluxcd-community.github.io/helm-charts
-  condition: false
+  condition: flux2.enabled
--- a/charts/kaito/workspace/values.yaml
+++ b/charts/kaito/workspace/values.yaml
@@ -38,6 +38,7 @@ cloudProviderName: "azure"
 clusterName: "kaito"
 # Set values for the Flux controller to only install the Helm controller
 flux2:
+  enabled: false
   helmController:
     create: false
   sourceController:
--- a/charts/kaito/workspace/Chart.lock
+++ b/charts/kaito/workspace/Chart.lock
@@ -2,5 +2,5 @@ dependencies:
 - name: flux2
   repository: https://fluxcd-community.github.io/helm-charts
   version: 2.16.3
-digest: sha256:5f483e03127c02e0494ac5890d386b5fa1dffe7db335a8d974a4d2b279b9dcc3
-generated: "2025-08-06T16:18:20.927671241-04:00"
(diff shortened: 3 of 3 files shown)


charts/kaito/workspace/values.yaml
```diff
@@ -16,7 +16,7 @@ securityContext:
     drop:
       - "ALL"
 featureGates:
-  vLLM: "true"
+  vLLM: true
```
chewong: Could you use the `gatewayAPIInferenceExtension` feature gate from https://github.com/kaito-project/kaito/pull/1252/files#diff-b591626b039f6bb7fe8aebf16f6f940e240dbc580024668ca2b05732cbba5081R36 instead? It can default to false for now with minor code change in https://github.com/kaito-project/kaito/blob/main/pkg/featuregates/featuregates.go.

charts/kaito/workspace/values.yaml
chewong, line 1: Could you customize the [helm values of flux2](https://github.com/fluxcd-community/helm-charts/tree/main/charts/flux2) so we only install source controller (for HelmRepository) and helm controller (for HelmRelease)?

In KAITO's values.yaml, you can add:

```yaml
flux2:
  <dependency values...>
````

**Review by chewong — CHANGES_REQUESTED** [review]

File: charts/kaito/workspace/values.yaml (hunk 1 above)
Comment (chewong) [reply charts/kaito/workspace/values.yaml]: There is also a semi-breaking change here: when we convert the value from type string to boolean, we need to modify https://github.com/kaito-project/kaito/blob/main/charts/kaito/workspace/templates/deployment.yaml#L36 to accept boolean

**Review by chewong — COMMENTED** [review]
Could you modify/add a GitHub Action that runs `helm templates` (dry-run to make sure there is no error) + `helm dependency update` to make sure the helm chart is up-to-date?

File: charts/kaito/workspace/Chart.yaml [hunk 2]
```diff
@@ -36,4 +36,4 @@ dependencies:
 - name: flux2
   version: "2.16.3"
   repository: https://fluxcd-community.github.io/helm-charts
-  condition: featureGates.vLLM
+  condition: featureGates.GatewayAPIInferenceExtension
```
chewong:
```suggestion
  condition: featureGates.gatewayAPIInferenceExtension
```

Fei-Guo (approved):

Issue #1354, opened 2025-08-05
Add gpt-oss-20b and gpt-oss-120b as KAITO Presets


chewong:
**Is your feature request related to a problem? Please describe.**

https://openai.com/open-models/

According to https://huggingface.co/openai/gpt-oss-120b, we need to use a pre-released version of vLLM so we might need to wait until the official release of 0.10.1.

```
uv pip install --pre vllm==0.10.1+gptoss \
    --extra-index-url https://wheels.vllm.ai/gpt-oss/ \
    --extra-index-url https://download.pytorch.org/whl/nightly/cu128 \
    --index-strategy unsafe-best-match

vllm serve openai/gpt-oss-120b
```

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

sdesai345:
While vLLM v0.10.1 awaits release, let's investigate an upgrade to HF Transformers v4.55 to support gpt-oss as presets.

sdesai345:
Update: https://github.com/vllm-project/vllm/releases/tag/v0.10.1 pre-release is out

PR #1373, opened 2025-08-07
chore: upgrade vllm and transformers to the latest stable versions

chewong:
**Reason for Change**:

Upgrade vllm and transformers to the latest stable versions as a prerequisite to #1354.

- Transformers 4.55.0 conflicts with vLLM 0.9.0 so we are forced to upgrade to 0.10.0. We have to open a follow-up PR to upgrade vLLM to 0.10.1 anyway, which would start supporting models in #1354.
- Added `"max_new_tokens": None` in `presets/workspace/inference/text-generation/tests/test_inference_api.py` `generate_kwargs` since max_new_tokens (default 200) takes precedence over `max_length`, causing the test to generate more tokens than it supposed to.

TODO after merging this PR:
- Deprecate T4 skus which is not supported with vLLM V1

TODO when upgrading vLLM to 0.11.0 (v0 engine will be completely removed)
- Deprecate and remove phi-2, which is not supported with vLLM V1

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


chewong:
Got flagged by https://github.com/kaito-project/kaito/actions/runs/16813076971/job/47623004035?pr=1373 for using torch==2.7.1 due to https://github.com/advisories/GHSA-887c-mr87-cxwp, but I can't upgrade torch to 2.8.0 until vLLM also upgrades it.

chewong:
Hold until we release 0.6.0

chewong:
Closing in favor of #1378

PR #1386, opened 2025-08-12
chore: polish local-csi-driver charts

zhuangqh:
**Reason for Change**:

- remove /etc/kubernetes/ hostpath mount
- rename logging configmap to avoid confliction

**Notes for Reviewers**:

both helm upgrade and rollback are tested.

### DIFF
--- a/hack/local-csi-driver-charts-sync/local-csi-driver-charts.diff
+++ b/hack/local-csi-driver-charts-sync/local-csi-driver-charts.diff
@@ -1,5 +1,5 @@
 diff --git a/./local-csi-driver-charts.yaml b/./charts/kaito/workspace/templates/local-csi-driver-ds.yaml
-index 93d0b57..54cb7e4 100644
+index 93d0b571..cd8173e2 100644
 --- a/./local-csi-driver-charts.yaml
 +++ b/./charts/kaito/workspace/templates/local-csi-driver-ds.yaml
 @@ -4,13 +4,9 @@ apiVersion: v1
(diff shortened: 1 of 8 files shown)


hack/local-csi-driver-charts-sync/local-csi-driver-charts.diff
```diff
@@ -1,5 +1,5 @@
 diff --git a/./local-csi-driver-charts.yaml b/./charts/kaito/workspace/templates/local-csi-driver-ds.yaml
-index 93d0b57..54cb7e4 100644
```
andyzhangx, line 2: why there is a diff file here?

zhuangqh: local-csi-driver is an individual repo which have there own helm charts. https://github.com/Azure/local-csi-driver/tree/main/charts/latest

It's a part of the kaito helm chart here. Thus I add a diff file to maintain this chart, sync from original repo.
manifests rendered by helm will have some specific label which we don't want to include here.

hack/local-csi-driver-charts-sync/local-csi-driver-charts.diff
chewong, line 1: Could be a follow-up but I think we should use https://helm.sh/docs/helm/helm_dependency/ to take a dependency on https://github.com/Azure/local-csi-driver/tree/main/charts/latest as a best practice instead of using patch files (see #1252 as an example). However, It would involve updating local-csi-driver chart's templates and `values.yaml` to the point where we can customize container images, labels, volumes and volume mounts with helm values.

zhuangqh: Cool. I didn't know this. Let's do that.

chewong: Opened #1393

chewong (approved):

PR #1405, opened 2025-08-19
chore: bump vLLM to 0.10.1

### LINKED ISSUE #1374: workspace: consider CUDA driver memory usage when calculating GPU memory utilization
**Describe the bug**

Currently, when we calculate GPU memory utilization, the memory occupied by the CUDA driver is not included. 
KAITO reserved a fixed empirical value 1.5GiB for that, which is not compatible with all gpu type.  https://github.com/kaito-project/kaito/blob/main/pkg/model/interface.go#L70
https://github.com/kaito-project/kaito/blob/main/pkg/model/interface.go#L347

Beginning with vllm 0.9.1, the application will verify that sufficient free VRAM is available before starting the server. Please note that a portion of VRAM is reserved by the CUDA driver, which may lead to insufficient memory if GPU memory utilization is set too high.
```
ValueError: Free memory on device (21.37/23.72 GiB) on startup is less than desired GPU memory utilization (0.95, 22.54 GiB).
```

**Steps To Reproduce**

KAITO by default will set `--gpu-memory-utilization to 0.93` with nvidia A10 GPU. leads to this error.

**Expected behavior**

Make sure that the CUDA driver memory is included when calculating GPU memory utilization for different gpu.

From my testing, cuda memory is around 2.4GiB for A10 gpu.

code to get cuda mem
launch an empty process
```python
free_memory, total_memory = torch.cuda.mem_get_info()
cuda_memory = self.total_memory - self.free_memory
```

**Logs**

**Environment**

- Install tools:
- Others:

**Additional context**

chewong:
**Reason for Change**:

chore: bump vLLM to 0.10.1 to support gpt-oss

- Used `torch.cuda.mem_get_info()` to calculate the optimal `gpu_memory_utilization` instead of assuming a hardcoded 1.5Gi memory overhead
- LMCache size should be split among GPUs when tensor parallelism > 1, instead of taking 50% of available CPU memory per GPU, which would have led to OOM
- Disabled image building actions in PRs. They should be triggered in main branch only.
- Because of transformers upgraded to 4.55.2, we had to upgrade `accelerate` to 1.3.0 to fix the problematic tuning script. I also discovered that we didn't mount the data dir to the data downloader init container so this PR also fixes that.
- Removed deprecated `enable_reasoning` vLLM flag

**Requirements**

**Issue Fixed**:

Fixes #1374 

Ref #1354

**Notes for Reviewers**:


presets/workspace/inference/vllm/inference_api.py
```diff
@@ -292,6 +292,24 @@ def try_get_max_available_seq_len(args: argparse.Namespace) -> int | None:
...
+    # Derive an optimal gpu_memory_utilization from the current CUDA free/total
+    # bytes on the active device. This caps vLLM's GPU usage to what is actually
+    # available at runtime (accounting for driver/context/other processes) to reduce
+    # OOM risk during inference. Floor to 2 decimals to avoid rounding
+    # up. See https://github.com/kaito-project/kaito/issues/1374.
+    free_memory, total_memory = torch.cuda.mem_get_info()
+
+    # Reserve 1.5 GiB for vLLM runtime overhead to prevent OOM errors during initialization.
+    # This buffer is critical for smaller GPUs (e.g. A10) where memory margins are tight.
+    free_memory -= 1.5 * 1024**3
```
zhuangqh: This buffer was reserved for pytorch memory fragments. 
I didn't notice the memory reserved for cuda driver. So, I think now we don't need to reserve an extra 1.5G vram. maybe 512M is enough

chewong: Still encountering OOM with 512M 😞

.github/actions/e2e-base-setup/action.yaml
```diff
@@ -114,11 +114,32 @@ runs:
       run: |
         make generate
 
+    - name: build KAITO base image
+      shell: bash
+      if: ${{ inputs.build_kaito_base_image == 'true' }}
```
zhuangqh, line 119: run full e2e test if ${{ inputs.build_kaito_base_image == 'true' }}

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -292,6 +292,24 @@ def try_get_max_available_seq_len(args: argparse.Namespace) -> int | None:
         return None
 
 
+def get_max_gpu_memory_utilization() -> float:
+    # Derive an optimal gpu_memory_utilization from the current CUDA free/total
+    # bytes on the active device. This caps vLLM's GPU usage to what is actually
+    # available at runtime (accounting for driver/context/other processes) to reduce
+    # OOM risk during inference. Floor to 2 decimals to avoid rounding
+    # up. See https://github.com/kaito-project/kaito/issues/1374.
+    free_memory, total_memory = torch.cuda.mem_get_info()
```
zhuangqh: calling this method will initialize the underlying cuda context, which will consume an extra 230MB vram and cannot be released unless process exit.

use instead
```python
def gpu_mem_get_info(device_index: int = 0):
    """Return (free_bytes, total_bytes) for the given GPU device using NVML."""
    pynvml.nvmlInit()
    handle = pynvml.nvmlDeviceGetHandleByIndex(device_index)
    info = pynvml.nvmlDeviceGetMemoryInfo(handle)
    pynvml.nvmlShutdown()
    return info.free, info.total

free_memory, total_memory = gpu_mem_get_info()
```

zhuangqh: pip install nvidia-ml-py3

zhuangqh: Image

presets/workspace/dependencies/requirements.txt
```diff
@@ -1,10 +1,10 @@
 # Dependencies for TFS
 
 # Core Dependencies
-vllm==0.9.1
-transformers == 4.51.3
-torch==2.7.0
-accelerate==1.0.0
+vllm==0.10.1
```
zhuangqh: let's update to https://github.com/vllm-project/vllm/releases/tag/v0.10.1.1

vllm have a critical bug fix.

zhuangqh (approved):
LGTM

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -21,6 +21,7 @@
 from typing import Any
 
 import psutil
+import pynvml
```
andyzhangx, line 24: this is missing `pynvml` package, I have worked out a PR to fix this: 
https://github.com/kaito-project/kaito/pull/1424

PR #1407, opened 2025-08-20
feat: improve workspace controller for supporting scale subresource api

### LINKED ISSUE #1322: [Scale API] enhance workspace controller for supporting scale API
In order to support scale API for workspace, we need to improve workspace controller for reconciling TargetNodeCount in order to keep consistent with scaling action. then scale underlay workloads(Deployment) and update workspace status. 

- detailed design can be found here:
https://github.com/kaito-project/kaito/blob/main/website/docs/proposals/20250611-workspace-subresource-scale-api.md#workspace-controller

ehance workspace controller.

**Additional context**
The parent issue: #1104

rambohe-ch:
**Reason for Change**:

- improve workspace controller for supporting scale subresource api
1. remove applyWorkspaceResource in workspace controller
2. add nodeclaim controller for managing(create/delete) gpu nodes
3. add noderesource controller for ensuring labels and device plugin
4. when deployment has already existed and replicas is not equal to targetNodeCount, scale deployment
5. define node esitimator interface for calculating desired gpu count for a replica.

**Requirements**

**Issue Fixed**:

Fixes #1322 #1320 

**Notes for Reviewers**:


rambohe-ch:
@Fei-Guo  @zhuangqh @bangqipropel @chewong PTAL

Makefile
```diff
@@ -563,7 +563,7 @@ $(ENVTEST): $(LOCALBIN)
 
 .PHONY: vet
 vet: ## Run go vet against code.
-	go vet ./...
+	go vet $(shell go list ./... | grep -v /website/build/)
```
andyzhangx: what does this change mean?

rambohe-ch: I reverted this change.

pkg/utils/common.go
```diff
@@ -383,3 +383,18 @@ func ClientObjectSpecEqual(a, b client.Object) (bool, error) {
 	}
 	return aOK && bOK && equality.Semantic.DeepEqual(aSpec, bSpec), nil
 }
+
+// IsNodeReady checks if a Kubernetes node is in ready state
+func IsNodeReady(node *corev1.Node) bool {
```
chewong, line 388: nit: let's reuse `nodeIsReadyAndNotDeleting` from `pkg/workspace/controllers/workspace_controller.go`

pkg/workspace/controllers/noderesource/node_resource_controller.go
```diff
@@ -0,0 +1,411 @@
...
+		"NodeClaimNotReady", "Waiting for NodeClaims to be ready before resources can be provisioned")
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (c *NodeResourceReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	c.Client = mgr.GetClient()
+
+	builder := ctrl.NewControllerManagedBy(mgr).
+		Named("NodeResourceController").
+		For(&kaitov1beta1.Workspace{}).
```
chewong, line 387: What is the reason for having 3 separate controllers reconciling for the same Workspace? Isn't there going to be a lot of stale API server cache / update conflict issues?

rambohe-ch: The best practice of controller is: each controller should focus a specific feature, and the current status for workspace controller is complicated, because it includes nodeclaims management, node addons ensuring, tuning job, inference workloads, etc.

rambohe-ch: It will not cause stale metadata cache, because all metdata cache is managed by controller-runtime, and each resource has only one copy of cache.

pkg/utils/common.go
Jont828: Do we wan `nodeIsReadyAndNotDeleting` to be in utils instead? I think it makes sense to have it `utils/resource` where we also have `ListNodes`, but that leads to a circular import with `v1beta1`.

pkg/workspace/controllers/noderesource/node_resource_controller.go
rambohe-ch: @chewong My above reply is not correct. Yes, Multiple controllers use the same resource(workspace) to coordinate workflow will casue race conflict. I'll go to change these controllers(nodeclaim, noderesource) into libraries, and workspace controller will invoke them.

rambohe-ch:
I'll go to split this pull request into multiple small ones, so I'll close this one at first.

Issue #1297, opened 2025-07-22
Make pre-provisioned nodes experience easier


ritazh:
**Is your feature request related to a problem? Please describe.**

1. Confusion for new users: Most kubernetes users already have a cluster and may already have GPU nodes available. Kaito’s current workflow is optimized for the case where GPU lifecycle is bound to inference server lifecycle. It's an anti pattern for k8s. e.g. to deploy an nginx pod, typically k8s users do not expect k8s to create a new node for the pod to run.
2. Time takes to deploy a new model: To deploy a new model, the current GPU provisioner experience takes much longer than bring your own node experience since it also includes node provision time. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Jont828:
https://github.com/kaito-project/kaito/pull/1412

Jont828:
Are we good to close this since the BYO changes landed in #1536?

Issue #1222, opened 2025-06-27
generic way to handle model scheduling


sozercan:
**Is your feature request related to a problem? Please describe.**

Today, kaito has hardcoded requirements for preset model (example: https://github.com/kaito-project/kaito/blob/b1ae54622201f6c4718404256f29e631477e8353/presets/workspace/models/phi4/model.go#L76C3-L76C28)

these presets gets compared to hardcoded to specific cloud provider skus
https://github.com/kaito-project/kaito/blob/b1ae54622201f6c4718404256f29e631477e8353/api/v1beta1/workspace_validation.go#L361C7-L361C25
https://github.com/kaito-project/kaito/blob/b1ae54622201f6c4718404256f29e631477e8353/pkg/sku/azure_sku_handler.go

when there's a new preset or when user brings their own model, this becomes challenging since it'll require work from maintainer since requirements are hardcoded today.

when there's a new cloud provider SKU or if user wants to deploy this outside of Azure or AWS, this is also not extensible.

**Describe the solution you'd like**
 
- calculate the GPU VRAM requirements for a given model dynamically. example: https://apxml.com/tools/vram-calculator
- look at the node label info (example: `nvidia.com/gpu.memory) or DRA ResourceSlice to gather capacity
- schedule if capacity is greater than estimated VRAM need. 
- fast fail if estimated VRAM usage is more than capacity. another option here is to offload to RAM if desired by user.

**Describe alternatives you've considered**

**Additional context**

Fei-Guo:
@sozercan , please create a sub issue and link to #1178.

ritazh:
Let's make sure we document migration step: update node label selector to use accelerator sku instead of cloud provide sku

PR #1412, opened 2025-08-20
docs: add BYO nodes redesign proposal

Jont828:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

#1297 

**Notes for Reviewers**:

### DIFF
--- a/docs/proposals/20250820-byo-nodes.md
+++ b/docs/proposals/20250820-byo-nodes.md
@@ -41,16 +41,17 @@ The initial BYO nodes scenario was implemented with the preferred nodes list, wh
 
 The new design for the BYO scenario is for the user to label them with the corresponding Workspace label selector. This solution will incorporate the NAP feature gate, which we also propose to add as a controller flag. The behavior is as follows:
 
-
 ### NAP Disabled:
 
-All nodes matching the label selector will be selected and if the amount of nodes is less than the desired amount in the Workspace spec, an error will be returned immediately. The amount is indicated by `spec.inference.replicas` or the deprecated `spec.resource.count`. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the desired number of replicas, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the desired number of replicas, any of the nodes can be selected by the scheduler.
+All nodes matching the label selector will be selected and if the amount of nodes is less than the desired amount in the Workspace spec, an error will be returned immediately. The amount is indicated by `.inference.replicas` or the deprecated `.resource.count`. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the desired number of replicas, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the desired number of replicas, any of the nodes can be selected by the scheduler.
 
-We also propose that the instance type in `spec.resource.InstanceType` still doesn't need to be set by the user when using BYO nodes, but instead it will be empty instead of defaulted to `Standard_NC24ads_A100_v4` to avoid confusion.
+We also propose that the instance type in `.resource.InstanceType` still doesn't need to be set by the user when using BYO nodes, but instead it will be empty instead of defaulted to `Standard_NC24ads_A100_v4` to avoid confusion.
 
 ### NAP Enabled
 
 Like the previous scenario, all the BYO nodes will be selected and verified. However, the instance type is now required for the auto-provisioner to determine what type of node to create. The user nodes must be of the same instance type or the webhook will return an error. KAITO uses the value of the label `node.kubernetes.io/instance-type` to determine the instance type, so the user provided nodes must also have this label as well. If the amount of ready nodes of the correct instance is less than the count, the auto-provisioner will create new nodes. The scheduler will not distinguish between user provided and auto provisioned nodes.
+
+Additionally, if `.resource.InstanceType` is not set, the auto-provisioner will not create new nodes for the Workspace. Although NAP is a controller flag, in effect the user can disable it for a given Workspace by leaving `.resource.InstanceType` unset.
 
 # History
 
--- a/docs/proposals/20250820-byo-nodes.md
+++ b/docs/proposals/20250820-byo-nodes.md
@@ -1,0 +1,50 @@
+---
+title: Redesign BYO Nodes Scenario 
+authors:
+  - "Jonathan Tong"
+reviewers:
+  - "Jonathan Tong"
+creation-date: `2025-08-20`
+last-updated: `2025-08-20`
+status: provisional
+---
+
+# Title
+
+Redesign BYO Nodes Scenario and Deprecate `preferredNodes` field
+
+## Glossary
+
+- BYO nodes: bring-your-own nodes
+- NAP: node auto-provisioning
+
+## Summary
+
+This proposal aims to redesign the bring-your-own (BYO) use case and deprecate the `preferredNodes` list from the resource spec. The existing BYO scenario is handled by the `preferredNodes` list, and instead BYO nodes will only need to match the Workspace label selector.
+
+## Motivation
+
+The initial BYO nodes scenario was implemented with the `preferredNodes` list, where users would specify the names of the nodes they want to use in the Workspace spec. This approach had several issues that pertain to the design, not just the code, so rather than attempting to introduce awkward fixes one at a time, we will redesign the BYO nodes scenario to simplify the user experience and improve the overall system architecture. The existing limiations are as follows:
+
+1. In order for any node to be selected by the Workspace, it must be labeled to match the Workspace label selector. This is the case for BYO and NAP nodes, so there already is an affiliation between BYO nodes and the Workspace, which means that providing the names of the nodes in the `preferredNodes` list is redundant. This has been mentioned several times in issues like [this](https://github.com/kaito-project/kaito/pull/1337#pullrequestreview-3122605167).
+2. When NAP is enabled and the user has also provided nodes in the `preferredNodes` list, the name of the field implies that these will be scheduled first. The existing code to [select the nodes](https://github.com/kaito-project/kaito/blob/8c8585a138c4a9273de6149e184ff5b951cd4c18/pkg/utils/common.go#L243-L288) sorts them in order of priority with preferred nodes first, but there isn't a way to guarantee that the scheduler actually picks the nodes from the list out of all possible nodes. For example, if one node is needed and there's an existing NAP node with a preferred node that got added, the scheduler could pick either one. There's no clear way to have the scheduler prioritize nodes in this way, and so dropping the idea of `preferredNodes` in favor of all qualified nodes sharing the same label would be more clear to the user.
+
+## Goals
+
+1. Simplify the BYO nodes scenario by removing the `preferredNodes` list and relying solely on the Workspace label selector for node selection.
+2. Clearly define the behavior of how user provided nodes will be handled when NAP is enabled and disabled.
+3. Surface errors to the user when there is an error with their provided nodes and cannot run the workload.
+
+
+## Solution Design
+
+The new design for the BYO scenario is for the user to label them with the corresponding Workspace label selector. This also will affect the NAP feature gate, and the behavior is as follows:
+
+- NAP disabled: All nodes matching the label selector will be selected and if the amount of nodes is less than the `count` field in the Workspace spec, an error will be returned immediately. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the `count`, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the `count`, any of the nodes can be selected by the scheduler.
+- NAP enabled: Like the previous scenario, all the BYO nodes will be selected and verified. However, the user provided nodes must also have the matching `instanceType`. If the amount of ready nodes of the correct instance is less than the count, the auto-provisioner will create new nodes. The scheduler will not distinguish between user provided and auto provisioned nodes.
+
+# History
+
+- [x] 2025/08/20: Open proposal PR.
+- [ ] MM/DD/YYYY: Start model integration.
+- [ ] MM/DD/YYYY: Complete model support.
--- a/docs/proposals/20250820-byo-nodes.md
+++ b/docs/proposals/20250820-byo-nodes.md
@@ -20,7 +20,7 @@ Redesign BYO Nodes Scenario and Deprecate `preferredNodes` field
 
 ## Summary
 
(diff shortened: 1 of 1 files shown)


docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,50 @@
...
+## Goals
+
+1. Simplify the BYO nodes scenario by removing the `preferredNodes` list and relying solely on the Workspace label selector for node selection.
+2. Clearly define the behavior of how user provided nodes will be handled when NAP is enabled and disabled.
+3. Surface errors to the user when there is an error with their provided nodes and cannot run the workload.
+
+
+## Solution Design
+
+The new design for the BYO scenario is for the user to label them with the corresponding Workspace label selector. This also will affect the NAP feature gate, and the behavior is as follows:
```
chewong: nit: If this is a permanent on/off button, I propose making it a controller flag instead of feature gate. Feature gate implies that it would be removed in the future once a particular feature is graduated.

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,50 @@
...
+
+- NAP disabled: All nodes matching the label selector will be selected and if the amount of nodes is less than the `count` field in the Workspace spec, an error will be returned immediately. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the `count`, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the `count`, any of the nodes can be selected by the scheduler.
+- NAP enabled: Like the previous scenario, all the BYO nodes will be selected and verified. However, the user provided nodes must also have the matching `instanceType`. If the amount of ready nodes of the correct instance is less than the count, the auto-provisioner will create new nodes. The scheduler will not distinguish between user provided and auto provisioned nodes.
```
chewong: Let's mention how KAITO is relying on the node label value of `node.kubernetes.io/instance-type` to determine whether a node has the correct

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,50 @@
...
+1. Simplify the BYO nodes scenario by removing the `preferredNodes` list and relying solely on the Workspace label selector for node selection.
+2. Clearly define the behavior of how user provided nodes will be handled when NAP is enabled and disabled.
+3. Surface errors to the user when there is an error with their provided nodes and cannot run the workload.
+
+
+## Solution Design
+
+The new design for the BYO scenario is for the user to label them with the corresponding Workspace label selector. This also will affect the NAP feature gate, and the behavior is as follows:
+
+- NAP disabled: All nodes matching the label selector will be selected and if the amount of nodes is less than the `count` field in the Workspace spec, an error will be returned immediately. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the `count`, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the `count`, any of the nodes can be selected by the scheduler.
```
chewong: `count` field has been deprecated in favor of [`replicas`](https://github.com/kaito-project/kaito/blob/main/api/v1beta1/workspace_types.go#L89) in preparation for autoscaling feature. Let's reflect it in this doc

Jont828:
@Fei-Guo @zhuangqh @l

rambohe-ch:
There are two concerns:
1. preferred nodes and ANP can work at the same time, ANP will create nodes dynamically when BYO nodes is not enough. After remove preferred nodes, this scenario will be not supported.
2. preferred nodes is a workspace level field, and disable node provisioner is a component level field. this means if we disable node provisioner, all workspace can not use ANP.

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+## Solution Design
+
+The new design for the BYO scenario is for the user to label them with the corresponding Workspace label selector. This solution will incorporate the NAP feature gate, which we also propose to add as a controller flag. The behavior is as follows:
+
+
+### NAP Disabled:
+
+All nodes matching the label selector will be selected and if the amount of nodes is less than the desired amount in the Workspace spec, an error will be returned immediately. The amount is indicated by `spec.inference.replicas` or the deprecated `spec.resource.count`. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the desired number of replicas, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the desired number of replicas, any of the nodes can be selected by the scheduler.
```
chewong:
```suggestion
All nodes matching the label selector will be selected and if the amount of nodes is less than the desired amount in the Workspace spec, an error will be returned immediately. The amount is indicated by `.inference.replicas` or the deprecated `.resource.count`. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the desired number of replicas, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the desired number of replicas, any of the nodes can be selected by the scheduler.
```

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,59 @@
...
+
+We also propose that the instance type in `spec.resource.InstanceType` still doesn't need to be set by the user when using BYO nodes, but instead it will be empty instead of defaulted to `Standard_NC24ads_A100_v4` to avoid confusion.
```
chewong:
```suggestion
We also propose that the instance type in `.resource.instanceType` still doesn't need to be set by the user when using BYO nodes, but instead it will be empty instead of defaulted to `Standard_NC24ads_A100_v4` to avoid confusion.
```

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,59 @@
...
+
+### NAP Enabled
+
+Like the previous scenario, all the BYO nodes will be selected and verified. However, the instance type is now required for the auto-provisioner to determine what type of node to create. The user nodes must be of the same instance type or the webhook will return an error. KAITO uses the value of the label `node.kubernetes.io/instance-type` to determine the instance type, so the user provided nodes must also have this label as well. If the amount of ready nodes of the correct instance is less than the count, the auto-provisioner will create new nodes. The scheduler will not distinguish between user provided and auto provisioned nodes.
```
zhuangqh: > KAITO uses the value of the label `node.kubernetes.io/instance-type` to determine the instance type, so the user provided nodes must also have this label as well.

make sense

Jont828:
> There are two concerns:
> 
> 1. preferred nodes and ANP can work at the same time, ANP will create nodes dynamically when BYO nodes is not enough. After remove preferred nodes, this scenario will be not supported.
> 2. preferred nodes is a workspace level field, and disable node provisioner is a component level field. this means if we disable node provisioner, all workspace can not use ANP.

@rambohe-ch So I think the first one isn't a problem as this will still be supported. If you don't have enough nodes matching the label selector, NAP should still trigger and create new ones. For the second point, maybe we can remove the defaulting for `instanceType` and say that if NAP is enabled and say that if `instanceType` is not set, then NAP won't be triggered for this. Wdyt?

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,60 @@
...
+
+We also propose that the instance type in `.resource.InstanceType` still doesn't need to be set by the user when using BYO nodes, but instead it will be empty instead of defaulted to `Standard_NC24ads_A100_v4` to avoid confusion.
+
+### NAP Enabled
+
+Like the previous scenario, all the BYO nodes will be selected and verified. However, the instance type is now required for the auto-provisioner to determine what type of node to create. The user nodes must be of the same instance type or the webhook will return an error. KAITO uses the value of the label `node.kubernetes.io/instance-type` to determine the instance type, so the user provided nodes must also have this label as well. If the amount of ready nodes of the correct instance is less than the count, the auto-provisioner will create new nodes. The scheduler will not distinguish between user provided and auto provisioned nodes.
+
+Additionally, if `.resource.InstanceType` is not set, the auto-provisioner will not create new nodes for the Workspace. Although NAP is a controller flag, in effect the user can disable it for a given Workspace by leaving `.resource.InstanceType` unset.
```
Fei-Guo: I'd suggest making resource.InstanceType a required field if NAP is enabled to simplify the workflow behavior. The sku is critical for the gpu memory estimator to calculate how many nodes are needed in provisioning and autoscaling cases.

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,60 @@
...
+3. Surface errors to the user when there is an error with their provided nodes and cannot run the workload.
+
+
+## Solution Design
+
+The new design for the BYO scenario is for the user to label them with the corresponding Workspace label selector. This solution will incorporate the NAP feature gate, which we also propose to add as a controller flag. The behavior is as follows:
+
+### NAP Disabled:
+
+All nodes matching the label selector will be selected and if the amount of nodes is less than the desired amount in the Workspace spec, an error will be returned immediately. The amount is indicated by `.inference.replicas` or the deprecated `.resource.count`. The webhook will then verify that each node is ready/running. If the total amount of ready nodes is less than the desired number of replicas, then an error will be returned again. The Workspace will schedule the model to run on these nodes, and if the amount of nodes exceeds the desired number of replicas, any of the nodes can be selected by the scheduler.
```
Fei-Guo: If you leave the sku field empty, our gpu memory estimator cannot tell how many gpu nodes are needed. You need to either make the count a required field or agree we default on only requiring one node.

docs/proposals/20250820-byo-nodes.md
Jont828: Okay, I added this change in to address [Linbo's comment](https://github.com/kaito-project/kaito/pull/1412#issuecomment-3208633381) on a way to enable/disable NAP for specific workspaces. I'm okay with removing that if we're fine keeping NAP as a controller level field instead.

rambohe-ch:
> > There are two concerns:
> > 
> > 1. preferred nodes and ANP can work at the same time, ANP will create nodes dynamically when BYO nodes is not enough. After remove preferred nodes, this scenario will be not supported.
> > 2. preferred nodes is a workspace level field, and disable node provisioner is a component level field. this means if we disable node provisioner, all workspace can not use ANP.
> 
> @rambohe-ch So I think the first one isn't a problem as this will still be supported. If you don't have enough nodes matching the label selector, NAP should still trigger and create new ones. For the second point, maybe we can remove the defaulting for `instanceType` and say that if NAP is enabled and say that if `instanceType` is not set, then NAP won't be triggered for this. Wdyt?

@Jont828 Thanks for replying. It seems that there is no need to add controller flag to disable/enable NAP.  because we can use the following rules to satisfy requirements:
1. Use `workspace.Resource.LabelSelector` all ready nodes(including BYO nodes). 
2. If BYO nodes are enough for running workloads of workspace(compare to `workspace.Status.Inference.TargetNodeCount`), no NAP nodes will be created.
3. If BYO nodes is not enough, the insufficient nodes will be created according to `workspace.Resource.InstanceType`.

by the way, I agree with the proposal to remove the `PreferredNodes` fields.

chewong:
I agree with @rambohe-ch and @Jont828 that the NAP-enabled path is well-lit and uncontroversial, and `workspace.Resource.InstanceType` should be required. We should also deprecate the preferred nodes field.

For the NAP-disabled case, can we agree that `workspace.Resource.InstanceType` should be an empty string? Currently, KAITO uses this field to:

1) Identify the SKU when provisioning additional nodes via gpu-provisioner.
2) Obtain GPU count and memory hardcoded in [SKU handler files](https://github.com/kaito-project/kaito/blob/8f6f312d4f6fe7ededd33b40025dcb5019c58060/pkg/sku/azure_sku_handler.go) to ensure smooth model execution without OOM issues.

Point 1 is irrelevant when NAP is disabled. For point 2, alternatives like [gpu-feature-discovery](https://github.com/NVIDIA/k8s-device-plugin/blob/4579a60b444627d3afdbcfbbad4b16f43ed4029c/docs/gpu-feature-discovery/README.md) can provide GPU information via node labels (`nvidia.com/gpu.memory` and `nvidia.com/gpu.count`), which should suffice for our use case. The disadvantage is it's another dependency we need to package/install with KAITO.

However, the general drawback of requiring `workspace.Resource.InstanceType` in every Workspace is that it's cloud-provider-specific and the need to continuously track new GPU SKUs and deprecate retired ones by maintaining the [SKU handler files](https://github.com/kaito-project/kaito/blob/8f6f312d4f6fe7ededd33b40025dcb5019c58060/pkg/sku/azure_sku_handler.go), so I think gpu-feature-discovery is more flexible (ref: #1222).

docs/proposals/20250820-byo-nodes.md
```diff
@@ -0,0 +1,60 @@
...
+
+We also propose that the instance type in `.resource.InstanceType` still doesn't need to be set by the user when using BYO nodes, but instead it will be empty instead of defaulted to `Standard_NC24ads_A100_v4` to avoid confusion.
```
sozercan: technically, this is a breaking change

docs/proposals/20250820-byo-nodes.md
sozercan: this is related to #1222. maybe #1222 should be handled before this change?

docs/proposals/20250820-byo-nodes.md
Jont828: I think this is more of a bug fix. Right now, if I bring an H100 node for a large model like Llama 70B and do not specify the instance type, the Workspace will check the GPU memory using the default A100 size and fail. We can either make instance type optional or required, but I don't think it makes sense to have it default to this value if it can cause validation failure.

docs/proposals/20250820-byo-nodes.md
Jont828: What exactly is the current behavior for the GPU memory estimator? I agree that ideally we should be able to figure out the memory requirement per #1222 and drop the instance type. Currently, if I bring an H100 node to run Llama 70B, for example, and I leave the instance type blank, it defaults to the A100 node size. The validation fails saying the node doesn't have enough memory when it does, and it's doing the calculation on the wrong size. If we're going to do this webhook validation, we should be validating on the actual size of the node so that there aren't any false positives so I think we should move away from relying on a user-provided SKU field.

Jont828:
@chewong I am +1 on this. Right now, for NAP being disabled, if you don't specify the instance type, it defaults to A100. If I'm running a H100 with a large model, the webhook will fail because it tried to compute the memory requirement for the wrong GPU. Given that the validation is failing when it shouldn't when the instance type is not provided, it doesn't make sense to have this default value and we can work towards finding another way to validate the memory requirements.

chewong:
Merging this since it's a doc-only proposal. We can continue iterating on this with follow-up PRs.

PR #1417, opened 2025-08-21
feat: add nodeclaim manager for creating/deleting nodeclaims of workspace

rambohe-ch:
**Reason for Change**:

add nodeclaim manager for creating/deleting nodeclaims of workspace

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -433,3 +434,113 @@ func CheckNodeClass(ctx context.Context, kClient client.Client) error {
...
+	// If there's a label selector, add it to the list options
+	if wObj.Resource.LabelSelector != nil {
+		selector, err := metav1.LabelSelectorAsSelector(wObj.Resource.LabelSelector)
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert label selector: %w", err)
+		}
+		listOpts = append(listOpts, client.MatchingLabelsSelector{Selector: selector})
+	}
+
+	if err := c.List(ctx, nodeList, listOpts...); err != nil {
```
chewong: nit: reuse existing util function

```suggestion
	nodeList, err := resources.ListNodes(...)
```

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -433,3 +434,113 @@ func CheckNodeClass(ctx context.Context, kClient client.Client) error {
...
+
+		// Check if the node is ready
+		if !IsNodeReady(node) {
+			klog.V(4).InfoS("BYO node is not ready, skipping",
+				"node", node.Name,
+				"workspace", klog.KObj(wObj))
+			continue
+		}
+
+		availableBYONodes = append(availableBYONodes, node)
```
chewong: `availableBYONodes` might not be a good name because nodes in this list can either be BYO or nodes that are provisioned by KAITO. Did you mean "qualified" nodes?

pkg/workspace/resource/node_claim.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	// Alternative check: if NodeClaim has been launched successfully
+	// Check if the NodeClaim has a node name assigned (indicates it's provisioned)
+	if nodeClaim.Status.NodeName != "" {
+		return true
+	}
+
+	// If no Ready condition found and no node assigned, consider it not ready
+	return false
```
chewong: nit:

```suggestion
	return nodeClaim.Status.NodeName != ""
```

pkg/utils/nodeclaim/nodeclaim.go
chewong: Per existing logic, we should also check if there is `nvidia.com/gpu` resource for the node right? https://github.com/kaito-project/kaito/blob/main/pkg/workspace/controllers/workspace_controller.go#L618

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -433,3 +434,113 @@ func CheckNodeClass(ctx context.Context, kClient client.Client) error {
...
+	}
+
+	if err := c.List(ctx, nodeClaimList, listOpts...); err != nil {
+		return nil, fmt.Errorf("failed to list NodeClaims: %w", err)
+	}
+
+	// Convert to slice of pointers for easier manipulation
+	nodeClaims := make([]*karpenterv1.NodeClaim, 0, len(nodeClaimList.Items))
+	for i := range nodeClaimList.Items {
+		nodeClaims = append(nodeClaims, &nodeClaimList.Items[i])
```
chewong, line 455: Do we need to check if DeletionTimestamp is zero before appending?

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -433,3 +434,113 @@ func CheckNodeClass(ctx context.Context, kClient client.Client) error {
...
+	targetNodeCount := 1
+	if wObj.Inference != nil && wObj.Status.Inference != nil {
+		targetNodeCount = int(wObj.Status.Inference.TargetNodeCount)
+	}
+
+	// Calculate the number of NodeClaims needed (target - BYO nodes)
+	return max(0, targetNodeCount-len(availableBYONodes)), nil
+}
+
+func IsNodeReady(node *v1.Node) bool {
```
chewong: Same here. Let's ensure DeletionTimestamp is zero, or just use the existing util function `nodeIsReadyAndNotDeleting`

pkg/workspace/resource/node_claim.go
```diff
@@ -0,0 +1,380 @@
...
+		// Need to create more NodeClaims
+		nodesToCreate := requiredNodeClaimsCount - currentNodeClaimCount
+		klog.InfoS("Creating additional NodeClaims",
+			"workspace", klog.KObj(wObj),
+			"current", currentNodeClaimCount,
+			"required", requiredNodeClaimsCount,
+			"toCreate", nodesToCreate)
+
+		// Update status condition to indicate NodeClaims are being created
+		if updateErr := status.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeNodeClaimStatus, metav1.ConditionFalse,
```
chewong, line 104:
```suggestion
		if updateErr := status.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeNodeClaimStatus, metav1.ConditionTrue,
```

pkg/workspace/resource/node_claim.go
```diff
@@ -0,0 +1,380 @@
...
+			"CreatingNodeClaims", fmt.Sprintf("Creating %d additional NodeClaims (current: %d, required: %d)", nodesToCreate, currentNodeClaimCount, requiredNodeClaimsCount)); updateErr != nil {
```
chewong, line 105: This condition is only set to true once. I can't find where this condition is being set to False once creation is complete. Do we need this condition at all?

pkg/utils/nodeclaim/nodeclaim.go
zhuangqh: Yes. We need to confirm it's not in deleting state.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -433,3 +434,113 @@ func CheckNodeClass(ctx context.Context, kClient client.Client) error {
...
+	return availableBYONodes, nil
+}
+
+// GetExistingNodeClaims retrieves all NodeClaims associated with the given workspace
+func GetExistingNodeClaims(ctx context.Context, c client.Reader, wObj *kaitov1beta1.Workspace) ([]*karpenterv1.NodeClaim, error) {
+	nodeClaimList := &karpenterv1.NodeClaimList{}
+
+	// List NodeClaims with labels that match this workspace
+	listOpts := []client.ListOption{
+		client.InNamespace(wObj.Namespace),
```
zhuangqh: nodeclaim is not namespaced object?

pkg/utils/nodeclaim/nodeclaim.go
rambohe-ch: fixed

pkg/utils/nodeclaim/nodeclaim.go
rambohe-ch: only nodes in preferred list or BYO nodes(when disabled node provision feature gate is true), so only BYO nodes will be collected. so name availableBYONodes is more suitable.

pkg/workspace/resource/node_claim.go
rambohe-ch: fixed

pkg/utils/nodeclaim/nodeclaim.go
rambohe-ch: the above login is extracted as node resource manager, the pull request link: https://github.com/kaito-project/kaito/pull/1418

then nodeclaim manager need to manage nodeclaim resource only.

pkg/utils/nodeclaim/nodeclaim.go
rambohe-ch: no, this function is used to list all existing nodeclaims. I think we should wait all nodeclaim resource is removed before we can create new nodeclaims for the workspace.

pkg/utils/nodeclaim/nodeclaim.go
Fei-Guo: yes, nodeclaim is cluster scoped.

pkg/workspace/resource/node_claim.go
```diff
@@ -0,0 +1,380 @@
...
+					// Successfully created
+					created = true
+					break
+				} else if apierrors.IsAlreadyExists(err) {
+					// NodeClaim with this name already exists, wait and retry with new name
+					klog.V(4).InfoS("NodeClaim already exists, generating new name and retrying",
+						"nodeClaim", nodeClaim.Name,
+						"workspace", klog.KObj(wObj),
+						"retry", retryAttempt+1)
+					time.Sleep(100 * time.Millisecond)
```
Fei-Guo: Why is there a sleep here?

pkg/utils/nodeclaim/nodeclaim.go
rambohe-ch: fixed

pkg/utils/nodeclaim/nodeclaim.go
rambohe-ch: fixed

pkg/workspace/resource/node_claim.go
```diff
@@ -0,0 +1,329 @@
...
+				if err == nil {
+					created = true
+					return nil
+				}
+				return err
+			})
+
+			if !created {
+				// Failed to create, decrement expectations
+				c.expectations.CreationObserved(c.logger, workspaceKey)
```
zhuangqh, line 111: I think you need to add expectation one by one. 

now you have `c.expectations.ExpectCreations(c.logger, workspaceKey, nodesToCreate)`
if the first nodeclaim failed to created,  you will only decrease once.

rambohe-ch: fixed, create/delete nodeclaim will continue when the create/delete request fails.

pkg/utils/nodeclaim/nodeclaim.go
chewong: But nodes created by NodeClaims also have the same labels as BYO nodes

PR #1423, opened 2025-08-22
refactor: replace .github/determine_models.py with bash script

chewong:
**Reason for Change**:

This PR converts the existing Python-based model build matrix generation script to a pure Bash implementation that outputs a GitHub Actions matrix for models that need container image builds.

Main differences comparing to the Python script
- Only includes models in the build matrix when their container image tags are missing from the registry
- Does not assume building images in PRs

**Requirements**

Testing:

```
.github/determine_models.sh
[{"name":"base","type":"text-generation","runtime":"tfs","tag":"0.0.6"}]
```

The current python script is broken:

image

**Issue Fixed**:

**Notes for Reviewers**:


chewong:
Test run:
image

image

PR #1427, opened 2025-08-22
feat: only run vector search on latest user message

bfoley13:
**Reason for Change**:
This adds handling for only running vector search on the latest "user" type input. We make sure there is a `user` message at the end of the request messages, and combine them if there are multiple.  This also cleans up the handling of messages, from combining them into a single string, into handling them as llamaindex `ChatMessage` objects.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:
SS of the request within the RAG before sending to LLM. RAG attaches the extra prompt + context at the beginning of the messages, so it will be applied before chat history as well if there is any.
image


presets/ragengine/models.py
```diff
@@ -122,6 +122,15 @@ def messages_to_prompt(messages: list[dict]) -> str:
     return "\n".join(string_messages)
 
 
+def messages_to_simplified_messages(messages: list[dict]):
```
Fei-Guo: Can you add some examples in a comment to explain what exactly the `get_message_content` and `message_to_simplified_messages` methods do?

presets/ragengine/vector_store/base.py
```diff
@@ -359,8 +365,43 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
...
+                    0, new_message
+                )  # Insert at beginning to maintain order
+
+        # Combine the last consecutive user messages into a single string as the prompt
+        # Generally there should only be 1 user message on the end but there could be more
+        user_prompt = (
+            "\n\n".join(user_messages_for_prompt) if user_messages_for_prompt else ""
+        )
+
+        # validate we have a user prompt. if not using tools/etc. we should have a user prompt
```
Fei-Guo: I feel this check is overly restricted. I think as long as the chat template allows reordering user and non-user, we should support it.

presets/ragengine/vector_store/base.py
```diff
@@ -381,20 +422,21 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
         logger.info(
             f"Creating chat engine for index '{request.get('index_name')}' with prompt size: {prompt_len}"
         )
-        chat_engine = self.index_map[
-            request.get("index_name")
-        ].as_chat_engine(
+        # Calculate top_k based on available context. For larger windows, we can afford to retrieve more documents.
+        # 750 tokens is a rough estimate based off max 1500 for code and 1000 for text
```
Fei-Guo: Can you elaborate on how to derive 750 from 1500 and 1000?

presets/ragengine/vector_store/base.py
```diff
@@ -381,20 +422,21 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
...
+        top_k = max(100, int((self.llm.metadata.context_window - prompt_len) / 750))
+        chat_engine = self.index_map[request.get("index_name")].as_chat_engine(
             llm=self.llm,
-            similarity_top_k=100,  # Might want to make this a function of avg doc node size in an index but this should be a wide enough default
-            chat_mode=ChatMode.CONDENSE_PLUS_CONTEXT,
+            similarity_top_k=top_k,
+            chat_mode=ChatMode.CONTEXT,
```
Fei-Guo, line 446: What is the difference between the CONDENSE_PLUS_CONTEXT vs CONTEXT?

presets/ragengine/vector_store/base.py
```diff
@@ -381,20 +422,21 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
...
             node_postprocessors=[
                 ContextSelectionProcessor(
                     rag_context_token_fill_ratio=request.get(
                         "context_token_ratio", RAG_DEFAULT_CONTEXT_TOKEN_FILL_RATIO
                     ),
                     llm=self.llm,
                     max_tokens=max_tokens,
-                    similarity_threshold=0.8,
+                    similarity_threshold=0.85,
```
Fei-Guo: Why change this? I'd prefer having a comment for every hardcoded number in the code base.

presets/ragengine/vector_store/base.py
```diff
@@ -404,10 +446,14 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
             if self.use_rwlock:
                 async with self.rwlock.reader_lock:
                     self.llm.set_params(llm_params)
-                    chat_result = await chat_engine.achat(prompt)
+                    chat_result = await chat_engine.achat(
+                        user_prompt, chat_history=chat_history
```
Fei-Guo, line 465: What is the reasoning for adding chat_history here?

presets/ragengine/vector_store/base.py
bfoley13: CONDENSE_PLUS_CONTEXT condenses chat history and tries to run vector search off all chat + history. Context will only add the context on the latest chat message.

presets/ragengine/vector_store/base.py
bfoley13: The similarity search values resulting from FAISS search are not consistent but are usually within some range. The [test here](https://github.com/kaito-project/kaito/blob/fe7648fd4cacd3bbef9c6ceefe90734383dc10ae/presets/ragengine/tests/api/test_node_processing.py#L44) was flaky with some similarity score hovering > 0.8 causing it to fail. I'll make an env var for this one as well, I think it might be work having the ability to tweak

presets/ragengine/vector_store/base.py
bfoley13: When a request comes in, we split it into the latest user chat message, and the rest. For most use cases you might have a system/developer message and a user message. Putting everything else into the `chat_history` allows us to only run vector search on user prompt, rather than the system + user prompts. But the other messages will be added to the request sent to the LLM

presets/ragengine/vector_store/base.py
bfoley13: Theres absolutely no way we know what chat template is being used here. Perhaps a better solution would be the user messages from the last `assistant` message (aka the llm response) become the user_prompt and the chat_history becomes reordered.

presets/ragengine/vector_store/chat_completions_wrapper.md
```diff
@@ -0,0 +1,208 @@
+# RAGEngine Chat Completion Algorithm
```
bfoley13, line 1: I added an in depth dive into the handling of the chat_completions wrapper algorithm. Didnt think it was necessary for the website docs but can move it if needed.

presets/ragengine/vector_store/base.py
bfoley13: Updated this to find the user messages since the last `assistant` message. In theory we should have a later user message when running the RAG. If we want to allow it but passthrough that could be an option as well. Added some info in the md doc if you want to readthrough the handling

presets/ragengine/models.py
bfoley13: I swapped this code to a simpler `input_message_to_llamaindex_message` function that will convert the input openAI spec messages into LlamaIndex messages.

presets/ragengine/vector_store/base.py
bfoley13: I already had this as a env var at 0.85 but forgot to swap it. Moved to using the env var as well as created another env var for document node token approximation.

presets/ragengine/vector_store/base.py
bfoley13: Added a env var for this and added comments on how it was calculated, also changed to 500.

presets/ragengine/vector_store/base.py
```diff
@@ -359,8 +366,49 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
...
+        # Generally there should only be 1 user message after the latest assistant message
+        # but there could be more in some cases
+        user_prompt = (
+            "\n\n".join(user_messages_for_prompt) if user_messages_for_prompt else ""
+        )
+
+        # Validate we have a user prompt. if not using tools/etc. we should have a user prompt
+        # as rag retrieval should only be run on user input.
+        if user_prompt == "":
+            logger.error(
```
bfoley13, line 403: we could passthrough here instead of error

PR #1446, opened 2025-08-29
fix: add nodeSelector on Linux to avoid crash on Windows node

andyzhangx:
**Reason for Change**:

fix: add nodeSelector on Linux to avoid crash on Windows node

```
kaito-workspace     kaito-workspace-6db9666b47-xjz7m                      0/1     ImagePullBackOff    0               12d     10.244.4.201   akswgen2000000                      <none>           <none>
kaito-workspace     nvidia-device-plugin-daemonset-nx8d7                  0/1     ImagePullBackOff    0               27d     10.244.4.237   akswgen2000000                      <none>           <none>
```

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/charts/kaito/workspace/values.yaml
+++ b/charts/kaito/workspace/values.yaml
@@ -32,7 +32,8 @@ resources:
   requests:
     cpu: 10m
     memory: 64Mi
-nodeSelector: {}
+nodeSelector:
+  kubernetes.io/os: linux
 tolerations: []
 affinity: {}
 # Values can be "azure" or "aws" or "arc"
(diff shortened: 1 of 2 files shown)


charts/kaito/workspace/values.yaml
```diff
@@ -32,7 +32,8 @@ resources:
   requests:
     cpu: 10m
     memory: 64Mi
-nodeSelector: {}
```
Fei-Guo, line 35: Is this change required? I think workspace controller should work in windows node.

Fei-Guo: Nvm, we need to build a windows image for that which does not exist today.

PR #1461, opened 2025-09-02
feat: separate EnsureNodeClaims into ScaleUpNodeClaims and ScaleDownNodeClaims

rambohe-ch:
**Reason for Change**:

It's not a good idea to put both `create nodeclaims` and `delete nodeclaims` in EnsureNodeClaims function. because `create nodeclaims(scaleup)` should be invoked before scaling up workloads of workspace. and `delete nodeclaims(scaledown)` should be invoked after scaling down workloads of workspace instead.

EnsureNodeClaims function has been separated into 4 functions as follows:
- DiffNodeClaims
- ScaleUpNodeClaims
- ReadyNodeClaimsMeetTarget
- ScaleDownNodeClaims

Moreover, NodeClaims that have no pods of workspace running on the node will be scaled down.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/workspace/resource/node_claim.go
```diff
@@ -48,172 +49,213 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
-				return deletingI // being deleted comes first
+	if nodesToDelete == 0 {
+		if curCondition := meta.FindStatusCondition(wObj.Status.Conditions, string(kaitov1beta1.ConditionTypeScalingDownStatus)); curCondition != nil {
+			if updateErr := workspace.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeScalingDownStatus, metav1.ConditionTrue, "ScalingDownNodeClaimsCompleted", "Scaling down excess NodeClaims completed"); updateErr != nil {
+				klog.ErrorS(updateErr, "failed to update scaling down status condition(ScalingDownNodeClaimsCompleted)", "workspace", workspaceKey)
+				return false, fmt.Errorf("failed to update scaling down status condition(ScalingDownNodeClaimsCompleted): %w", updateErr)
 			}
+		}
+		return true, nil
+	}
```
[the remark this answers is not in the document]
rambohe-ch: Because ConditionTypeScalingDownStatus condition is only confgured for scaling down action, so this condition should not be set to true when it doesn't exist.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -478,3 +478,30 @@ func GetRequiredNodeClaimsCount(ctx context.Context, c client.Client, wObj *kait
...
+
+	if err := c.List(ctx, podList, listOpts...); err != nil {
+		return false
+	}
+
+	for _, pod := range podList.Items {
+		if pod.Spec.NodeName == nodeClaimObj.Status.NodeName {
+			return true
+		}
+	}
```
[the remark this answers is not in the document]
rambohe-ch: I think we should include all pods even pods is terminating. this means workloads is definitely cleared before scaling down nodeclaims.

pkg/workspace/resource/node_claim.go
```diff
@@ -48,179 +49,203 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
 	}
 }
 
-// EnsureNodeClaims ensures the correct number of NodeClaims for the workspace
-// based on the TargetNodeCount in the workspace status, considering BYO nodes
-// only when all required NodeClaims are ready, it will return true.
-func (c *NodeClaimManager) EnsureNodeClaims(ctx context.Context, wObj *kaitov1beta1.Workspace) (bool, error) {
+// DiffNodeClaims compares the current state of NodeClaims with the desired state
+func (c *NodeClaimManager) DiffNodeClaims(ctx context.Context, wObj *kaitov1beta1.Workspace) (bool, int, int, []*karpenterv1.NodeClaim, error) {
```
zhuangqh: is the bool variable here really necessary?

pkg/workspace/resource/node_claim.go
```diff
@@ -48,179 +49,203 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
+			return err
+		})
 
-			klog.InfoS("NodeClaim created successfully",
-				"nodeClaim", nodeClaim.Name,
-				"workspace", klog.KObj(wObj))
+		if !created {
+			// Failed to create, decrement expectations
+			c.expectations.CreationObserved(c.logger, workspaceKey)
+			if err != nil {
```
zhuangqh: is `err` always != nil if `created` is false?

pkg/workspace/resource/node_claim.go
```diff
@@ -48,179 +49,203 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
+func (c *NodeClaimManager) ScaleDownNodeClaims(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim, nodesToDelete int) (bool, error) {
+	workspaceKey := client.ObjectKeyFromObject(wObj).String()
+	klog.InfoS("Scaling down excess NodeClaims", "workspace", workspaceKey, "toDelete", nodesToDelete)
 
-			deletingI := nodeClaimI.DeletionTimestamp != nil
-			deletingJ := nodeClaimJ.DeletionTimestamp != nil
-			if deletingI != deletingJ {
-				return deletingI // being deleted comes first
+	if nodesToDelete == 0 {
+		if curCondition := meta.FindStatusCondition(wObj.Status.Conditions, string(kaitov1beta1.ConditionTypeScalingDownStatus)); curCondition != nil {
```
zhuangqh, line 162: you don't need to check if the condition exists, `UpdateStatusConditionIfNotMatch` will assign the condition.

pkg/workspace/resource/node_claim.go
```diff
@@ -48,179 +49,203 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
-			return condition.Status == "True"
+			c.recorder.Eventf(wObj, "Normal", "NodeClaimDeleted",
+				"Successfully deleted NodeClaim %s for workspace %s", nodeClaim.Name, wObj.Name)
+		} else {
+			c.expectations.DeletionObserved(c.logger, workspaceKey)
 		}
 	}
 
-	return nodeClaim.Status.NodeName != ""
+	return true, nil
```
zhuangqh: is the bool variable here really necessary? 
what about simply return error

pkg/workspace/resource/node_claim.go
rambohe-ch: yes, this bool return value is necessary, and will be used for judging `return reconcile directly` or `go ahead to scaleup/scaledown`.

pkg/workspace/resource/node_claim.go
rambohe-ch: fixed

pkg/workspace/resource/node_claim.go
rambohe-ch: Because ConditionTypeScalingDownStatus condition is only confgured for scaling down action, so this condition should not be set to true when it doesn't exist.

pkg/workspace/resource/node_claim.go
rambohe-ch: fixed

chewong (approved):
LGTM, pending e2e test

Issue #1393, opened 2025-08-13
Convert local-csi-driver to helm dependency instead of relying of git patches


chewong:
Could be a follow-up but I think we should use https://helm.sh/docs/helm/helm_dependency/ to take a dependency on https://github.com/Azure/local-csi-driver/tree/main/charts/latest as a best practice instead of using patch files (see #1252 as an example). However, It would involve updating local-csi-driver chart's templates and `values.yaml` to the point where we can customize container images, labels, volumes and volume mounts with helm values.

_Originally posted by @chewong in https://github.com/kaito-project/kaito/pull/1386#discussion_r2270743787_

wangxiaoxuan273:
The patch file removes some hard-coded labels such as this [one](https://github.com/Azure/local-csi-driver/blob/9e21c20aa8fd707f2918f5ca3e3abaf1842383e1/charts/latest/templates/daemonset.yaml#L7). Is removing labels like this possible when using `local-csi-driver` as a helm dependency? I don't think it's possible based on my limited experience.

Using `helm template` with patch may be the most feasible way to achieve the current level of customization. But if this issue is to be committed, I'm interested in working on this.

brutusbuckeye:
@chewong  Perhaps as an intermediate step to this, you could still customize the local-csi-driver as needed for KAITO, but add it as a proper helm  sub chart. This would allow users of Kaito to tweak values within the local-csi-driver subchart. This would also be a step closer to proper helm dependency support, while you work with the local-csi-driver team to integrate your changes.

The reason I like this approach is that it exposes the images that the local-csi-driver needs to Kaito users: https://github.com/Azure/local-csi-driver/blob/9e21c20aa8fd707f2918f5ca3e3abaf1842383e1/charts/latest/values.yaml#L5

As it stands today, these transitive image dependencies from local-csi-driver are hidden from users. This makes it very hard to use Kaito in environments where the MCR is not directly accessible and these containers must be mirrored

chewong:
@wangxiaoxuan273 thanks for your interest. @brutusbuckeye is right - defining local-csi-driver as a proper Helm dependency of the KAITO workspace chart lets users customize dependency values. Patch-based operations can be brittle and prone to merge conflicts during upgrades.

[...]

PR #1395, opened 2025-08-15
ci: aikit test

sozercan:
**Reason for Change**:

- add aikit test with llama3.2:1b image
- runs on a kind cluster
- this does not require any gpus, and can be pulled and run fast on default github runners (takes ~30 seconds from workspace deployment to LLM response)

**Requirements**

- [x] added unit tests and e2e tests (if applicable).

**Issue Fixed**:

Fixes #1375 

**Notes for Reviewers**:

PR #1483, opened 2025-09-09
chore: make local-csi-driver a helm dependency

wangxiaoxuan273:
**Reason for Change**:

Makes `local-csi-driver` a helm dependency and removed the sync script and the diff file. Note that only the DaemonSet's container image is customized. Other customizations specified in the diff file (changing labels, adding namespace etc.) is not possible.

**Requirements**

**Issue Fixed**:

Part of #1393 

**Notes for Reviewers**:

### DIFF
--- a/charts/kaito/workspace/Chart.lock
+++ b/charts/kaito/workspace/Chart.lock
@@ -5,5 +5,8 @@ dependencies:
 - name: gpu-feature-discovery
   repository: https://nvidia.github.io/k8s-device-plugin
   version: 0.18.0
-digest: sha256:16e3d87455ed85594037098f49ba96aba5fb693fbff6f1e81834e00a26effff5
-generated: "2025-10-21T14:40:13.868037273-04:00"
+- name: local-csi-driver
+  repository: oci://localcsidriver.azurecr.io/acstor/charts
+  version: 0.2.6
+digest: sha256:82a17788b188dead73501eac49fc06b1000efd373fd8a745bd0fb83f7dee02a3
+generated: "2025-10-28T15:13:17.399314887+08:00"


wangxiaoxuan273:
depends on https://github.com/Azure/local-csi-driver/pull/185

charts/kaito/workspace/values.yaml
```diff
@@ -19,6 +19,8 @@ featureGates:
   vLLM: true
   disableNodeAutoProvisioning: false
   gatewayAPIInferenceExtension: false
+localCSIDriver:
+  useLocalCSIDriver: false
```
zhuangqh: This dependency is not optional.

wangxiaoxuan273: Updated default value to true. https://github.com/kaito-project/kaito/pull/1395#discussion_r2279446471 is a case that this variable may need to be set false.

zhuangqh: alright, that scenario is for testing. default to true works for me.

wangxiaoxuan273:
Running `helm dependency update` gives `Error: chart file "pack-dd945e65e21d61920cd4b90d9f74b8bcdd331f09.pack" is larger than the maximum file size 5242880`. Not sure what I need to do with this.

chewong:
> Running `helm dependency update` gives `Error: chart file "pack-dd945e65e21d61920cd4b90d9f74b8bcdd331f09.pack" is larger than the maximum file size 5242880`. Not sure what I need to do with this.

Could you try `helm dependency update ./charts/kaito/workspace`?

wangxiaoxuan273:
> @wangxiaoxuan273 There is a new [release](https://github.com/Azure/local-csi-driver/releases/tag/v0.2.6) for local-csi-driver that includes your change. Could you update this PR to take a dependency on v0.2.6?

Updated. PTAL.

charts/kaito/workspace/values.yaml
```diff
@@ -57,4 +59,9 @@ gpu-feature-discovery:
   devicePlugin:
     enabled: false
   gfd:
-    enabled: true
\ No newline at end of file
+    enabled: true
+local-csi-driver:
```
jmclong, line 63: You need to set these values as well. They are currently being used in the sync script to disable webhooks and metrics

  --set webhook.enforceEphemeral.enabled=false \
  --set webhook.hyperconverged.enabled=false \
  --set observability.metrics.enabled=false

wangxiaoxuan273: Updated. PTAL again.

zhuangqh:
@wangxiaoxuan273 can you resolve the conflict? Once the test pass, I will merge this PR. Thanks for your contribution!

wangxiaoxuan273:
> `helm dependency update ./charts/kaito/workspace`?

Updated. PTAL.

jmclong (approved):

PR #1503, opened 2025-09-18
docs: Introduce a new InferenceSet CRD and Controller for scaling inference workloads automatically

andyzhangx:
**Reason for Change**:

doc: Introduce a new InferenceSet CRD and Controller for scaling inference workloads automatically, related to https://docs.google.com/presentation/d/1qCpFWuWzvoi5yA2rAa7OhCAUHPRoAmR3-ODJ3SDIgI4/edit?slide=id.g37fd1d8dc1a_0_22#slide=id.g37fd1d8dc1a_0_22

**Requirements**

- [ ] added unit tests and e2e tests (if applicable).

**Issue Fixed**:

**Notes for Reviewers**:

PR #1522, opened 2025-09-25
[WIP] chore: generate inferenceset structure

andyzhangx:
**Reason for Change**:

chore: generate inferenceset structure

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


api/v1alpha1/inferenceset_types.go
```diff
@@ -0,0 +1,83 @@
...
+// +kubebuilder:resource:path=inferencesets,scope=Namespaced,categories=inferenceset,shortName={is,isets}
+// +kubebuilder:storageversion
+// +kubebuilder:printcolumn:name="Replicas",type="integer",JSONPath=".spec.replicas",description=""
+// +kubebuilder:printcolumn:name="ReadyReplicas",type="integer",JSONPath=".status.readyReplicas",description=""
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description=""
+type InferenceSet struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   *InferenceSetSpec  `json:"spec,omitempty"`
```
zhuangqh: do we should use pointer here? It's not optional.

api/v1alpha1/inferenceset_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// InferenceSetTemplate defines the template for creating InferenceSet instances.
+type InferenceSetTemplate struct {
+	Resource  ResourceSpec   `json:"resource,omitempty"`
+	Inference *InferenceSpec `json:"inference,omitempty"`
```
zhuangqh: same. I think this field is not optional.

docs/proposals/20250918-introduce_inferenceset_autoscaling.md
```diff
@@ -0,0 +1,173 @@
...
+  replicas: 3 # number of workspace CR created by InferenceSet controller
+  nodeCountLimit: 10 # optional, total GPU node count limit for InferenceSet
+  selector:
+    matchLabels:
+      # workspace created by InferenceSet controller would use this label in resource.labelSelector
+      apps: large-model
+  template:
+    resource:
+      instanceType: "Standard_NC24ads_A100_v4"
+    inference: # fields in inference are the same as in workspace.resource.inference
```
zhuangqh:
```suggestion
    inference: # fields in inference are the same as in workspace.inference
```

api/v1alpha1/inferenceset_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// +kubebuilder:validation:Schemaless
+	// +kubebuilder:pruning:PreserveUnknownFields
+	UpdateStrategy appsv1.StatefulSetUpdateStrategy `json:"updateStrategy,omitempty"`
+}
+
+// InferenceSetStatus defines the observed state of InferenceSet
+type InferenceSetStatus struct {
+	// Conditions report the current conditions of the InferenceSet.
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
zhuangqh, line 68: do we need to add other status field for this controller?
like https://github.com/kaito-project/kaito/commit/f46c916d626e801a324868f24adf311ac7e5a279#diff-802d68cb9e30b05ed85f62332cadfe391e3ca842694c657c111c1d454316b045R110

zhuangqh: ref replicaset.status https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/replica-set-v1/#ReplicaSetStatus

docs/proposals/20250918-introduce_inferenceset_autoscaling.md
andyzhangx: fixed

api/v1alpha1/inferenceset_types.go
andyzhangx: I have added `Replicas` and  `Conditions` in `InferenceSetStatus` for now, we could add more during implementation if that's necessary. The InferenceSet controller should be only responsible to create workspace CR and when ensure `.status.TargetNodeCount  * replica <= nodeCountLimit` during creating multiple replicas of workspace CRs.

zhuangqh (approved):

zhuangqh:
@rambohe-ch  PTAL

andyzhangx:
> @rambohe-ch PTAL

@zhuangqh this PR includes the unmature API change which is still WIP, PTAL at this design doc RP: https://github.com/kaito-project/kaito/pull/1503, thanks

api/v1alpha1/inferenceset_types.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
+)
+
+type InferenceSetResourceSpec struct {
+	// InstanceType specifies the GPU node SKU.
+	// +required
```
Fei-Guo, line 25: If it is required, you should remove "omitempty" annotation.

charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -8,10 +8,16 @@ metadata:
 rules:
   - apiGroups: ["kaito.sh"]
     resources: ["workspaces"]
-    verbs: ["update", "patch","get","list","watch"]
+    verbs: ["update", "patch","get","list","watch", "create", "delete"]
```
Fei-Guo: I would suggest creating separate clusterrole, clusterrolebindings, sa for the new controller. The new clusterrole only includes the RBAC rules needed by the inferenceset controller.

pkg/workspace/controllers/metrics.go
```diff
@@ -97,3 +98,46 @@ func determineWorkspacePhase(ws *kaitov1beta1.Workspace) string {
...
+			var isList kaitov1alpha1.InferenceSetList
+
+			if err := k8sClient.List(ctx, &isList); err != nil {
+				klog.Errorf("failed to list all inferencesets: %v", err)
+				continue
+			}
+
+			for _, is := range isList.Items {
+				phase := determineInferenceSetPhase(&is)
+				klog.InfoS("InferenceSet phase", "inferenceset", klog.KObj(&is), "phase", phase)
```
Fei-Guo: Don't you need a Prometheus counter to record the number of inferenceset under different phases?

config/rbac/inferenceset_viewer_role_binding.yaml
```diff
@@ -0,0 +1,19 @@
...
+    app.kubernetes.io/managed-by: kustomize
+  name: inferenceset-viewer-role-rolebinding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: inferenceset-viewer-role
+subjects:
+- kind: ServiceAccount
+  name: controller-manager
+  namespace: system
```
rambohe-ch, line 19: would you like to explain the reason for configuring rbac for a new SA?

examples/inference/kaito_inferenceset_llama-3.1_8b_instruct.yaml
```diff
@@ -0,0 +1,21 @@
...
+      # workspace created by InferenceSet controller would use this label in resource.labelSelector
+      apps: large-model
+  template:
+    resource:
+      instanceType: "Standard_NC24ads_A100_v4"
+    inference: # fields in inference are the same as in workspace.resource.inference
+      preset:
+        name: llama-3.1-8b-instruct
+        presetOptions:
+          modelAccessSecret: hf-token
```
rambohe-ch, line 28: It seems that a secret named hf-token has been forgotten in this yaml file.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
+// Copyright (c) KAITO authors.
```
rambohe-ch, line 1: how about add a sub folder named inferenceset for managing `inferenceset_controller.go` file? so these controllers can be well categorized.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	for i := range wsList.Items {
+		if wsList.Items[i].DeletionTimestamp.IsZero() {
+			klog.InfoS("Deleting associated Workspace...", "workspace", wsList.Items[i].Name)
+			if deleteErr := c.Delete(ctx, &wsList.Items[i], &client.DeleteOptions{}); deleteErr != nil {
+				klog.ErrorS(deleteErr, "failed to delete the workspace", "workspace", klog.KObj(&wsList.Items[i]))
+				return ctrl.Result{}, deleteErr
+			}
+		}
+	}
+
```
rambohe-ch, line 145: Should we wait all workspaces deleted before removing finalizer from inferenceset?

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+	wsList, err := inferenceset.ListWorkspaces(ctx, isObj, c.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	klog.InfoS("Found workspaces for inference set", "name", isObj.Name, "current", len(wsList.Items), "desired", isObj.Spec.Replicas)
+
+	if len(wsList.Items) > isObj.Spec.Replicas {
+		// We should delete the extra workspaces that are created by this inferenceset
+		klog.InfoS("Found extra workspaces, deleting...", "current", len(wsList.Items), "desired", isObj.Spec.Replicas)
+		for i := isObj.Spec.Replicas; i < len(wsList.Items); i++ {
```
rambohe-ch: It's a good idea to sort the workspace list before deleting extra workspaces, like sort as follows:
1. delete `Terminating` workspace at first
2. then delete not ready workspace
3. then delete latest created workspace.
4. ...

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+
+		if createWorkspace {
+			klog.InfoS("creating workspace", "workspace", workspaceObj.Name, "index", i)
+			if err := c.Client.Create(ctx, workspaceObj); err != nil {
+				klog.ErrorS(err, "failed to create workspace", "workspace", workspaceObj.Name)
+				return reconcile.Result{}, err
+			}
+		} else {
+			klog.InfoS("updating workspace", "workspace", workspaceObj.Name, "index", i)
+			if err := c.Client.Update(ctx, workspaceObj); err != nil {
```
rambohe-ch: we should update the workspace when changed

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+
+	klog.InfoS("successfully removed the inferenceset finalizers", "inferenceset", klog.KObj(isObj))
+	return ctrl.Result{}, nil
+}
+
+func (c *InferenceSetReconciler) addOrUpdateInferenceSet(ctx context.Context, isObj *kaitov1alpha1.InferenceSet) (reconcile.Result, error) {
+	if isObj == nil {
+		return reconcile.Result{}, nil
+	}
+
```
rambohe-ch, line 166: `c.expectations` is unused

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+				klog.ErrorS(err, "failed to update workspace", "workspace", workspaceObj.Name)
+				return reconcile.Result{}, err
+			}
+		}
+	}
+
+	// update the replicas in the status
+	if err = inferenceset.UpdateInferenceSetStatus(ctx, c.Client, &client.ObjectKey{Name: isObj.Name, Namespace: isObj.Namespace}, func(status *kaitov1alpha1.InferenceSetStatus) error {
+		status.Replicas = isObj.Spec.Replicas
+		status.ReadyReplicas = isObj.Spec.Replicas
```
rambohe-ch: we should update Inferenceset.Status according to the workspaces in the WorkspaceList.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil
+	}); err != nil {
+		klog.ErrorS(err, "failed to update inferenceset replicas", "inferenceset", klog.KObj(isObj))
+		return reconcile.Result{}, err
+	}
+
+	if err = inferenceset.UpdateStatusConditionIfNotMatch(ctx, c.Client, isObj, kaitov1alpha1.InferenceSetConditionTypeSucceeded, metav1.ConditionTrue,
```
rambohe-ch: we should update InferenceSetConditionTypeSucceeded condition to true when all workspaces ready.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
rambohe-ch: maybe we can use the revision to verify workspace is changed or not

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,337 @@
...
+					return ctrl.Result{}, err
+				}
+			}
+		}
+	}
+
+	klog.InfoS("begin to create/update workspaces for inference set", "name", isObj.Name, "replicas", isObj.Spec.Replicas)
+	for i := 0; i < isObj.Spec.Replicas; i++ {
+		createWorkspace := false
+		workspaceName := isObj.Name + "-" + strconv.Itoa(i)
```
rambohe-ch: maybe we should use a rand digit string to build the workspace name, because we can not make sure scale down the workspace according to the digit order.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: good point, I think it's better using similar naming sequence as statefulset(use -0, -1, -2 ... index), that would be more determined when scale down. but anyway, need to check the worksapce name during workspace deleting when scaling down.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: I think it's better using similar naming sequence as statefulset(use -0, -1, -2 ... index)

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: there is only one inferenceset_controller.go file now, if there is more, it's better creating a sub folder.

config/rbac/inferenceset_viewer_role_binding.yaml
andyzhangx: it's bound to the same account as workspace_viewer_role

https://github.com/kaito-project/kaito/blob/c70caf1cf2151b58c6d92307f8fe77c5b5b64b45/config/rbac/workspace_viewer_role_binding.yaml#L17-L19

andyzhangx: I think I also need to create `inferenceset_editor_role.yaml` and `inferenceset_editor_role_binding.yaml`

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: after discussing, we would use statufulset way first since it's simple and more determined, and if there is really such requirements in the near future, we could add optional fields to change this default behavior.

api/v1alpha1/inferenceset_types.go
andyzhangx: fixed

charts/kaito/workspace/templates/clusterrole.yaml
andyzhangx: fixed, use different clusterrole, clusterrolebindings files now.

examples/inference/kaito_inferenceset_llama-3.1_8b_instruct.yaml
andyzhangx: fixed

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: fixed

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: fixed

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
andyzhangx: c.Delete is a sync operation, so I think it's ok not waiting, and I checked the logic of workspace deletion, it did not wait for nodeClaim deletion also. Anyway, that's not blocking issue now, we could refine this when it's needed later on.

api/v1alpha1/condition_types.go
```diff
@@ -41,6 +41,14 @@ const (
 	//WorkspaceConditionTypeDeleting is the Workspace state when starts to get deleted.
 	WorkspaceConditionTypeDeleting = ConditionType("WorkspaceDeleting")
 
+	//InferenceSetConditionTypeDeleting is the InferenceSet state when starts to get deleted.
+	InferenceSetConditionTypeDeleting = ConditionType("InferenceSetDeleting")
+
+	// InferenceSetConditionTypeReady is the InferenceSet state when starts to get ready.
+	InferenceSetConditionTypeReady = ConditionType("InferenceSetReady")
+
+	InferenceSetConditionTypeSucceeded ConditionType = ConditionType("InferenceSetSucceeded")
```
rambohe-ch: InferenceSetConditionTypeReady is unused, but I think we should use InferenceSetConditionTypeReady instead InferenceSetConditionTypeSucceeded, and remove InferenceSetConditionTypeSucceeded condition.

andyzhangx: fixed

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
zhuangqh: agree. index suffix is better.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,390 @@
...
+
+		workspaceObj.Labels = map[string]string{
+			consts.InferenceSetMemberLabel:             workspaceObj.Name,
+			consts.WorkspaceCreatedByInferenceSetLabel: isObj.Name,
+		}
+		workspaceObj.Resource = kaitov1beta1.ResourceSpec{
+			InstanceType:  isObj.Spec.Template.Resource.InstanceType,
+			LabelSelector: isObj.Spec.Selector,
+		}
+		workspaceObj.Inference = &isObj.Spec.Template.Inference
```
zhuangqh: set inferenceset object to be the owner of this workspace object.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,390 @@
...
+func (c *InferenceSetReconciler) garbageCollectInferenceSet(ctx context.Context, isObj *kaitov1alpha1.InferenceSet) (ctrl.Result, error) {
+	klog.InfoS("garbageCollectInferenceSet", "inferenceset", klog.KObj(isObj))
+	// Check if there are any workspaces associated with this inferenceset.
+	wsList, err := inferenceset.ListWorkspaces(ctx, isObj, c.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	// We should delete all the workspaces that are created by this inferenceset
+	for i := range wsList.Items {
+		if wsList.Items[i].DeletionTimestamp.IsZero() {
```
zhuangqh, line 137: set the inferenceset object as the owner of workspace objects. Then kubernetes gc controller will ensure that workspace objects are deleted once we place a deletiontimestamp at inference set.
It's a best practice for this use case .
workspace controller didn't do this is because nodeclaim is not a namespaced object.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		klog.ErrorS(err, "failed to update inferenceset status", "inferenceset", klog.KObj(isObj))
+		return reconcile.Result{}, err
+	}
+
+	return c.garbageCollectInferenceSet(ctx, isObj)
+}
+
+// garbageCollectInferenceSet remove finalizer associated with inferenceset object.
+func (c *InferenceSetReconciler) garbageCollectInferenceSet(ctx context.Context, isObj *kaitov1alpha1.InferenceSet) (ctrl.Result, error) {
```
zhuangqh: This abbreviation is confusing. `isObj`, I thought it's a boolean variable.
What about using `inferenceSet` literally?

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	klog.InfoS("Found workspaces for inference set", "name", isObj.Name, "current", len(wsList.Items), "desired", isObj.Spec.Replicas)
+
+	if len(wsList.Items) > isObj.Spec.Replicas {
+		// We should delete the extra workspaces that are created by this inferenceset
+		klog.InfoS("Found extra workspaces, deleting...", "current", len(wsList.Items), "desired", isObj.Spec.Replicas)
+		for i := isObj.Spec.Replicas; i < len(wsList.Items); i++ {
+			deleteWorkspaceName := isObj.Name + "-" + strconv.Itoa(i)
```
zhuangqh: please add an individual function to get this naming. It's a critical path for this functionality.
like https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_set_utils.go#L110

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	// Check if there are any existing workspaces associated with this inferenceset.
+	wsList, err := inferenceset.ListWorkspaces(ctx, isObj, c.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	klog.InfoS("Found workspaces for inference set", "name", isObj.Name, "current", len(wsList.Items), "desired", isObj.Spec.Replicas)
+
+	if len(wsList.Items) > isObj.Spec.Replicas {
```
zhuangqh: Let say I have a inference set with 5 replicas. If I scale it down to 4 replicas and delete the workspace with index 2 manually.
The behavior is
1. controller delete workspace-4.
2. recreate workspace-2

am I understand correctly?

api/v1alpha1/inferenceset_types.go
```diff
@@ -0,0 +1,102 @@
...
+	InstanceType string `json:"instanceType"`
+}
+
+// InferenceSetTemplate defines the template for creating InferenceSet instances.
+type InferenceSetTemplate struct {
+	Resource  InferenceSetResourceSpec   `json:"resource"`
+	Inference kaitov1beta1.InferenceSpec `json:"inference"`
```
zhuangqh, line 32: This type is defined in v1alpha1. But you import the type from v1beta1. 
I am not sure if this is a best practice. would like a confirmation from @Fei-Guo

Fei-Guo: I think it is ok.

pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+
+		// After deleting the extra workspaces, we should requeue to wait for the deletion to complete
+		if wsList, err = inferenceset.ListWorkspaces(ctx, isObj, c.Client); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	klog.InfoS("begin to create/update workspaces for inference set", "name", isObj.Name, "replicas", isObj.Spec.Replicas)
+	for i := 0; i < isObj.Spec.Replicas; i++ {
```
Fei-Guo: Each workspace in the inferenceSet is considered to be stateless, and the workspace order does not really matter for the functionality. In this case, I think the best practice is to name the workspace with random hash instead of index. 

The problem of having a fix index is that, if any workspace is not in ready state, from controller's perspective, a new workspace should be created to match the replica count. If you use fixed index as the name, this cannot be done. A fixed name in general will make dealing with corner cases harder.

andyzhangx: fixed, I have used random string suffix instead of index now.

pkg/utils/workspace/workspace.go
```diff
@@ -87,3 +153,22 @@ func UpdateWorkspaceWithRetry(ctx context.Context, c client.Client, wObj *kaitov
...
+	r := rand.New(source)
+	charset := "abcdefghijklmnopqrstuvwxyz0123456789"
+	result := make([]byte, length)
+	for i := range result {
+		result[i] = charset[r.Intn(len(charset))]
+	}
+	return string(result)
+}
+
+func GetWorkspaceNameWithRandomSuffix(baseName string) string {
```
andyzhangx, line 172: @zhuangqh FYI, I used `GetWorkspaceNameWithRandomSuffix` to get random workspace name

PR #1523, opened 2025-09-25
feat: [BREAKING] use statefulset for all workspace

zhuangqh:
**Reason for Change**:

- enable nvme disk acceleration for normal workspace.
   if nvme disk is unavailable, use emptyDir instead.
- cleanup deployment workload for existing workspace.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/workspace/controllers/workspace_controller.go
+++ b/pkg/workspace/controllers/workspace_controller.go
@@ -396,17 +396,6 @@ func ComputeHash(w *kaitov1beta1.Workspace) string {
 	return hex.EncodeToString(hasher.Sum(nil))
 }
 
-// getPresetName returns the preset name from wObj if available
-func getPresetName(wObj *kaitov1beta1.Workspace) string {
-	if wObj.Inference != nil && wObj.Inference.Preset != nil {
-		return string(wObj.Inference.Preset.Name)
-	}
-	if wObj.Tuning != nil && wObj.Tuning.Preset != nil {
-		return string(wObj.Tuning.Preset.Name)
-	}
-	return ""
-}
-
 func (c *WorkspaceReconciler) ensureService(ctx context.Context, wObj *kaitov1beta1.Workspace) error {
 	serviceType := corev1.ServiceTypeClusterIP
 	wAnnotation := wObj.GetAnnotations()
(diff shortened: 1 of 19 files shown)


pkg/workspace/controllers/workspace_controller.go
```diff
@@ -525,17 +525,8 @@ func (c *WorkspaceReconciler) applyInference(ctx context.Context, wObj *kaitov1b
...
-			var desiredPodSpec *corev1.PodSpec
-			switch workloadObj := workloadObj.(type) {
-			case *appsv1.StatefulSet:
-				existingObj = &appsv1.StatefulSet{}
-				desiredPodSpec = &workloadObj.Spec.Template.Spec
-			case *appsv1.Deployment:
-				existingObj = &appsv1.Deployment{}
-				desiredPodSpec = &workloadObj.Spec.Template.Spec
-			}
+			existingObj := &appsv1.StatefulSet{}
```
andyzhangx, line 534: so this PR would convert all deployments to statefulsets even for small models? and some azure vm GPU sku does not have local nvme disk, what would be the volume created on those skus? for k8s deployment based inference pod, it's using emptyDir in the before.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -151,41 +151,31 @@ func GeneratePresetInference(ctx context.Context, workspaceObj *v1beta1.Workspac
...
-		ssOpts = append(ssOpts, manifests.SetStatefulSetPodSpec(podSpec))
+	ssOpts := []generator.TypedManifestModifier[generator.WorkspaceGeneratorContext, appsv1.StatefulSet]{
+		manifests.GenerateStatefulSetManifest(revisionNum, numNodes),
+	}
 
-		return generator.GenerateManifest(gctx, ssOpts...)
+	if checkIfNVMeAvailable(ctx, gpuConfig, kubeClient) {
+		ssOpts = append(ssOpts, manifests.AddStatefulSetVolumeClaimTemplates(GenerateModelWeightsCacheVolume(ctx, workspaceObj, model)))
 	} else {
 		podOpts = append(podOpts, SetDefaultModelWeightsVolume)
```
andyzhangx, line 169: if nvme disk is not available, it just downloads mode weights into local path inside container now, right? why we use emptyDir in deployment in the before?

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -518,6 +487,24 @@ func (c *WorkspaceReconciler) applyTuning(ctx context.Context, wObj *kaitov1beta
...
+		klog.InfoS("Delete existing deployment workload for workspace", "workspace", klog.KObj(wObj))
+		err = c.Delete(ctx, &existingDeploy, &client.DeleteOptions{
+			Preconditions: &metav1.Preconditions{
+				UID: &existingDeploy.UID,
+			},
+		})
+		if err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to delete old inference deployment: %w", err)
+		}
+	}
```
Fei-Guo, line 508: If the GetResource call fails, we should return the err.

pkg/workspace/inference/preset_inferences.go
zhuangqh: what's the problem of using emptyDir😂

andyzhangx: that's just an ask, I think you also need to mention this in the PR description about this storage volume change.

andyzhangx (approved):

Fei-Guo (approved):

PR #1536, opened 2025-09-30
feat: add support for generic BYO nodes using NVIDIA GPU feature discovery 

### LINKED ISSUE #1222: generic way to handle model scheduling

### LINKED ISSUE #1280: kaito should not default instanceType
currently, kaito defaults to `Standard_NC24ads_A100_v4`. not only this is very Azure specific, it is also going to be sunset sooner or later

Jont828:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

Fixes #1222 
Fixes #1280

**Notes for Reviewers**:

### DIFF
--- a/pkg/workspace/estimator/advancednodesestimator/estimator_test.go
+++ b/pkg/workspace/estimator/advancednodesestimator/estimator_test.go
@@ -21,8 +21,8 @@ import (
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 
 	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
@@ -340,10 +340,10 @@ func TestAdvancedNodesEstimator_EstimateNodeCount_BYO(t *testing.T) {
 						Name: "byo-gpu-node",
 						Labels: map[string]string{
 							"node.kubernetes.io/instance-type": "Standard_NC96ads_A100_v4",
-							"kubernetes.azure.com/accelerator":  "nvidia-tesla-a100",
-							"nvidia.com/gpu.product":            "Tesla-A100-SXM4-80GB",
-							"nvidia.com/gpu.count":              "4",
-							"nvidia.com/gpu.memory":             "81920", // 80GB in MiB
+							"kubernetes.azure.com/accelerator": "nvidia-tesla-a100",
+							"nvidia.com/gpu.product":           "Tesla-A100-SXM4-80GB",
+							"nvidia.com/gpu.count":             "4",
+							"nvidia.com/gpu.memory":            "81920", // 80GB in MiB
 						},
 					},
 					Status: corev1.NodeStatus{
--- a/pkg/workspace/resource/node_claim.go
+++ b/pkg/workspace/resource/node_claim.go
@@ -30,7 +30,6 @@ import (
 	"github.com/kaito-project/kaito/pkg/utils/consts"
 	"github.com/kaito-project/kaito/pkg/utils/nodeclaim"
 	"github.com/kaito-project/kaito/pkg/utils/plugin"
-	"github.com/kaito-project/kaito/pkg/utils/resources"
 	"github.com/kaito-project/kaito/pkg/utils/workspace"
 )
 
@@ -140,12 +139,6 @@ func (c *NodeClaimManager) CreateUpNodeClaims(ctx context.Context, wObj *kaitov1
 
 // EnsureNodeClaimsReady is used for checking the number of ready nodeclaims(isNodeClaimReadyNotDeleting) meet the target NodeClaim count needed. Updates the
 func (c *NodeClaimManager) EnsureNodeClaimsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, readyNodes []*corev1.Node, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
-	// Since NodeClaims are only used when NAP is enabled, we know that all target nodes will be provisioned by NodeClaims instead of BYO nodes.
-	readyNodes, err := resources.GetReadyNodes(ctx, c.Client, wObj)
-	if err != nil {
-		return false, fmt.Errorf("failed to list ready nodes: %w", err)
-	}
-
 	targetNodeClaimCount := c.GetNumNodeClaimsNeeded(ctx, wObj, readyNodes)
 
 	readyCount := 0
--- a/pkg/ragengine/controllers/preset_rag.go
+++ b/pkg/ragengine/controllers/preset_rag.go
@@ -116,7 +116,7 @@ func CreatePresetRAG(ctx context.Context, ragEngineObj *v1alpha1.RAGEngine, revi
 	var resourceReq corev1.ResourceRequirements
 
 	if ragEngineObj.Spec.Embedding.Local != nil {
-		var gpuConfig *sku.GPUConfig = nil
+		var gpuConfig *sku.GPUConfig
 		var err error
 		gpuConfig, err = utils.GetGPUConfigBySKU(ragEngineObj.Spec.Compute.InstanceType)
 		if err != nil {
--- a/pkg/utils/test/test_utils.go
+++ b/pkg/utils/test/test_utils.go
@@ -141,7 +141,7 @@ var (
 		},
 		Spec: &v1alpha1.RAGEngineSpec{
 			Compute: &v1alpha1.ResourceSpec{
-				Count: &gpuNodeCount,
+				Count:        &gpuNodeCount,
 				InstanceType: "Standard_NC12s_v3",
 				LabelSelector: &metav1.LabelSelector{
 					MatchLabels: map[string]string{
--- a/pkg/workspace/estimator/basicnodesestimator/estimator_test.go
+++ b/pkg/workspace/estimator/basicnodesestimator/estimator_test.go
@@ -21,8 +21,8 @@ import (
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
 
 	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
@@ -388,10 +388,10 @@ func TestBasicNodesEstimator_EstimateNodeCount_BYO(t *testing.T) {
 						Name: "byo-gpu-node",
 						Labels: map[string]string{
 							"node.kubernetes.io/instance-type": "Standard_NC96ads_A100_v4",
-							"kubernetes.azure.com/accelerator":  "nvidia-tesla-a100",
-							"nvidia.com/gpu.product":            "Tesla-A100-SXM4-80GB",
-							"nvidia.com/gpu.count":              "4",
-							"nvidia.com/gpu.memory":             "81920", // 80GB in MiB
+							"kubernetes.azure.com/accelerator": "nvidia-tesla-a100",
+							"nvidia.com/gpu.product":           "Tesla-A100-SXM4-80GB",
+							"nvidia.com/gpu.count":             "4",
+							"nvidia.com/gpu.memory":            "81920", // 80GB in MiB
 						},
 					},
 					Status: corev1.NodeStatus{
--- a/pkg/workspace/resource/node_claim_test.go
+++ b/pkg/workspace/resource/node_claim_test.go
@@ -188,7 +188,7 @@ func TestCheckNodeClaims(t *testing.T) {
 			},
 			readyNodes: []*corev1.Node{
 				// 1 ready BYO node (does not have karpenter.sh/nodepool label)
-				&corev1.Node{
+				{
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "byo-ready-node-1",
 						Labels: map[string]string{
--- a/pkg/workspace/resource/node_test.go
+++ b/pkg/workspace/resource/node_test.go
@@ -1157,8 +1157,8 @@ func TestEnsureNodesReady(t *testing.T) {
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "node-1",
 						Labels: map[string]string{
-							"workload":                         "test",
-							corev1.LabelInstanceTypeStable:     "Standard_NC6s_v3", // Wrong instance type - logged but doesn't fail
+							"workload":                     "test",
+							corev1.LabelInstanceTypeStable: "Standard_NC6s_v3", // Wrong instance type - logged but doesn't fail
 						},
 					},
 					Status: corev1.NodeStatus{
@@ -1199,8 +1199,8 @@ func TestEnsureNodesReady(t *testing.T) {
 					ObjectMeta: metav1.ObjectMeta{
 						Name: "node-1",
 						Labels: map[string]string{
-							"workload":                         "test",
-							corev1.LabelInstanceTypeStable:     "Standard_NC12s_v3", // Correct instance type
+							"workload":                     "test",
+							corev1.LabelInstanceTypeStable: "Standard_NC12s_v3", // Correct instance type
 						},
 					},
 					Status: corev1.NodeStatus{


api/v1beta1/workspace_validation.go
```diff
@@ -319,80 +326,87 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
-								machineTotalNumGPUs.String(),
-								presetName,
-								modelGPUCount.Value(),
-							),
-							"instanceType",
-						))
+		// Check if instancetype exists in our SKUs map for the particular cloud provider
+		if skuConfig := skuHandler.GetGPUConfigBySKU(instanceType); skuConfig != nil {
+			if runtime != model.RuntimeNameVLLM {
+				if presetName != "" {
```
Jont828: There's an existing check here, can preset name ever be empty or can we remove this?

chewong: Sure you can remove it

api/v1beta1/workspace_validation.go
```diff
@@ -404,6 +418,169 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
+	}
+
+	// Step 3: Verify uniformity - all nodes must have identical nvidia.com/gpu.product, nvidia.com/gpu.count and nvidia.com/gpu.memory
+	var referenceGPUProduct, referenceGPUCount, referenceGPUMemory string
+	var referenceGPUCountPerNode, referenceGPUMemoryInt int64
+
+	for i, node := range nodeList.Items {
+		gpuProduct, hasGPUProduct := node.Labels["nvidia.com/gpu.product"]
+		gpuCount, hasGPUCount := node.Labels["nvidia.com/gpu.count"]
+		gpuMemory, hasGPUMemory := node.Labels["nvidia.com/gpu.memory"]
```
chewong: Let's reuse the constants from pkg/utils/consts/consts.go

charts/kaito/workspace/values.yaml
chewong, line 1: Do we need to add a new section for gpu-feature-discovery's values?

```yaml
...
gpu-feature-discovery:
  devicePlugin:
    enabled: false
  gfd
    enabled: true
...
```

charts/kaito/workspace/Chart.yaml
```diff
@@ -37,3 +37,11 @@ dependencies:
   version: "2.16.3"
   repository: https://fluxcd-community.github.io/helm-charts
   condition: featureGates.gatewayAPIInferenceExtension
+- name: nvidia-device-plugin
+  version: 0.17.2
```
chewong: Let's bump it to the latest, which is 0.17.4

pkg/utils/common.go
```diff
@@ -194,6 +211,70 @@ func TryGetGPUConfigFromNode(ctx context.Context, kubeClient client.Client, work
...
+	// Convert from MiB to GB: 1 MiB = 1.048576 MB, 1 GB = 1000 MB
+	// So 1 MiB = 1.048576/1000 GB = 0.001048576 GB
+	// Round to nearest integer for compatibility with GPUConfig.GPUMemGB int type
+	gpuMemGB := int((float64(gpuMemoryMiB) * 1.048576 / 1000) + 0.5)
+
+	return &sku.GPUConfig{
+		SKU:      "unknown", // SKU is not available from node labels
+		GPUCount: gpuCount,
+		GPUModel: gpuProduct,
+		GPUMemGB: gpuMemGB,
```
chewong: GPU memory is measured in GiB, not GB, despite the name of the variable. We should rename this field to `GPUMemGiB` to avoid confusion. That said, you shouldn't need to convert between GiB and GB. Just keep everything in GiB.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -111,6 +111,8 @@ func (c *WorkspaceReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 
 	klog.InfoS("Reconciling", "workspace", req.NamespacedName)
 
+	klog.InfoS("Disable NAP flag is", "value", featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning])
```
chewong: I don't think we need to log this every time we are reconciling the workspace

pkg/workspace/estimator/advancednodesestimator/estimator.go
```diff
@@ -52,12 +54,39 @@ func (c *AdvancedNodesEstimator) EstimateNodeCount(ctx context.Context, workspac
...
+			gpuConfig = &sku.GPUConfig{
+				GPUCount: 1,
+				GPUMemGB: 0, // Unknown memory, will skip memory-based optimization
+			}
+		}
+	} else {
+		// NAP is enabled - instanceType is required and must be valid
+		if workspace.Resource.InstanceType == "" {
+			return 0, fmt.Errorf("instanceType is required when node auto-provisioning is enabled")
+		}
```
chewong: The validating webhook should already take care of this scenario, so we don't have to include another if statement here

pkg/workspace/estimator/advancednodesestimator/estimator_test.go
```diff
@@ -232,6 +256,113 @@ func TestAdvancedNodesEstimator_EstimateNodeCount(t *testing.T) {
...
+	}{
+		{
+			name: "Should fallback to BYO logic for invalid instance type (NAP disabled)",
+			workspace: &kaitov1beta1.Workspace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "test-workspace",
+					Namespace: "default",
+				},
+				Resource: kaitov1beta1.ResourceSpec{
+					Count:        ptr.To(1),
```
chewong: nit: we should stop setting `.resource.count` when possible since the field is already depcreated

pkg/utils/common.go
```diff
@@ -194,6 +211,70 @@ func TryGetGPUConfigFromNode(ctx context.Context, kubeClient client.Client, work
 		SKU:      "unknown", // SKU is not available from nodes
 		GPUCount: skuGPUCount,
 		GPUModel: "unknown", // GPU model is not available from nodes
+		GPUMemGB: 0,         // GPU memory is not available from capacity
+	}, nil
+}
+
+func TryGetGPUConfigFromNodeList(ctx context.Context, workerNodes []*corev1.Node) (*sku.GPUConfig, error) {
```
chewong: Why create a new but similar function to TryGetGPUConfigFromNode?

charts/kaito/workspace/values.yaml
Jont828: I think it seems to work on its own as is. I don't think it's attempting to install device plugin twice based on what we see in the Helm template.

pkg/utils/common.go
Jont828: Oops thought I already merged those together

.github/workflows/aikit.yaml
```diff
@@ -84,7 +84,7 @@ jobs:
           # Install via Helm
           # CSI Local Node cannot be configured to not deploy and it crashes on kind on github runners for unknown reasons
           # so we need to filter it out its components
-          helm template kaito-workspace ./charts/kaito/workspace --namespace ${{ env.KAITO_NAMESPACE }} --include-crds --debug --wait | awk '
+          helm template kaito-workspace ./charts/kaito/workspace --namespace ${{ env.KAITO_NAMESPACE }} --set disableNodeAutoProvisioning="true" --include-crds --debug --wait | awk '
```
chewong:
```suggestion
          helm template kaito-workspace ./charts/kaito/workspace --namespace ${{ env.KAITO_NAMESPACE }} --set featureGates.disableNodeAutoProvisioning="true" --include-crds --debug --wait | awk '
```

api/v1beta1/workspace_validation.go
```diff
@@ -404,6 +417,169 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
+		}
+	}
+
+	// If there were uniformity errors, don't proceed with capacity checks
+	if errs != nil {
+		return errs
+	}
+
+	// Step 4: Calculate total GPU memory per node
+	referenceGPUMemoryBytes := referenceGPUMemoryInt * 1024 * 1024 // Note: nvidia.com/gpu.memory is in MiB not GiB
```
chewong: nit:

```suggestion
	referenceGPUMemoryBytes := referenceGPUMemoryInt * consts.MiBToBytes // Note: nvidia.com/gpu.memory is in MiB not GiB
```

pkg/utils/common.go
```diff
@@ -184,16 +185,63 @@ func GetGPUConfigBySKU(instanceType string) (*sku.GPUConfig, error) {
...
+		return nil, fmt.Errorf("invalid nvidia.com/gpu.count value on node %s: %s", node.Name, gpuCountStr)
+	}
+
+	// Parse GPU memory (nvidia.com/gpu.memory is in MiB, convert to GB)
+	gpuMemoryMiB, err := strconv.Atoi(gpuMemoryStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid nvidia.com/gpu.memory value on node %s: %s", node.Name, gpuMemoryStr)
+	}
+
+	gpuMemGiB := int((float64(gpuMemoryMiB) / 1024) + 0.5)
```
chewong, line 214: Why do we need to add an extra 0.5 GiB?

pkg/workspace/resource/node.go
```diff
@@ -127,23 +128,83 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
+
+	for _, node := range nodeList.Items {
+		if resources.NodeIsReadyAndNotDeleting(&node) {
+			readyCount++
+		}
+	}
+
+	defer func() {
+		// Update the list of nodes, but don't block setting the condition on it.
+		if err := c.UpdateWorkerNodesInStatus(ctx, wObj, nodeList.Items); err != nil {
```
chewong: If a workspace only deploys to 1 node with sufficient GPU memory but `nodeList` contains more than 1 node (due to matching labels), are we going to write all the matching nodes to `.status.workerNodes` field? Personally, I find it confusing where some part of the codebase relies on `.status.workerNodes` and some simply call `ListNode` to get the list of nodes.

pkg/workspace/resource/node_claim.go
```diff
@@ -48,30 +48,27 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
 	}
 
 	nodeClaims := make([]*karpenterv1.NodeClaim, 0, len(ncList.Items))
 	for i := range ncList.Items {
 		nodeClaims = append(nodeClaims, &ncList.Items[i])
 	}
 
-	return addedNodeClaimsCount, nodeClaims, readyNodes, nil
+	return addedNodeClaimsCount, nodeClaims, nil
 }
```
chewong: Are we no longer considering existing nodes that match the label selector and instance type when calculating NodeClaim count? You refactored this function so that it returns the difference between targeting node count and the count of existing nodeclaims

pkg/utils/common.go
Jont828: Since we get the memory in MiB and convert to GiB and store as an int, we add 0.5 and cast to int to round to the nearest whole number.

chewong: Can be a follow-up but it's probably better to refactor `GPUConfig` struct to use `resource.Quantity` instead of `int` for GPU memory

pkg/workspace/resource/node_claim.go
chewong: @rambohe-ch @Fei-Guo are we still planning to support this scenario, or we want to make BYO and NAP mutually exclusive?

Jont828: Yes the idea is that since BYO and NAP are mutually exclusive, we can assume that that if we're making node claims, we're in the NAP case, in which case all nodes will come from node claims.

I'm pretty sure @Fei-Guo and I talked about making NAP and BYO to be mutually exclusive to simplify the logic, but yea let's double check.

pkg/workspace/resource/node.go
Jont828: Yeah I agree it's a bit awkward. It seems like the `status.workerNodes` list was designed with NAP in mind, since for NAP if `# selected nodes == # nodes needed`, we can guarantee each node is running the model. But like you mentioned, for BYO, if  `# selected nodes == # nodes needed`, we can't easily figure out which nodes K8s scheduled the model to, so I deferred to listing the possible nodes for now.

The parts that rely on `.status.workerNodes` instead of `ListNode` are from the NAP case, and I can follow up to refactor that in a separate PR so this one doesn't get too big.

api/v1beta1/workspace_validation.go
```diff
@@ -404,6 +417,169 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
+//  5. If the preset model does not support multi-node distributed inference, fail if the total GPU memory per node
+//     is less than the preset's total GPU memory requirement
+func (r *ResourceSpec) validateBYONodeSelection(presetName string) (errs *apis.FieldError) {
+	if r.LabelSelector == nil || r.LabelSelector.MatchLabels == nil {
+		errs = errs.Also(apis.ErrMissingField("labelSelector.matchLabels is required for BYO node selection"))
+		return errs
+	}
+
+	// Access Kubernetes client for node operations
+	if k8sclient.Client == nil {
```
Fei-Guo: You can remove this nil check and call k8sclient.GetGlobalClient() instead. There is an assumption that this client should be valid.

pkg/utils/common.go
Jont828: Okay yea let's do a follow up, PR is already big enough and don't want to scare people off from reviewing.

api/v1beta1/workspace_validation.go
```diff
@@ -404,6 +417,169 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
+		errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Failed to list nodes with labelSelector: %v", err)))
+		return errs
+	}
+
+	if len(nodeList.Items) == 0 {
+		errs = errs.Also(apis.ErrGeneric("No nodes found matching the labelSelector"))
+		return errs
+	}
+
+	// Step 3: Verify uniformity - all nodes must have identical nvidia.com/gpu.product, nvidia.com/gpu.count and nvidia.com/gpu.memory
```
Fei-Guo: I am wondering if this should be a must or an optional check? For existing users who cannot install NFD for some reason, this will be a breaking change. 

If those labels do not exist, maybe we can give a warning.

api/v1beta1/workspace_validation.go
```diff
@@ -404,6 +417,169 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
+			}
+			if gpuMemory != referenceGPUMemory {
+				errs = errs.Also(apis.ErrGeneric(fmt.Sprintf("Non-uniform GPU memory: node %s has %s GB memory, but reference node has %s GB memory", node.Name, gpuMemory, referenceGPUMemory)))
+			}
+		}
+	}
+
+	// If there were uniformity errors, don't proceed with capacity checks
+	if errs != nil {
+		return errs
```
Fei-Guo: Again, should we give a warning and skip the rest of the checks?

api/v1beta1/workspace_validation.go
```diff
@@ -404,6 +417,169 @@ func (r *ResourceSpec) validateCreateWithInference(inference *InferenceSpec, byp
...
+	totalGPUMemRequired := resource.MustParse(params.TotalSafeTensorFileSize)
+	totalGPUMemRequiredBytes := totalGPUMemRequired.Value()
+
+	// Calculate total resources available across all matching nodes
+	numNodes := len(nodeList.Items)
+	numGPUAcrossNodes := resource.NewQuantity(int64(numNodes*int(referenceGPUCountPerNode)), resource.DecimalSI)
+	totalGPUMemAcrossNodes := resource.NewQuantity(int64(numNodes)*totalGPUMemoryPerNode, resource.BinarySI) // Total GPU memory
+
+	// Check if there are enough GPUs across all nodes
+	if numGPUAcrossNodes.Cmp(numGPURequired) < 0 {
```
Fei-Guo: Is it possible to use a common method for the following checks to avoid dup code in both BYO and NAP checks? I know we need to add another check for BYO to check the node count, which is not needed in NAP though. This check will be BYO specific.

pkg/utils/common.go
```diff
@@ -184,16 +185,63 @@ func GetGPUConfigBySKU(instanceType string) (*sku.GPUConfig, error) {
...
+
+	// Parse GPU count
+	gpuCount, err := strconv.Atoi(gpuCountStr)
+	if err != nil {
+		return nil, fmt.Errorf("invalid nvidia.com/gpu.count value on node %s: %s", node.Name, gpuCountStr)
+	}
+
+	// Parse GPU memory (nvidia.com/gpu.memory is in MiB, convert to GB)
+	gpuMemoryMiB, err := strconv.Atoi(gpuMemoryStr)
+	if err != nil {
```
Fei-Guo, line 210: The err msg is ignored in the caller side, maybe we can print a log for this and the following conversion error cases.

pkg/utils/common.go
Fei-Guo: I don't quite understand, if float64(gpuMemoryMiB) / 1024) is 0.2, int(0.7) is still 0, right? You should use round() method?

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -144,46 +144,76 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
 	return nil
 }
 
-func (c *WorkspaceReconciler) addOrUpdateWorkspace(ctx context.Context, wObj *kaitov1beta1.Workspace) (reconcile.Result, error) {
+func (c *WorkspaceReconciler) reconcileNodes(ctx context.Context, wObj *kaitov1beta1.Workspace) (reconcile.Result, error) {
 	workspaceKey := client.ObjectKeyFromObject(wObj).String()
 	if !c.expectations.SatisfiedExpectations(c.Log, workspaceKey) {
```
Fei-Guo: You can remove this check? since you already check it in line 203.

pkg/workspace/estimator/advancednodesestimator/estimator.go
```diff
@@ -52,12 +54,43 @@ func (c *AdvancedNodesEstimator) EstimateNodeCount(ctx context.Context, workspac
...
-	}
-	if gpuConfig == nil {
-		return 0, fmt.Errorf("GPU config is nil for instance type %s", workspace.Resource.InstanceType)
+	var gpuConfig *sku.GPUConfig
+	var err error
+
+	if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+		// NAP is disabled (BYO scenario) - instanceType is optional
+		// Try to get GPU config from existing nodes
+		if client != nil {
```
Fei-Guo: I think you should remove the nil check. If it is nil, we should crash the program and debug.

charts/kaito/workspace/Chart.yaml
```diff
@@ -37,3 +37,11 @@ dependencies:
   version: "2.16.3"
   repository: https://fluxcd-community.github.io/helm-charts
   condition: featureGates.gatewayAPIInferenceExtension
+- name: nvidia-device-plugin
```
Fei-Guo: Can we just install FGD and keep the device plugin as the old yaml for backward compatibility as least for now?

pkg/workspace/inference/preset_inferences.go
```diff
@@ -352,14 +382,14 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
...
 						// First check if user provided explicit value in ConfigMap
-						//if v, ok2 := utils.ParseExplicitMaxModelLen(raw); ok2 {
-						//maxModelLen = v
-						//klog.Infof("[RuntimeContext] workspace=%s using user explicit max-model-len=%d", ctx.Workspace.Name, maxModelLen)
-						//} else {
-						// If no user value, compute planned value
-						maxModelLen = computeMaxModelLen(presetParams, gpuConfig, numNodes)
-						klog.Infof("[RuntimeContext] workspace=%s using computed max-model-len=%d (gpuConfig=%+v, numNodes=%d)", ctx.Workspace.Name, maxModelLen, *gpuConfig, numNodes)
-						//}
+						if v, ok2 := utils.ParseExplicitMaxModelLen(raw); ok2 {
```
Fei-Guo: @bangqipropel needs to comment what the right change should be. This old logic seems to be problematic to me.

pkg/workspace/resource/node.go
```diff
@@ -41,27 +42,34 @@ func NewNodeManager(c client.Client) *NodeManager {
 	}
 }
 
-// AreNodePluginsReady is used for ensuring node label(accelerator:nvidia) and GPU capacity on all auto-provisioned nodes for the workspace.
-func (c *NodeManager) AreNodePluginsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
+// SetNodePluginsReadyCondition is used for ensuring node label(accelerator:nvidia) and GPU capacity on all auto-provisioned nodes for the workspace.
+func (c *NodeManager) SetNodePluginsReadyCondition(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
```
Fei-Guo: Maybe rename to "EnsureNodePluginsReady".

pkg/workspace/resource/node.go
```diff
@@ -127,23 +140,83 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
-			return err
+			return nil
 		}
 	}
 
+	return nil
+}
+
+// SetResourceReadyCondition updates the worker nodes list in workspace status.
+func (c *NodeManager) SetResourceReadyCondition(ctx context.Context, wObj *kaitov1beta1.Workspace) error {
```
Fei-Guo: It is ok to refactor the code for this, but the name is confusing, maybe 

 SetResourceReadyCondition ->  SetResourceReadyConditionByStatus

pkg/workspace/resource/node.go
```diff
@@ -127,23 +140,83 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
+			klog.ErrorS(updateErr, "failed to update Node status condition NodesReady to false", "workspace", klog.KObj(wObj))
+			return false, fmt.Errorf("failed to update Node status condition(NodeNotReady): %w", updateErr)
+		}
+
+		return false, nil
+	}
+}
+
+// propagateOwnedConditions checks the owned conditions, and if any one is false, it sets the condition to false with the reason and message of the first false condition found.
+func (c *NodeManager) propagateOwnedConditions(ctx context.Context, wObj *kaitov1beta1.Workspace, condition kaitov1beta1.ConditionType, conditionTypes []kaitov1beta1.ConditionType) error {
```
Fei-Guo: propagateOwnedConditions-> SetConditionForSpecificTypes.

pkg/workspace/resource/node_claim.go
Fei-Guo: Yes, mutually exclusive.  I cannot imagine how many more logic needs to be added if we want to make them work together.

pkg/workspace/resource/node_claim.go
```diff
@@ -116,13 +113,10 @@ func (c *NodeClaimManager) CreateUpNodeClaims(ctx context.Context, wObj *kaitov1
...
 
-// AreNodeClaimsReady is used for checking the number of ready nodeclaims(isNodeClaimReadyNotDeleting) meet the target NodeClaim count needed
-func (c *NodeClaimManager) AreNodeClaimsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
-	// Calculate the actual number of NodeClaims needed (if BYO, numtargetNodes == 0)
-	_, targetNodeClaimCount, err := nodeclaim.ResolveReadyNodesAndTargetNodeClaimCount(ctx, c.Client, wObj)
-	if err != nil {
-		return false, fmt.Errorf("failed to resolve target NodeClaim count: %w", err)
-	}
+// SetNodeClaimsReadyCondition is used for checking the number of ready nodeclaims(isNodeClaimReadyNotDeleting) meet the target NodeClaim count needed. Updates the
+func (c *NodeClaimManager) SetNodeClaimsReadyCondition(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
```
Fei-Guo: SetNodeClaimsReadyCondition -> EnsureNodeClaimsReady

pkg/workspace/inference/preset_inferences.go
bangqipropel: Please don't uncomment it, if you want, just remove it

charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
```diff
@@ -65,4 +65,4 @@ spec:
       volumes:
         - name: device-plugin
           hostPath:
-            path: /var/lib/kubelet/device-plugins
+            path: /var/lib/kubelet/device-plugins
```
bangqipropel: what is the change here?

pkg/workspace/estimator/basicnodesestimator/estimator.go
```diff
@@ -50,12 +53,42 @@ func (e *BasicNodesEstimator) EstimateNodeCount(ctx context.Context, wObj *kaito
...
-		return 0, fmt.Errorf("failed to get GPU config for instance type %s: %w", wObj.Resource.InstanceType, err)
-	}
-	if gpuConfig == nil {
-		return 0, fmt.Errorf("GPU config is nil for instance type %s", wObj.Resource.InstanceType)
+	// Import featuregates and consts for NAP check
+	var gpuConfig *sku.GPUConfig
+	var err error
+
+	// Check if NAP (Node Auto Provisioning) is enabled or disabled
+	napDisabled := featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning]
```
bangqipropel: My question: I think since they are the same feature, so basicnodeestimator should have the same change as advanced, but seems like they are quite different, may I know the reason?

charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
chewong: Missing a next-line character

pkg/utils/common.go
Jont828: I think for future maintainability, it's better to return an error and log it than to log in the function and not return one at all. That way, if we ever want to use the function and return w/ error if it fails, we can do so w/o any code changes.

api/v1beta1/workspace_validation.go
Jont828: I think it's fair to have a contract on what info the nodes provide about their specs, otherwise we'd be allowing heterogeneous nodes and IMO that's too confusing. If they can't install GFD, in the worst case, they can just label the nodes manually with the correct info as it's only 3 labels we care about. If the user is bringing them, it's fair to assume they know the product, GPU count, and memory.

pkg/workspace/inference/preset_inferences.go
Jont828: Sure, but why are we removing the option to set max model len using the configmap btw? I'm pretty sure I uncommented it as I got OOM when deploying Phi 4 mini on an A10, so I had to set the max model len smaller.

pkg/workspace/estimator/basicnodesestimator/estimator.go
Jont828: Should be the same now!

chewong (commented):
This PR should be ready for a re-review once rebased

test/e2e/utils/nodeclaim.go
```diff
@@ -42,14 +42,23 @@ func ValidateNodeClaimCreation(ctx context.Context, workspaceObj *v1beta1.Worksp
 			}
 
 			if len(nodeClaimList.Items) != expectedCount {
+				for _, nodeClaim := range nodeClaimList.Items {
```
Jont828: Will remove once e2e tests pass and no longer need for debugging.

test/e2e/preset_test.go
```diff
@@ -371,13 +380,31 @@ func updatePhi3TuningWorkspaceWithPresetPublicMode(workspaceObj *kaitov1beta1.Wo
 }
 
 func updateAndValidateWorkspace(workspaceObj *kaitov1beta1.Workspace) {
-	By("Creating workspace", func() {
+	By("Updating workspace", func() {
```
Jont828, line 383: Will double check this on e2e tests.

pkg/workspace/resource/node.go
```diff
@@ -127,23 +140,84 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
+	}
+
+	targetNodeCount := int(wObj.Status.TargetNodeCount)
+	readyCount := 0
+
+	for _, node := range nodeList.Items {
+		if resources.NodeIsReadyAndNotDeleting(&node) {
+			readyCount++
+		}
+	}
```
chewong: Can we replace this with `GetReadyNodes` instead?

pkg/utils/resources/nodes.go
```diff
@@ -124,53 +121,38 @@ func ExtractObjFields(obj client.Object) (instanceType, namespace, name string,
 	return
 }
 
-// GetBYOAndReadyNodes finds all BYO nodes and ready nodes that match the workspace's label selector
-func GetBYOAndReadyNodes(ctx context.Context, c client.Client, wObj *kaitov1beta1.Workspace) ([]*corev1.Node, []string, error) {
+// GetReadyNodes finds all ready nodes that match the workspace's label selector
+func GetReadyNodes(ctx context.Context, c client.Client, wObj *kaitov1beta1.Workspace) ([]*corev1.Node, error) {
 	var matchLabels client.MatchingLabels
 	if wObj.Resource.LabelSelector != nil {
 		matchLabels = wObj.Resource.LabelSelector.MatchLabels
```
chewong, line 128: I think we can make this label selector more restrictive by adding `node.kubernetes.io/instance-type: <workspace instance type>` to the existing label selector if the disable NAP feature gate is enabled. Worst case is we get some nodes that have matching labels but don't have the same instance type

pkg/workspace/estimator/advancednodesestimator/estimator.go
```diff
@@ -52,12 +54,37 @@ func (c *AdvancedNodesEstimator) EstimateNodeCount(ctx context.Context, workspac
...
+	} else {
+		// NAP is enabled - instanceType is required and must be valid
+		gpuConfig, err = utils.GetGPUConfigBySKU(workspace.Resource.InstanceType)
+		if err != nil {
+			return 0, fmt.Errorf("failed to get GPU config for instance type %s: %w", workspace.Resource.InstanceType, err)
+		}
+		if gpuConfig == nil {
+			return 0, fmt.Errorf("GPU config is nil for instance type %s", workspace.Resource.InstanceType)
+		}
 	}
```
chewong, line 77: Isn't this logic very similar to `getGPUConfig` function in `pkg/workspace/inference/preset_inferences.go`? Can we reuse it?

pkg/workspace/resource/node_claim.go
```diff
@@ -47,31 +49,45 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
 
-	// Calculate the number of NodeClaims required (target - BYO nodes)
-	readyNodes, targetNodeClaimsCount, err := nodeclaim.ResolveReadyNodesAndTargetNodeClaimCount(ctx, c.Client, wObj)
-	if err != nil {
-		return addedNodeClaimsCount, nil, nil, fmt.Errorf("failed to get required NodeClaims: %w", err)
+	// Count ready nodes that do NOT have a corresponding NodeClaim
+	readyNodesWithoutNodeClaim := 0
+
+	for _, node := range readyNodes {
+		if _, ok := node.Labels["karpenter.sh/nodepool"]; !ok {
```
chewong: nit: use `LabelNodePool` from `pkg/utils/consts/consts.go`

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -144,46 +144,69 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
...
-	// check nodeclaims meet the target count
-	if ready, err := c.nodeClaimManager.AreNodeClaimsReady(ctx, wObj, existingNodeClaims); err != nil {
+	// Check if selected nodes are ready in both NAP and BYO scenarios.
+	if ready, err := c.nodeResourceManager.EnsureNodesReady(ctx, wObj); err != nil {
 		return reconcile.Result{}, err
 	} else if !ready {
-		// Not enough ready nodeclaims, requeue and wait for next reconcile
+		// Not enough ready nodes, requeue and wait for next reconcile.
 		return reconcile.Result{}, nil
 	}
```
chewong: nit: we can remove this else-if since the return value a couple lines after is identical

pkg/workspace/resource/node.go
Jont828: I list all the nodes so we can use `len(nodeList.Items)` so we can update the condition with `fmt.Sprintf("Ready Nodes are not enough (TargetNodes: %d, CurrentReadyNodes: %d, SelectedNodes: %d)", targetNodeCount, readyCount, len(nodeList.Items))`. Do you think we should drop the # selected nodes in the condition?

chewong: Oh I see. Then it should fine to keep it

pkg/workspace/estimator/advancednodesestimator/estimator.go
Jont828: We can't import the getGPUConfig() function without causing an import cycle. If we try to move it to the utils package, that won't work either, unfortunately, since v1beta1 API imports utils so utils can't use the v1beta1 Workspace type.

charts/kaito/workspace/charts/gpu-feature-discovery-0.17.4.tgz
Jont828, line 1: Reverted version to 0.17.4 as node feature discovery didn't work with 0.18.0. I'm thinking we wait for 0.18.1 and do a follow up to bump the version.

Jont828:
Vulnerability scanner test indicates we need to update pip, which we can do in a separate PR.

pkg/workspace/estimator/advancednodesestimator/estimator.go
```diff
@@ -52,12 +54,37 @@ func (c *AdvancedNodesEstimator) EstimateNodeCount(ctx context.Context, workspac
...
+			if err != nil {
+				return 0, fmt.Errorf("failed to get GPU config from existing nodes: %w", err)
+			}
+		}
+
+		// If we can't get GPU config from nodes, fall back to minimal default
+		if gpuConfig == nil {
+			gpuConfig = &sku.GPUConfig{
+				GPUCount:  1,
+				GPUMemGiB: 0, // Unknown memory, will skip memory-based optimization
```
Fei-Guo: @bangqipropel , are we handling the case where GPUMemory is reported as 0?

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -152,37 +195,13 @@ func (c *WorkspaceReconciler) addOrUpdateWorkspace(ctx context.Context, wObj *ka
...
-
-	// check node plugins ready
-	if ready, err := c.nodeResourceManager.AreNodePluginsReady(ctx, wObj, existingNodeClaims); err != nil {
-		return reconcile.Result{}, err
-	} else if !ready {
-		// The node resource changes can not trigger workspace controller reconcile, so we need to requeue reconcile when don't proceed because of node resource not ready.
-		return reconcile.Result{RequeueAfter: 2 * time.Second}, nil
+	if result, err := c.reconcileNodes(ctx, wObj); err != nil {
+		return result, err
+	} else if result.RequeueAfter > 0 {
```
Fei-Guo: This looks like a behavior change. Line 172, return reconcile.Result{}, nil.  The check in line 200 skipped this case and does not return.

pkg/workspace/resource/node.go
```diff
@@ -127,23 +140,84 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
+	} else {
+		klog.InfoS("Ready Nodes for workspace are not enough currently", "workspace", client.ObjectKeyFromObject(wObj).String(),
+			"targetNodes", targetNodeCount, "currentReadyNodes", readyCount)
+		if updateErr := workspace.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeNodeStatus, metav1.ConditionFalse,
+			"NodeNotReady", fmt.Sprintf("Ready Nodes are not enough (TargetNodes: %d, CurrentReadyNodes: %d, SelectedNodes: %d)", targetNodeCount, readyCount, len(nodeList.Items))); updateErr != nil {
+			klog.ErrorS(updateErr, "failed to update Node status condition NodesReady to false", "workspace", klog.KObj(wObj))
+			return false, fmt.Errorf("failed to update Node status condition(NodeNotReady): %w", updateErr)
+		}
+
+		return false, nil
```
Fei-Guo, line 205: For node claim not ready case, we can just return nil so that we don't have to requeue the request because the controller watches "NodeClaim". Any changes in nodeclaim will trigger the reconciliation. 

The controller does not watch for node. If we don't requeue when nodes are not ready, the controller will not be triggered once nodes becomes ready. Do we need to add node watcher for BYO case?

pkg/utils/resources/nodes.go
```diff
@@ -124,53 +121,38 @@ func ExtractObjFields(obj client.Object) (instanceType, namespace, name string,
 	return
 }
 
-// GetBYOAndReadyNodes finds all BYO nodes and ready nodes that match the workspace's label selector
-func GetBYOAndReadyNodes(ctx context.Context, c client.Client, wObj *kaitov1beta1.Workspace) ([]*corev1.Node, []string, error) {
+// GetReadyNodes finds all ready nodes that match the workspace's label selector
+func GetReadyNodes(ctx context.Context, c client.Client, wObj *kaitov1beta1.Workspace) ([]*corev1.Node, error) {
```
Fei-Guo, line 125: This name is too generic, can we rename to "GetReadyNodesForWorkspace".

pkg/workspace/resource/node.go
```diff
@@ -127,23 +140,84 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
+		kaitov1beta1.ConditionTypeNodePluginStatus,
+	}
+
+	// If any owned condition is false, set the resource condition to false and return.
+	if err := c.SetConditionWithOwnedConditions(ctx, wObj, kaitov1beta1.ConditionTypeResourceStatus, ownedConditions); err != nil {
+		return err
+	}
+
+	// All owned conditions are true, set the resource condition to true.
 	if err := workspace.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeResourceStatus, metav1.ConditionTrue,
```
rambohe-ch: It seems that resource condition will be set to true even if all owned conditions are false.

pkg/workspace/resource/node_claim.go
```diff
@@ -47,31 +50,54 @@ func NewNodeClaimManager(c client.Client, recorder record.EventRecorder, expecta
...
+
+	for _, node := range readyNodes {
+		if _, ok := node.Labels[consts.LabelNodePool]; !ok {
+			readyNodesWithoutNodeClaim++
+		}
 	}
 
-	ncList, err := nodeclaim.ListNodeClaim(ctx, wObj, c.Client)
+	// Calculate how many NodeClaims we need, including those already provisioned.
+	targetNodeClaimCount := max(0, targetNodeCount-readyNodesWithoutNodeClaim)
```
rambohe-ch, line 66: I want to check two points for calculating targetNodeClaimCount:
1. We account on the number of ready BYO nodes to calculate targetNodeClaimCount, this means if one BYO node falls into not ready, a new NodeClaim will be created to replace it. and after some time, If the BYO node comes back to ready, the created NodeClaim will be leaked because it seems that we have not considered NodeClaim deletion logic.
2. If our design is: one BYO node falls into not ready, a new NodeClaim will be created to replace it. There is a problem is `workspace controller` has not watched nodes, so workspace reconcile will not be triggered when BYO node becomes not ready.

pkg/workspace/resource/node_claim.go
```diff
@@ -116,14 +142,16 @@ func (c *NodeClaimManager) CreateUpNodeClaims(ctx context.Context, wObj *kaitov1
...
 }
 
-// AreNodeClaimsReady is used for checking the number of ready nodeclaims(isNodeClaimReadyNotDeleting) meet the target NodeClaim count needed
-func (c *NodeClaimManager) AreNodeClaimsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
-	// Calculate the actual number of NodeClaims needed (if BYO, numtargetNodes == 0)
-	_, targetNodeClaimCount, err := nodeclaim.ResolveReadyNodesAndTargetNodeClaimCount(ctx, c.Client, wObj)
+// EnsureNodeClaimsReady is used for checking the number of ready nodeclaims(isNodeClaimReadyNotDeleting) meet the target NodeClaim count needed. Updates the
+func (c *NodeClaimManager) EnsureNodeClaimsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
+	// Since NodeClaims are only used when NAP is enabled, we know that all target nodes will be provisioned by NodeClaims instead of BYO nodes.
+	readyNodes, err := resources.GetReadyNodes(ctx, c.Client, wObj)
```
rambohe-ch: A optimize point: we have listed all nodes one time in `CheckNodeClaims` function, and we should use the same node resources in one reconcile routine. so how about pass the readyNodes as parameter for  `EnsureNodeClaimsReady ` instead of listing nodes one more time?

pkg/workspace/resource/node.go
```diff
@@ -41,27 +42,34 @@ func NewNodeManager(c client.Client) *NodeManager {
...
+			if updateErr := workspace.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeNodePluginStatus, metav1.ConditionFalse,
+				"NodePluginsNotReady", "waiting all node plugins to be ready"); updateErr != nil {
 				klog.ErrorS(updateErr, "failed to update workspace status", "workspace", klog.KObj(wObj))
 				return false, updateErr
 			}
 			return false, nil
 		}
 	}
+
+	if updateErr := workspace.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeNodePluginStatus, metav1.ConditionTrue,
```
rambohe-ch: If there is not a Known GPU config for the workspace, There is no need to check node plugins for the workspace. of course, there is no need to add `ConditionTypeNodePluginStatus` condition for the workspace. So I don't think we need to add a indepedent statuses like `ConditionTypeNodePluginStatus` and `ConditionTypeNodeStatus` for Workspace. and these statuses are the subset of `ConditionTypeResourceStatus` status.

rambohe-ch: In future version, If we need to some other verfications for the resource of Workspace, It will make us to add more statuses for workspace. so the status definition of workspace will become more and more complex.

pkg/workspace/resource/node.go
```diff
@@ -127,23 +140,84 @@ func (c *NodeManager) getReadyNodesFromNodeClaims(ctx context.Context, wObj *kai
...
-			status.WorkerNodes = workerNodes
-			return nil
-		}); err != nil {
-			if updateErr := workspace.UpdateStatusConditionIfNotMatch(ctx, c.Client, wObj, kaitov1beta1.ConditionTypeResourceStatus, metav1.ConditionFalse,
-				"workspaceResourceStatusFailed", err.Error()); updateErr != nil {
-				klog.ErrorS(updateErr, "failed to update workspace status", "workspace", klog.KObj(wObj))
-				return updateErr
+// EnsureNodesReady is used for checking the number of ready nodes meet the target node count. Updates the status condition accordingly.
+func (c *NodeManager) EnsureNodesReady(ctx context.Context, wObj *kaitov1beta1.Workspace) (ready bool, retErr error) {
+	nodeList, err := resources.ListNodes(ctx, c.Client, wObj.Resource.LabelSelector.MatchLabels)
```
rambohe-ch: A optimize point: we have listed all nodes one time in CheckNodeClaims function, and we should use the same node resources in one reconcile routine. so how about pass the readyNodes as parameter for EnsureNodesReady  instead of listing nodes one more time? 

Because CheckNodeClaims will not executed when NAP is disabled, so how about add a list nodes at the the begining for all of them.

pkg/workspace/resource/node.go
```diff
@@ -41,27 +42,34 @@ func NewNodeManager(c client.Client) *NodeManager {
 	}
 }
 
-// AreNodePluginsReady is used for ensuring node label(accelerator:nvidia) and GPU capacity on all auto-provisioned nodes for the workspace.
-func (c *NodeManager) AreNodePluginsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
+// EnsureNodePluginsReady is used for ensuring node label(accelerator:nvidia) and GPU capacity on all auto-provisioned nodes for the workspace.
+func (c *NodeManager) EnsureNodePluginsReady(ctx context.Context, wObj *kaitov1beta1.Workspace, existingNodeClaims []*karpenterv1.NodeClaim) (bool, error) {
```
rambohe-ch: `EnsureNodePluginsReady` name means we will ensure node plugin ready in ths function like waiting node plugins ready. But we only check the node plugins are ready or not, and not ensure. so I think the original function name: AreNodePluginsReady is more suitable.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -187,27 +188,55 @@ func GeneratePresetInference(ctx context.Context, workspaceObj *v1beta1.Workspac
...
-	//nolint:staticcheck // SA1019
-	if len(ctx.Workspace.Resource.PreferredNodes) == 0 {
-		gpuConfig, _ = utils.GetGPUConfigBySKU(ctx.Workspace.Resource.InstanceType)
-		if gpuConfig != nil {
-			return *gpuConfig
+
+	if featuregates.FeatureGates[consts.FeatureFlagDisableNodeAutoProvisioning] {
+		// NAP is disabled (BYO scenario) - prefer to get GPU config from matching nodes with nvidia.com labels
+		// Only try to find matching nodes if we have a labelSelector and if WorkerNodes is not already populated
+		if ctx.Workspace.Resource.LabelSelector != nil && len(ctx.Workspace.Status.WorkerNodes) == 0 {
```
rambohe-ch: why should add condition `len(ctx.Workspace.Status.WorkerNodes) == 0`?

pkg/workspace/estimator/advancednodesestimator/estimator.go
```diff
@@ -52,12 +54,37 @@ func (c *AdvancedNodesEstimator) EstimateNodeCount(ctx context.Context, workspac
 	presetName := string(workspace.Inference.Preset.Name)
 	model := plugin.KaitoModelRegister.MustGet(presetName)
 
-	gpuConfig, err := utils.GetGPUConfigBySKU(workspace.Resource.InstanceType)
```
rambohe-ch, line 55: It is recommended to add another util function for getting GPU Config, so `advancednodeestimator/basicnodeestimator/preset_tunning` can use it.

pkg/workspace/estimator/advancednodesestimator/estimator.go
bangqipropel: @Fei-Guo No, we don't handle this.

bangqipropel (commented):
In fact, about the node estimator, I would say we shall not re-use the EstimateNodeCount in this case. Our EstimateNodeCount is based on the node that we currently run, but we don't know what the cx' nodes are.  So, the better idea is that you write a new interface function in the estimator and put the BYO codes there.

Also, I think for most of the BYO node features, we should try to decouple the functions with our current node ones like the EstimateNodeCount

pkg/workspace/estimator/advancednodesestimator/estimator.go
Jont828: We still use the util function later on in the NAP case.`gpuConfig, err = utils.GetGPUConfigBySKU(workspace.Resource.InstanceType)`. As for the BYO part, unfortunately, we can't add easily add a util function for this logic here as  it would cause an import cycle (it uses the v1beta1 Workspace type, and v1beta1 imports utils). 

```		// Try to get GPU config from existing nodes
		if readyNodes, err := resources.GetReadyNodes(ctx, client, workspace); err != nil {
			return 0, fmt.Errorf("failed to list ready nodes: %w", err)
		} else if len(readyNodes) > 0 {
			gpuConfig, err = utils.TryGetGPUConfigFromNodes(ctx, readyNodes)
			if err != nil {
				return 0, fmt.Errorf("failed to get GPU config from existing nodes: %w", err)
			}
		}
```

pkg/workspace/controllers/workspace_controller.go
Jont828: Good catch!

pkg/workspace/inference/preset_inferences.go
Jont828: Yea we can drop that

pkg/workspace/resource/node.go
Jont828: I think we could consider it as an optimization/follow-up. It could be a bit expensive to watch all the nodes on a cluster or to figure out which ones to watch as you can add or remove the label from the Workspace.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -144,45 +144,80 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
 	return nil
 }
 
-func (c *WorkspaceReconciler) addOrUpdateWorkspace(ctx context.Context, wObj *kaitov1beta1.Workspace) (reconcile.Result, error) {
-	workspaceKey := client.ObjectKeyFromObject(wObj).String()
-	if !c.expectations.SatisfiedExpectations(c.Log, workspaceKey) {
-		klog.V(4).InfoS("Waiting for NodeClaim expectations to be satisfied",
-			"workspace", workspaceKey)
-		return reconcile.Result{}, nil
+func (c *WorkspaceReconciler) reconcileNodes(ctx context.Context, wObj *kaitov1beta1.Workspace) (*reconcile.Result, error) {
```
chewong, line 147: nit: it's not common to use pointers for `reconcile.Result` type

```suggestion
func (c *WorkspaceReconciler) reconcileNodes(ctx context.Context, wObj *kaitov1beta1.Workspace) (reconcile.Result, error) {
```

Fei-Guo (approved):

Issue #1085, opened 2025-04-30
Support multimodal models


zhuangqh:
**Is your feature request related to a problem? Please describe.**

Since kaito is open to support live download model from hf, it's time to support multimodal model without license issue.

**Describe the solution you'd like**

Onboarding image models. ref https://huggingface.co/spaces/opencompass/open_vlm_leaderboard

**Describe alternatives you've considered**

How can users upload/download image/audio inputs easily? Do we need to handle a file server?

**Additional context**

rambohe-ch:
Another idea comes into my eye, how about support models on [ModelScope](https://modelscope.cn/models)? It is hard for end user from China mainland to download models from huggingface, so ModelScope is option for users to download models.

chewong:
> Another idea comes into my eye, how about support models on [ModelScope](https://modelscope.cn/models)?

@rambohe-ch looks like there is a way to do it in vLLM:

<img width="884" height="208" alt="Image" src="https://github.com/user-attachments/assets/2d36a453-e190-43d3-9af1-05b0f9ff9aa5" />

zhuangqh:
Users use the same chat completion API to talk to multimodal model. See example: https://docs.vllm.ai/en/latest/examples/online_serving/openai_chat_completion_client_for_multimodal.html

abhisheksheth28:
#1572 adds support for multimodal Gemma 3 models.

PR #1540, opened 2025-10-01
docs: add proposal for gemma 3 models

abhisheksheth28:
**Reason for Change**:

proposal for gemma 3 models

Gemma 3 models (4B, 12B and 27B) are multimodal - Image-Text-to-Text to be specific. (Image and Text input -> Text output)

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/docs/proposals/20251001-gemma-3-27b-instruct.md
+++ b/docs/proposals/20251001-gemma-3-27b-instruct.md
@@ -34,15 +34,13 @@ metadata:
   name: workspace-gemma-3-27b-instruct
 spec:
   resource:
-    instanceType: Standard_ND96isr_H100_v5
+    instanceType: Standard_NC24ads_A100_v4
     labelSelector:
       matchLabels:
         apps: gemma-3-27b-instruct
   inference:
     preset:
       name: gemma-3-27b-instruct
-    accessModes:
-      - modelAccessSecret
   modelAccessSecret: hf-token-secret
 ```
 
--- a/docs/proposals/20251001-gemma-3-27b-instruct.md
+++ b/docs/proposals/20251001-gemma-3-27b-instruct.md
@@ -1,0 +1,78 @@
+---
+title: Proposal for new model support
+authors:
+  - "Abhishek Sheth"
+reviewers:
+  - "KAITO contributor"
+creation-date: 2025-10-01
+last-updated: 2025-10-01
+status: provisional
+---
+
+# Title
+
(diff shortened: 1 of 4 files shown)


abhisheksheth28:
> Could you also add the two models in https://github.com/kaito-project/kaito/blob/main/presets/workspace/models/supported_models.yaml so we can trigger preset image build once this PR is merged?

updated https://github.com/kaito-project/kaito/blob/main/presets/workspace/models/supported_models.yaml

chewong:
@abhisheksheth28 could you update https://github.com/kaito-project/kaito/blob/main/charts/kaito/workspace/templates/supported-models-configmap.yaml as well?

docs/proposals/20251001-gemma-3-27b-instruct.md
```diff
@@ -0,0 +1,78 @@
...
+Sample YAML:
+
+```yaml
+apiVersion: kaito.sh/v1beta1
+kind: Workspace
+metadata:
+  name: workspace-gemma-3-27b-instruct
+spec:
+  resource:
+    instanceType: Standard_ND96isr_H100_v5
```
chewong: We don't need this powerful instance type for a 27b model. Let's use `Standard_NC24ads_A100_v4` instead

```suggestion
    instanceType: Standard_NC24ads_A100_v4
```

docs/proposals/20251001-gemma-3-27b-instruct.md
```diff
@@ -0,0 +1,78 @@
...
+    labelSelector:
+      matchLabels:
+        apps: gemma-3-27b-instruct
+  inference:
+    preset:
+      name: gemma-3-27b-instruct
+    accessModes:
+      - modelAccessSecret
```
chewong: This field is deprecated. Let's remove it

docs/proposals/20251001-gemma-3-27b-instruct.md
abhisheksheth28: > We don't need this powerful instance type for a 27b model. Let's use `Standard_NC24ads_A100_v4` instead

okay

Fei-Guo (approved):
Approved. 

Ideally, we should study how vllm turns an image into tokens, does it use a fixed token budget for any image at the cost of potentially downgrading the image resolution?

abhisheksheth28:
> Approved.
> 
> Ideally, we should study how vllm turns an image into tokens, does it use a fixed token budget for any image at the cost of potentially downgrading the image resolution?

Agreed. I’ll be looking into this as part of the model implementation and #1085.

PR #1558, opened 2025-10-08
release: update manifest and helm charts for v0.7.1

zhuangqh:


charts/kaito/workspace/Chart.yaml
```diff
@@ -34,6 +34,6 @@ annotations:
   artifacthub.io/prerelease: "false"
 dependencies:
 - name: flux2
-  version: "2.16.3"
```
chewong, line 37: This shouldn't be updated to 0.7.1

zhuangqh: Updated. fix the make command to only update the first "version".

chewong (approved):

PR #1605, opened 2025-10-20
feat: Provide token usage in rag service

bangqipropel:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/ragengine/vector_store/base.py
+++ b/presets/ragengine/vector_store/base.py
@@ -483,6 +483,35 @@ class BaseVectorStore(ABC):
                 )
                 return await self.llm.chat_completions_passthrough(openai_request)
 
+            # Try to get usage from LLM response via the last_usage attribute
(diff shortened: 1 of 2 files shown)


presets/ragengine/vector_store/base.py
```diff
@@ -498,6 +498,16 @@ async def chat_completion(self, request: dict) -> ChatCompletionResponse:
                         "index": 0,
                     }
                 ],
+                usage={
```
bfoley13: I'm okay with this for now but realistically we should get it from the response to the LLM if we can: https://github.com/kaito-project/kaito/blob/51182cdcaa68cb33ece0260c0e9b06d4df75d1dd/presets/ragengine/inference/inference.py#L243

bfoley13 (approved):

Fei-Guo (approved):

PR #1611, opened 2025-10-21
chore: bump local-csi-driver to v0.2.6

jmclong:
**Reason for Change**:
 
 Updates local-csi-driver to v0.2.6.

**Requirements**

 * [ ]  added unit tests and e2e tests (if applicable).

**Issue Fixed**:

fix: count unassigned and assigned pv to vg for total capacity by @jmclong in https://github.com/Azure/local-csi-driver/pull/173
fix: remove corrupted logical volume when lvcreate killed by @jmclong in https://github.com/Azure/local-csi-driver/pull/193
Set csi-resizer --handle-volume-inuse-error=false to reduce memory footprint in large clusters by @Copilot in https://github.com/Azure/local-csi-driver/pull/210

Updated images and dependencies

**Notes for Reviewers**:

Release notes: https://github.com/Azure/local-csi-driver/releases/tag/v0.2.6

### DIFF
--- a/charts/kaito/workspace/templates/local-csi-driver-ds.yaml
+++ b/charts/kaito/workspace/templates/local-csi-driver-ds.yaml
@@ -209,7 +209,7 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
-        image: "mcr.microsoft.com/acstor/local-csi-driver:v0.2.5"
+        image: "mcr.microsoft.com/acstor/local-csi-driver:v0.2.6"
         imagePullPolicy: IfNotPresent
         livenessProbe:
           httpGet:
@@ -318,7 +318,7 @@ spec:
         args:
         - --csi-address=/csi/csi.sock
         - --timeout=240s
-        - --handle-volume-inuse-error=true
+        - --handle-volume-inuse-error=false
         - --http-endpoint=:8091
         - --v=2
         env:
--- a/hack/local-csi-driver-charts-sync/local-csi-driver-charts-sync.sh
+++ b/hack/local-csi-driver-charts-sync/local-csi-driver-charts-sync.sh
@@ -12,7 +12,7 @@
 SCRIPT_DIR="$(realpath "$(dirname "$0")")"
 
 REPO=https://github.com/Azure/local-csi-driver.git
-REVISION="06932013f5ab544172636a3d907a551914f64b68"
+REVISION="55a538ca60ed639ffd329d554506eae8866aea36"
 
 TEMP_DIR=$(mktemp -d)
 REPO_DIR="$TEMP_DIR/local-csi-driver"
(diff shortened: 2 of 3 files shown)


jmclong:
> ## PR Code Suggestions ✨
> Explore these optional code suggestions:
> 
> **Category**	**Suggestion                                                                                                                                    **	**Impact**
> Possible issue	
> Revert critical volume error handling flag
> **The flag `--handle-volume-inuse-error` was changed from `true` to `false`. This disables critical volume-in-use error handling during storage operations, which could lead to data corruption or volume expansion failures. Verify this change is intentional and required by the new driver version. If not, revert to `true` to maintain safe volume operations.**
> 
> [charts/kaito/workspace/templates/local-csi-driver-ds.yaml [321]](https://github.com/kaito-project/kaito/pull/1611/files#diff-41fdd564064e869517042d47a8ccdd21dff8f4cd04cfe3f98c826e43f2e3af55R321-R321)
> 
> ```diff
> -- --handle-volume-inuse-error=false
> +- --handle-volume-inuse-error=true
> ```
> 
> Suggestion importance[1-10]: 7
> __
> 
> Why: The suggestion correctly identifies a critical flag change but only requests verification rather than confirming an error. The driver version update might intentionally require this change.
> 
> Medium

https://github.com/kubernetes-csi/external-resizer
--handle-volume-inuse-error <true/false>: Enable or disable volume-in-use error handling in external-resizer. Defaults to true and resize-controller will watch for all pods in all namespaces to check if PVC being expanded is in-use by a pod or not before retrying volume expansion if CSI driver throws volume-in-use error. Setting this to false will cause external-resizer to ignore volume-in-use error and resize-controller will retry volume expansion even if volume is already in use by a pod and CSI driver does not support expansion of in-use volumes. If CSI driver being used supports online expansion, it might be desirable to set handle-volume-inuse-error to false - to save costs associated with watching all pods in the cluster.


Interesting catch. Technically this appears to be safe fix, given both ext4 and xfs support online resize. For raw block volume as long as application can handle online expansion (which the application should be if it is asking for expansion) we are safe.

What sizes of clusters did we test with when we noticed that the csi-resizer was getting oom killed?

@anson627 was finding this when testing some scaling on 543 nodes and 10992 pods, so not too large. I checked the rest of our code in local-csi-driver and we never do something like a List on all k8s resources (but csi-resizer does unless we set this to false)

https://github.com/kubernetes-csi/external-resizer
--handle-volume-inuse-error <true/false>: Enable or disable volume-in-use error handling in external-resizer. Defaults to true and resize-controller will watch for all pods in all namespaces to check if PVC being expanded is in-use by a pod or not before retrying volume expansion if CSI driver throws volume-in-use error. Setting this to false will cause external-resizer to ignore volume-in-use error and resize-controller will retry volume expansion even if volume is already in use by a pod and CSI driver does not support expansion of in-use volumes. If CSI driver being used supports online expansion, it might be desirable to set handle-volume-inuse-error to false - to save costs associated with watching all pods in the cluster.

we support online volume expansion, so there is no point in setting this to true, as far as I can tell from the description

jmclong:
PR: https://github.com/kaito-project/kaito/pull/1483

PR #1621, opened 2025-10-27
feat: [BREAKING] remove /query api call and adding FastAPI info and Tags

bfoley13:
**Reason for Change**:
This change removes the `/query` api from the rag engine service. All `/query` calls by consumers should be changed to use the `/v1/chat/completions` api. More information about the `/v1/chat/completions` api can be found [here](https://kaito-project.github.io/kaito/docs/rag-api#openai-compatible-chat-completions)

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


presets/ragengine/main.py
```diff
@@ -784,4 +797,4 @@ async def shutdown_event():
     # llama_index.core.set_global_handler("arize_phoenix")
     import uvicorn
 
-    uvicorn.run(app, host="0.0.0.0", port=5000, loop="asyncio")
+    uvicorn.run(app, host="0.0.0.0", port=5789, loop="asyncio")
```
zhuangqh: why do you update the port here? do k8s service need to be updated as well?

bfoley13: No this was a local testing thing, will revert!

PR #1637, opened 2025-11-02
chore: use built-in GenerateName to generate random workspace name by InferenceSet

andyzhangx:
**Reason for Change**:

chore: use built-in GenerateName to generate random workspace name by InferenceSet
this PR addresses the former comments which makes the workspace name generated by InferenceSet more clear:

 - the generated name is like following:
```
# k get workspace
NAME                 INSTANCE                   RESOURCEREADY   INFERENCEREADY   JOBSTARTED   WORKSPACESUCCEEDED   AGE
llama-3-1-8b-2brcn   Standard_NC24ads_A100_v4   False                                                              109s
llama-3-1-8b-mmc2p   Standard_NC24ads_A100_v4   False                                                              2s
```

```
	// GenerateName is an optional prefix, used by the server, to generate a unique
	// name ONLY IF the Name field has not been provided.
	// If this field is used, the name returned to the client will be different
	// than the name passed. This value will also be combined with a unique suffix.
	// The provided value has the same validation rules as the Name field,
	// and may be truncated by the length of the suffix required to make the value
	// unique on the server.
	//
	// If this field is specified and the generated name exists, the server will return a 409.
	//
	// Applied only if Name is not specified.
	// More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency
	// +optional
	GenerateName string `json:"generateName,omitempty" protobuf:"bytes,2,opt,name=generateName"`
```

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/workspace/controllers/inferenceset/inferenceset_controller.go
```diff
@@ -236,15 +236,10 @@ func (c *InferenceSetReconciler) addOrUpdateInferenceSet(ctx context.Context, iO
 	if replicaNumToCreate > 0 {
 		klog.InfoS("Need to create more workspaces...", "current", len(wsList.Items), "desired", iObj.Spec.Replicas)
 		for i := range replicaNumToCreate {
-			workspaceName := workspace.GetWorkspaceNameWithRandomSuffix(iObj.Name)
-			workspaceObj := inferenceset.GetWorkspace(workspaceName, wsList)
```
zhuangqh, line 240: Please remove the definition and testing code for `inferenceset.GetWorkspace` as well.

andyzhangx: removed

PR #1660, opened 2025-12-03
feat: add PV support to RAG service

bangqipropel:
**Reason for Change**:
1. Add PV support to RAG service
2. Add hook to auto trigger the pv support

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/website/docs/gateway-api-inference-extension.md
+++ b/website/docs/gateway-api-inference-extension.md
@@ -33,7 +33,8 @@ helm upgrade --install kaito-workspace kaito/workspace \
   --namespace kaito-workspace \
   --create-namespace \
   --set clusterName="$CLUSTER_NAME" \
-  --set featureGates.gatewayAPIInferenceExtension=true,enableInferenceSetController=true \
+  --set featureGates.gatewayAPIInferenceExtension=true \
+  --set featureGates.enableInferenceSetController=true \
   --wait
 ```
 


docker/ragengine/service/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN apt-get update && apt-get upgrade -y \
     g++ \
     make \
     perl \
+    curl \
```
andyzhangx: irrelevant change here?

presets/ragengine/lifecycle/manager.py
```diff
@@ -0,0 +1,372 @@
...
+import json
+import os
+import shutil
+import subprocess
+import time
+from contextlib import suppress
+from datetime import datetime
+from pathlib import Path
+
+try:
```
Fei-Guo: why not install this package?

docker/ragengine/service/Dockerfile
bangqipropel: we need curl in the new hook

presets/ragengine/lifecycle/manager.py
bangqipropel: done

bangqipropel:
> I think you also need to provide an example config or doc about how to use this pvc, and on AKS, azure disk volume pvc (ReadWriteOnce) would be ok, right?

Doc added

Fei-Guo (approved):

PR #1667, opened 2025-12-05
chore: [BREAKING] deprecate phi-2 model

zhuangqh:
**Reason for Change**:

- remove related preset conf and tests
- give an error in the webhook

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


docker/ragengine/Dockerfile
```diff
@@ -24,6 +24,7 @@ ARG BUILD_DATE
 COPY cmd/ cmd/
 COPY api/ api/
 COPY pkg/ pkg/
+COPY presets/ presets/
```
andyzhangx, line 27: irrelevant change here?

zhuangqh: It's required. This pkg presets/workspace/models/metadata.go
required by rag at this change.

Fei-Guo (approved):

PR #1679, opened 2025-12-08
docs: add keda-autoscaler-inference doc

andyzhangx:
**Reason for Change**:

doc: add keda-autoscaler-inference doc, this doc is for v0.8.0

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/website/docs/keda-autoscaler-inference.md
+++ b/website/docs/keda-autoscaler-inference.md
@@ -50,13 +50,11 @@ helm upgrade --install kaito-workspace kaito/workspace \
 ### Create a Kaito InferenceSet for running inference workloads
  - The following example creates an InferenceSet for the phi-4-mini model, using annotations with the prefix `scaledobject.kaito.sh/` to supply parameter inputs for the KEDA Kaito Scaler:
    - `scaledobject.kaito.sh/auto-provision`
-     - indicates whether KEDA Kaito Scaler will automatically provision a ScaledObject based on the `InferenceSet` object
-   - `scaledobject.kaito.sh/max-replicas`
-     - specifies the maximum number of replicas for the `InferenceSet` during autoscaling.
+     - required, specifies whether KEDA Kaito Scaler will automatically provision a ScaledObject based on the `InferenceSet` object
    - `scaledobject.kaito.sh/metricName`
-     - specifies the metric name collected from the vLLM pod, which is used for monitoring and triggering the scaling operation, default is `vllm:num_requests_waiting`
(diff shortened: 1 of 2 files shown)


website/docs/keda-autoscaler-inference.md
```diff
@@ -0,0 +1,99 @@
...
+  --create-namespace \
+  --set clusterName="$CLUSTER_NAME" \
+  --set featureGates.enableInferenceSetController=true \
+  --wait
+```
+
+## Quickstart
+
+### Create a Kaito InferenceSet for running inference workloads
+ - The following example creates an InferenceSet for the phi-4-mini model, using annotations with the prefix `scaledobject.kaito.sh/` to supply parameter inputs for the KEDA Kaito Scaler:
```
Fei-Guo, line 51: We need to make is clear that are those annotations required or optional.

and we should not let user to specify the max-replica, right? The number should be derived from inferenceset.spec.nodecountlimit and model's estimated GPU count (based on sku and model mem requirement).

andyzhangx: you are right, keda-kaito-scaler will calculate the max-replica automatically, so this `scaledobject.kaito.sh/max-replicas` annotation is not needed as long as Spec.NodeCountLimit is not empty

```
	if is.Spec.NodeCountLimit != 0 {
		workspaceList, err := inferenceset.ListWorkspaces(ctx, is, c.Client)
		if err != nil {
			return reconcile.Result{}, fmt.Errorf("failed to list workspaces for inference set %s/%s: %v", is.Namespace, is.Name, err)
		}

		var targetNodeCount int32
		for i := range workspaceList.Items {
			if workspaceList.Items[i].Status.TargetNodeCount != 0 {
				targetNodeCount = workspaceList.Items[i].Status.TargetNodeCount
				break
			}
		}

		if targetNodeCount == 0 {
			logger.Info("target node count from workspaces for inference set is zero, and requeue after 5 seconds", "namespace", is.Namespace, "name", is.Name)
			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
		}

		maxReplicas = is.Spec.NodeCountLimit / int(targetNodeCount)
		if maxReplicas < 1 {
			maxReplicas = 1
		}
	} else if maxReplicasStr, ok := is.Annotations[AnnotationKeyMaxReplicas]; ok {
		maxReplicas, _ = strconv.Atoi(maxReplicasStr)
	} else {
		maxReplicas = 1
	}
```

Fei-Guo (approved):

PR #1692, opened 2025-12-16
test: use helm v3 for Upgrade Compatibility Test

zhuangqh:
**Reason for Change**:

- use kind instead.
- pin helm cli version to v3. incompatible with helmv4 now.


zhuangqh:
upgrade from 0.7.1 https://github.com/kaito-project/kaito/actions/runs/20255835230
upgrade from 0.7.2 https://github.com/kaito-project/kaito/actions/runs/20255900304/job/58157900897

PR #1701, opened 2025-12-20
chore: upgrade ragengine crd to v1beta1

bangqipropel:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/charts/kaito/workspace/templates/_helpers.tpl
+++ b/charts/kaito/workspace/templates/_helpers.tpl
@@ -34,6 +34,9 @@ Create chart name and version as used by the chart label.
 Common labels
 */}}
 {{- define "kaito.labels" -}}
+{{- if .Values.commonLabels }}
+{{- toYaml .Values.commonLabels }}
+{{- else }}
 helm.sh/chart: {{ include "kaito.chart" . }}
 app.kubernetes.io/created-by: {{ include "kaito.chart" . }}
 {{ include "kaito.selectorLabels" . }}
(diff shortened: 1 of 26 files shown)


api/v1beta1/ragengine_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// +optional
+	Storage *StorageSpec `json:"storage,omitempty"`
+	// Embedding specifies whether the RAG engine generates embedding vectors using a remote service
+	// or using a embedding model running locally.
+	Embedding        *EmbeddingSpec        `json:"embedding"`
+	InferenceService *InferenceServiceSpec `json:"inferenceService"`
+	// QueryServiceName is the name of the service which exposes the endpoint for accepting user queries to the
+	// inference service. If not specified, a default service name will be created by the RAG engine.
+	// +optional
+	QueryServiceName string `json:"queryServiceName,omitempty"`
```
Fei-Guo: Should we remove the QueryServiceName and IndexServiceName fields? I don't think they are actively used and we can just support default service for now. We may add them later if users request.

api/v1beta1/ragengine_types.go
```diff
@@ -0,0 +1,146 @@
...
+	//
+	// Must match the token limit of the LLM backend being used (e.g., 8096, 16384, 32768 tokens).
+	ContextWindowSize int `json:"contextWindowSize"`
+}
+
+type RAGEngineSpec struct {
+	// Compute specifies the dedicated GPU resource used by an embedding model running locally if required.
+	// +optional
+	Compute *ResourceSpec `json:"compute,omitempty"`
+	// Storage specifies how to access the vector database used to save the embedding vectors.
```
Fei-Guo, line 95: Maybe we should change the comments for storage field. It is now used to persist vector DB data for in memory DB. 

Later, we may need to add a vector DB spec to support different types of vector dbs.

cmd/ragengine/main.go
```diff
@@ -68,7 +68,7 @@ var (
 
 func init() {
 	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
-	utilruntime.Must(kaitov1alpha1.AddToScheme(scheme))
+	utilruntime.Must(kaitov1beta1.AddToScheme(scheme))
```
bfoley13, line 72: I believe we should still register both here for now

charts/kaito/ragengine/crds/kaito.sh_ragengines.yaml
```diff
@@ -31,7 +31,7 @@ spec:
     - jsonPath: .metadata.creationTimestamp
       name: Age
       type: date
-    name: v1alpha1
+    name: v1beta1
```
bfoley13: I believe this should also contain both version when generated for now too. not sure why its overwriting vs having both

api/v1beta1/ragengine_types.go
bfoley13: I also thing we place the PVC claims within a sub struct of the storage spec.  Rag -> StorageSpec -> PersistentVolumeConfig -> { PersistentVolumeClaim, MountPath}. That way we keep it cleaner and seperate items if we add more storage options later

pkg/ragengine/controllers/preset_rag.go
```diff
@@ -159,31 +160,36 @@ func CreatePresetRAG(ctx context.Context, ragEngineObj *v1alpha1.RAGEngine, revi
...
-		skuNumGPUs := gpuConfig.GPUCount
-
-		resourceReq = corev1.ResourceRequirements{
-			Requests: corev1.ResourceList{
-				corev1.ResourceName(resources.CapacityNvidiaGPU): *resource.NewQuantity(int64(skuNumGPUs), resource.DecimalSI),
-			},
-			Limits: corev1.ResourceList{
-				corev1.ResourceName(resources.CapacityNvidiaGPU): *resource.NewQuantity(int64(skuNumGPUs), resource.DecimalSI),
-			},
+		instanceType := ragEngineObj.Spec.Compute.InstanceType
```
Fei-Guo, line 161: After a 2nd thought, I think we should avoid vendor specific sku in the code. So the alternative is that if GetGPUConfigBySKU returns an err, we skip setting GPU requests, etc.

Fei-Guo (approved):

PR #1713, opened 2025-12-29
feat: simplify vLLM inference model support flow

### LINKED ISSUE #1411: Add DeepSeek v3.1 as KAITO presets

https://huggingface.co/deepseek-ai/DeepSeek-V3.1-Base

**Additional context**

andyzhangx:
**Reason for Change**:

This PR simplifies vLLM inference model support flow a lot, with this PR, you only need to append following yaml config into `supported_models_best_effort.yaml`, take `Qwen/Qwen3-Coder-30B-A3B-Instruct` model as an example:

```yaml
  - name: qwen3-coder-30b-a3b-instruct
    version: https://huggingface.co/Qwen/Qwen3-Coder-30B-A3B-Instruct
    modelFileSize: 57Gi
    diskStorageRequirement: 107Gi
    bytesPerToken: 98304
    modelTokenLimit: 262144
    toolCallParser: hermes
    reasoningParser: qwen3
```

Note: I need to set `gpu_memory_utilization: 0.7` in e2e test to make qwen2-7b test pass on A10 node due to following error:

```
(APIServer pid=7) (EngineCore_DP0 pid=270) ERROR 01-06 10:59:32 [core.py:843] RuntimeError: Worker failed with error 'CUDA out of memory occurred when warming up sampler with 256 dummy requests. Please try lowering `max_num_seqs` or `gpu_memory_utilization` when initializing the engine.', please check the stack trace above for the root cause
```

**Requirements**

**Issue Fixed**:

Fixes #1411

**Notes for Reviewers**:

### DIFF
--- a/presets/workspace/models/vllm_model.go
+++ b/presets/workspace/models/vllm_model.go
@@ -70,8 +70,8 @@ func (m *vLLMCompatibleModel) GetInferenceParameters() *model.PresetParam {
 	runParamsVLLM := map[string]string{
 		"trust-remote-code": "",
 	}
-	if m.model.ModelType != "" {
-		runParamsVLLM["dtype"] = m.model.ModelType
+	if m.model.DType != "" {
+		runParamsVLLM["dtype"] = m.model.DType
 	} else {
 		runParamsVLLM["dtype"] = "bfloat16"
 	}
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -102,39 +102,44 @@ type Metadata struct {
 	// +optional
 	Deprecated bool `yaml:"deprecated,omitempty"`
 
+	// DType specifies the data type used by the model (e.g., "bfloat16", "float16", "float32").
+	// This field is part of the shared Metadata struct and is primarily populated in YAML for best effort supported models.
+	// +optional
+	DType string `yaml:"dtype,omitempty"`
+
 	// ModelFileSize is the size of the model file, example: 14Gi.
-	// only used for best_effort supported models
+	// This field is part of the shared Metadata struct and is primarily populated in YAML for best effort supported models.
 	ModelFileSize string `yaml:"modelFileSize,omitempty"`
 
 	// DiskStorageRequirement is the disk storage requirement for the model, example: 90Gi.
-	// only used for best_effort supported models
+	// This field is part of the shared Metadata struct and is primarily populated in YAML for best effort supported models.
 	DiskStorageRequirement string `yaml:"diskStorageRequirement,omitempty"`
 
 	// BytesPerToken is the number of bytes used to represent each token in the model.
-	// only used for best_effort supported models
+	// This field is part of the shared Metadata struct and is primarily populated in YAML for best effort supported models.
 	BytesPerToken int `yaml:"bytesPerToken,omitempty"`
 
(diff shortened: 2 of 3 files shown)


presets/workspace/models/vllm_model.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (m *vllmModel) GetInferenceParameters() *model.PresetParam {
+	metaData := &model.Metadata{
+		Name:                 m.model.Name,
+		ModelType:            "text-generation",
+		Version:              m.model.Version,
+		Runtime:              "tfs",
+		DownloadAtRuntime:    true,
+		DownloadAuthRequired: true,
```
zhuangqh: most models are public

presets/workspace/models/vllm_model.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	return presetParam
+}
+
+func (*vllmModel) GetTuningParameters() *model.PresetParam {
+	return nil
+}
+
+func (*vllmModel) SupportDistributedInference() bool {
```
zhuangqh: we need to consider this param in the preset generator. We can just set default to true at this now. The estimator will choose least nodes to run.

presets/workspace/models/vllm_model.go
```diff
@@ -0,0 +1,122 @@
...
+		utilruntime.Must(m.Validate())
+		plugin.KaitoModelRegister.Register(&plugin.Registration{
+			Name:     m.Name,
+			Instance: &vllmModel{model: m},
+		})
+		klog.InfoS("Registered VLLM model preset", "model", m.Name)
+	}
+}
+
+type vllmModel struct {
```
zhuangqh: vLLMCompatibleModel

presets/workspace/inference/chat_templates/minimax_m1.jinja
```diff
@@ -0,0 +1,91 @@
+{{ '<begin_of_document>' -}}
```
zhuangqh: If you want to build base image, you need to bump the tag in https://github.com/kaito-project/kaito/blob/main/presets/workspace/models/supported_models.yaml#L6.

Let's think about it. Is it possible to decouple thes static files from base image?
Otherwise, update this file need to bump base image and controller tag.
We can consider downloading this small template file from a remote registry.

andyzhangx: then we could just upload to `presets/workspace/inference/chat_templates/`, and always download from this kaito dir.

andyzhangx: I have removed the two jinja files in this PR first since it's not needed at least for this PR, I think we could download all vllm jinja files in one shot, and baked in base image since that's not frequently changed: https://github.com/vllm-project/vllm/tree/main/examples. We could improve this later if there is new jinja file added for new model.

andyzhangx:
> Since this PR focuses on the flow of using the best effort model yaml, I would suggest just putting one test model in the yaml file. The entire best effort model yaml file is supposed to be autogenerated later.

@Fei-Guo I have used two models in best effort model yaml now, one for single node e2e test, and one for multi-node e2e test, and the two e2e tests are passing now, this PR is ready for review, thanks.

PR #1727, opened 2026-01-15
feat: support generic huggingface vLLM inference model

andyzhangx:
**Reason for Change**:

feat: support generic huggingface model, with this PR, user could set a huggingface model card ID in KAITO workspace or InferenceSet config, e.g. `Qwen/Qwen2-7B`, kaito controller would download the model file from https://huggingface.co/Qwen/Qwen2-7B, and then create vLLM inference workload, that would make KAITO support generic huggingface vLLM model without pre-defining model.

Note, with this PR, `inference.preset.name` would accept huggingface model card ID, e.g. `Qwen/Qwen2-7B` which represents model: https://huggingface.co/Qwen/Qwen2-7B

```yaml
apiVersion: kaito.sh/v1alpha1
kind: InferenceSet
metadata:
  name: qwen2-7b
spec:
  replicas: 1
  labelSelector:
    matchLabels:
      apps: qwen2-7b
  template:
    resource:
    inference:
      preset:
        name: Qwen/Qwen2-7B
```

Note: I have not added `chat-template` parse in this PR since it requires downloading the corresponding chat-template file, and since this `chat-template` is optional, I will address this issue in next PR.

Implementation details:

This PR parse the following 3 tables (depends on vllm version, it's now `v0.12.0`):
 - vLLM model supported architecture list list: 
     - https://github.com/vllm-project/vllm/blob/v0.12.0/docs/models/supported_models.md
 - `reasoning-parser` filed parse: 
     - https://github.com/vllm-project/vllm/blob/v0.12.0/docs/features/reasoning_outputs.md
 - `tool-call-parser` filed parse: 
     - https://github.com/vllm-project/vllm/blob/v0.12.0/docs/features/tool_calling.md

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/utils/plugin/plugin.go
+++ b/pkg/utils/plugin/plugin.go
@@ -83,9 +83,9 @@ func (reg *ModelRegister) Has(name string) bool {
 	return ok
 }
 
-// IsValidPreset return true
-// 1. if the given preset name is registered in the KaitoModelRegister.
-// 2. if the given preset name is a valid huggingface model card ID, e.g. "Qwen/Qwen2.5-Coder-7B-Instruct"
+// IsValidPreset returns true if:
+// 1. the given preset name is registered in the KaitoModelRegister.
+// 2. the given preset name is a valid huggingface model card ID, e.g. "Qwen/Qwen2.5-Coder-7B-Instruct"
 func IsValidPreset(preset string) bool {
 	if KaitoModelRegister.Has(preset) {
 		return true
--- a/presets/workspace/models/vllm_model.go
+++ b/presets/workspace/models/vllm_model.go
@@ -103,15 +103,16 @@ func GetModelByName(ctx context.Context, modelName, secretName, secretNamespace 
 
 	// if name contains "/", get model data from HuggingFace
 	if strings.Contains(modelName, "/") {
-		if shortName, ok := builtinVLLMModels[modelName]; ok {
-			klog.InfoS("Using built-in VLLM model preset", "model", modelName, "shortName", shortName)
-			return plugin.KaitoModelRegister.MustGet(shortName), nil
+		if builtinModelName, ok := builtinVLLMModels[modelName]; ok {
+			klog.InfoS("Using built-in VLLM model preset", "model", modelName, "builtinModelName", builtinModelName)
+			return plugin.KaitoModelRegister.MustGet(builtinModelName), nil
 		}
 
 		klog.InfoS("Generating VLLM model preset for HuggingFace model", "model", modelName, "secretName", secretName, "secretNamespace", secretNamespace)
 		token, err := GetHFTokenFromSecret(ctx, kubeClient, secretName, secretNamespace)
 		if err != nil {
-			klog.ErrorS(err, "Failed to get HF token from secret", "secretName", secretName, "secretNamespace", secretNamespace)
+			// only log the error here since token may not be required for public models
+			klog.ErrorS(err, "failed to get huggingface token from secret", "secretName", secretName, "secretNamespace", secretNamespace)
 		}
 		param, err := generator.GeneratePreset(modelName, token)
 		if err != nil {


pkg/utils/plugin/plugin.go
```diff
@@ -27,6 +28,11 @@ type Registration struct {
 	// retrieve the model metadata and instance.
 	Name string
 
+	// HFModelRepoName is the name of the model on HuggingFace. It is used to
+	// retrieve the model metadata and instance from HuggingFace.
+	// example: Qwen/Qwen2-7B
+	HFModelRepoName string
```
Fei-Guo: Change to ModelCardID ? 

"In the context of the Hugging Face (HF) Hub, the Model Card ID is the unique identifier for a model repository"

presets/workspace/models/vllm_model_arch_list.go
```diff
@@ -0,0 +1,189 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package models
+
+var vLLMModelArchMap = map[string]string{
```
zhuangqh: what about embed a yaml config for this? It would be easier to maintain than go code.

andyzhangx: I am using a script to generate the list automatically from website, so yaml config or .go file is actually the same approach, and using .go file to define a map is more straightforward and does not change the build setting. Let me add the script into this PR, that would be more convinced.

andyzhangx: `presets/workspace/generator/gen_model_arch_list.sh` is for generating `vLLMModelArchMap` in go, that's more  straightforward

presets/workspace/models/vllm_model.go
```diff
@@ -14,43 +14,78 @@
...
+	// if name contains "/", get model data from HuggingFace
+	if strings.Contains(modelName, "/") {
+		klog.InfoS("Generating VLLM model preset for HuggingFace model", "model", modelName, "secretName", secretName, "secretNamespace", secretNamespace)
+		token, err := GetHFTokenFromSecret(ctx, kubeClient, secretName, secretNamespace)
+		if err != nil {
+			klog.ErrorS(err, "Failed to get HF token from secret", "secretName", secretName, "secretNamespace", secretNamespace)
+		}
+		param, err := generator.GeneratePreset(modelName, token)
+		if err != nil {
+			panic("could not generate preset for model: " + modelName + ", error: " + err.Error())
```
zhuangqh: This is an expected error. should return an error message back to webhook instead of panic.

presets/workspace/generator/generator.go
```diff
@@ -234,11 +272,53 @@ func (g *Generator) ParseModelMetadata() {
...
+			}
+		}
+	}
+
+	// Override architectures for specific model families only when none were parsed
+	if len(g.Param.Metadata.Architectures) == 0 {
+		if strings.HasPrefix(g.Param.Metadata.Name, "mistral-large-3") {
+			g.Param.Metadata.Architectures = []string{"MistralLarge3ForCausalLM"}
+		} else if strings.HasPrefix(g.Param.Metadata.Name, "ministral-3") {
+			g.Param.Metadata.Architectures = []string{"MistralForCausalLM"}
```
zhuangqh: This one is `Mistral3ForConditionalGeneration`.
`MistralForCausalLM` is the arch for mistral v1 model which released one year ago.

presets/workspace/models/vllm_model.go
```diff
@@ -14,43 +14,78 @@
...
+// secretName and secretNamespace using kubeClient and ctx.
+//
+// The returned model.Model represents the registered or newly generated model.
+// If preset generation fails, or if the modelName does not correspond to a
+// registered or generatable model, this method panics instead of returning an
+// error. Callers should ensure that modelName is valid and be aware of this
+// panic behavior when integrating this method.
+func GetModelByName(ctx context.Context, modelName, secretName, secretNamespace string, kubeClient client.Client) model.Model {
+	modelName = strings.ToLower(modelName)
+	model := plugin.KaitoModelRegister.MustGet(modelName)
```
zhuangqh, line 99: This method will panic. Add a Get method?

presets/workspace/models/vllm_model.go
andyzhangx: fixed

presets/workspace/models/vllm_model.go
andyzhangx: I have changed the behavior of panic, it's now returning error which is more recommended.

pkg/utils/plugin/plugin.go
```diff
@@ -83,5 +84,5 @@ func (reg *ModelRegister) Has(name string) bool {
 }
 
 func IsValidPreset(preset string) bool {
-	return KaitoModelRegister.Has(preset)
+	return KaitoModelRegister.Has(preset) || strings.Contains(preset, "/")
```
Fei-Guo: I think it is worth adding a Regex check for the model ID:  namespace/name.

presets/workspace/generator/generator.go
Fei-Guo: Where is this Config field used?

presets/workspace/generator/generator.go
```diff
@@ -234,11 +272,53 @@ func (g *Generator) ParseModelMetadata() {
...
+	// set reasoning parser based on model name prefix
+	for prefix, parser := range reasoningParserMap {
+		if strings.HasPrefix(g.Param.Metadata.Name, prefix) {
+			g.Param.Metadata.ReasoningParser = parser
+			break
+		}
+	}
+
+	// set tool call parser based on model name prefix
+	for prefix, parser := range toolCallParserMap {
```
Fei-Guo: If you change the toolcallParserMap to an array, and Do Not `break` for a matching prefix, maybe you can avoid extra checks at 311 and 315. Given the size of the array, I think it is ok to scan all from performance perspective.

presets/workspace/generator/generator.go
Fei-Guo, line 195: Can we double check if we need LoadFormat, configFormat and TokenizerMode fields? It seems that the defaults are auto for all of them and only Mistral has a special handling here. And there is a special handling in presets/workspace/models/vllm_model.go  line 133. already.

zhuangqh: It's required for mistralai/Mistral-Large-3 series model. @andyzhangx

andyzhangx: yes, here are  reference doc of the above 3 parameters : https://docs.vllm.ai/en/v0.8.5/serving/engine_args.html, by default it's `auto`, so I have removed the hardcoding logic in `presets/workspace/models/vllm_model.go` since it's not necessary

presets/workspace/generator/generator.go
andyzhangx: I have sorted the keys in map, and then compare, so there is no extra check now

presets/workspace/generator/generator.go
andyzhangx: removed

pkg/utils/plugin/plugin.go
andyzhangx: added a more restrict check for `a/b` format

PR #1732, opened 2026-01-19
feat: add retrieve API for RAG service

bangqipropel:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/examples/gateway-api-inference-extension/destinationrule-mistral-7b-instruct.yaml
+++ b/examples/gateway-api-inference-extension/destinationrule-mistral-7b-instruct.yaml
@@ -1,9 +1,9 @@
 apiVersion: networking.istio.io/v1
 kind: DestinationRule
 metadata:
-  name: mistral-7b-instruct-inference-epp
+  name: mistral-7b-instruct-inferencepool-epp
 spec:
-  host: mistral-7b-instruct-inference-epp
+  host: mistral-7b-instruct-inferencepool-epp
   trafficPolicy:
     tls:
       mode: SIMPLE
--- a/examples/gateway-api-inference-extension/gateway.yaml
+++ b/examples/gateway-api-inference-extension/gateway.yaml
@@ -4,6 +4,13 @@ metadata:
   name: inference-gateway
 spec:
   gatewayClassName: istio
+  # NOTE: Uncomment the following section allow incoming traffic from Azure external load balancers.
+  #       See: https://istio.io/latest/docs/setup/platform-setup/azure/ for more details.
+  # infrastructure:
+  #   annotations:
+  #     service.beta.kubernetes.io/port_80_health-probe_protocol: tcp
+  #     service.beta.kubernetes.io/port_443_health-probe_protocol: tcp
+  #     service.beta.kubernetes.io/port_15021_health-probe_protocol: tcp
   listeners:
   - name: http
     port: 80
--- a/website/docs/inference.md
+++ b/website/docs/inference.md
@@ -27,6 +27,27 @@ inference:
     name: "falcon-7b"
 ```
 
+:::info
+
+Starting from KAITO v0.9.0, generic Hugging Face models are supported on a best-effort basis. By specifying a Hugging Face model card ID as `inference.preset.name` in the KAITO workspace, you can run any Hugging Face model with a model architecture supported by vLLM on KAITO. Below is an example illustrating how to create a Hugging Face inference workload using the model card ID `Qwen/Qwen3-0.6B` from https://huggingface.co/Qwen/Qwen3-0.6B:
+
+```yaml
+apiVersion: kaito.sh/v1beta1
+kind: Workspace
+metadata:
+  name: qwen3-06b
+resource:
+  instanceType: Standard_NC24ads_A100_v4
+  labelSelector:
+    matchLabels:
+      apps: qwen3-06b
+inference:
+  preset:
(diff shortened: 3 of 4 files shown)


presets/ragengine/vector_store/base.py
```diff
@@ -812,3 +814,158 @@ async def _load_internal(self, index_name: str, path: str, overwrite: bool):
...
+            chat_history = []
+
+            if not user_prompt or user_prompt.strip() == "":
+                raise HTTPException(
+                    status_code=400,
+                    detail="Query string cannot be empty.",
+                )
+
+            prompt_len = self.llm.count_tokens(user_prompt)
+            if prompt_len > self.llm.metadata.context_window:
```
zhuangqh: context window was set in `InferenceServiceSpec.ContextWindowSize` previously. 
how to set this field now?

presets/ragengine/models.py
```diff
@@ -66,6 +66,31 @@ class NodeWithScore(BaseModel):
     metadata: dict | None = None
 
 
+class RetrievalRequest(BaseModel):
+    index_name: str = Field(..., description="Name of the index to retrieve from")
+    query: str = Field(..., description="User query string for retrieval")
+    context_token_ratio: float | None = Field(
```
bfoley13: I'm of the opinion we should remove `context_token_ratio` and `max_tokens` in favor of a single `max_node_count` param since there is no "Context Window" we need to worry about within this function. The MCP likely wont know about these things, and other services that just want to collect context metadata might not either

presets/ragengine/vector_store/base.py
bfoley13: If we go the route of only having a `max_node_count` we dont need to care about the context_window vs prompt size or max_tokens.

presets/ragengine/vector_store/base.py
```diff
@@ -812,3 +814,158 @@ async def _load_internal(self, index_name: str, path: str, overwrite: bool):
...
+                )
+
+            if (
+                max_tokens
+                and max_tokens > self.llm.metadata.context_window - prompt_len
+            ):
+                max_tokens = self.llm.metadata.context_window - prompt_len
+
+            # Calculate top_k same as chat_completion
+            top_k = max(
```
bfoley13: `top_k` would then be set to `max_node_count`

presets/ragengine/vector_store/base.py
```diff
@@ -812,3 +814,130 @@ async def _load_internal(self, index_name: str, path: str, overwrite: bool):
...
+            chat_history = []
+
+            if not user_prompt or user_prompt.strip() == "":
+                raise HTTPException(
+                    status_code=400,
+                    detail="Query string cannot be empty.",
+                )
+
+            # Use max_node_count directly as top_k
+            top_k = max_node_count
```
bfoley13: We might want to add an upper limit on this, maybe 300 to start? In an environment variable so we can override easily if need be

presets/ragengine/vector_store/base.py
```diff
@@ -812,3 +814,130 @@ async def _load_internal(self, index_name: str, path: str, overwrite: bool):
         except Exception as e:
             logger.error(f"Failed to load index {index_name}. Error: {str(e)}")
             raise HTTPException(status_code=500, detail=f"Loading failed: {str(e)}")
+
+    async def retrieval(
```
bfoley13: And can we get some unit tests for this as well

presets/ragengine/models.py
```diff
@@ -66,6 +66,28 @@ class NodeWithScore(BaseModel):
     metadata: dict | None = None
 
 
+class RetrievalRequest(BaseModel):
+    index_name: str = Field(..., description="Name of the index to retrieve from")
+    query: str = Field(..., description="User query string for retrieval")
+    max_node_count: int = Field(
+        default=5,
+        ge=1,
+        le=300,
```
bfoley13: This also needs to be set to the RAG_MAX_TOP_K env variable or we risk request validation failure if they are different

presets/ragengine/metrics/prometheus_metrics.py
```diff
@@ -95,6 +95,18 @@
     labelnames=[STATUS_LABEL],
 )
 
+# Indexes search API metrics
+rag_indexes_search_latency = Histogram(
+    "rag_indexes_search_latency_seconds",
+    "Time to call post '/search' API in seconds",
```
Fei-Guo: For these two methods, %s/search/retrieve/g ?

presets/ragengine/main.py
```diff
@@ -652,6 +664,75 @@ async def load_index(
         rag_load_latency.labels(status=status).observe(time.perf_counter() - start_time)
 
 
+@app.post(
+    "/retrieval",
```
Fei-Guo: retrieval is a noun. Should we use the verb form? retrieve.

presets/ragengine/main.py
```diff
@@ -652,6 +664,75 @@ async def load_index(
...
+          "score": 0.85,
+          "metadata": {"category": "technical", "author": "John Doe"}
+        }
+      ],
+      "count": 5
+    }
+    ```
+    """,
+)
+async def retrieval_index(request: RetrievalRequest):
```
Fei-Guo: retrieval_index -> retrieve_from_index

PR #1745, opened 2026-01-27
feat: add status field into WorkspaceStatus for presenting Workspace current state.

rambohe-ch:
**Reason for Change**:

According to the proposal: https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260121-workspace-status-phase.md, Field status is added into WorkspaceStatus for presenting current state of Workspace instance.

btw: The improvement of Workspace in kubectl tool output: TargetNodeCount and Status are output instead of original each conditions. 

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/api/v1beta1/workspace_types.go
+++ b/api/v1beta1/workspace_types.go
@@ -199,9 +199,9 @@ type WorkspaceStatus struct {
 	// +optional
 	WorkerNodes []string `json:"workerNodes,omitempty"`
 
-	// Status represents the current high-level state of the workspace.
+	// State represents the current high-level state of the workspace.
 	// +optional
-	Status WorkspaceState `json:"status,omitempty"`
+	State WorkspaceState `json:"state,omitempty"`
 
 	// Conditions report the current conditions of the workspace.
 	// +optional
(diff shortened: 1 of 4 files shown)


api/v1beta1/workspace_types.go
```diff
@@ -176,12 +176,33 @@ type TuningSpec struct {
 	Output *DataDestination `json:"output"`
 }
 
+// WorkspaceState indicates the high-level state of the workspace.
+type WorkspaceState string
+
+const (
```
Fei-Guo, line 182: Should we make the api name and struct name consistent? 
Either Status or state is ok, but not both.

api/v1beta1/workspace_types.go
```diff
@@ -197,10 +218,8 @@ type WorkspaceStatus struct {
 // +kubebuilder:resource:path=workspaces,scope=Namespaced,categories=workspace,shortName={wk,wks}
 // +kubebuilder:storageversion
 // +kubebuilder:printcolumn:name="Instance",type="string",JSONPath=".resource.instanceType",description=""
-// +kubebuilder:printcolumn:name="ResourceReady",type="string",JSONPath=".status.conditions[?(@.type==\"ResourceReady\")].status",description=""
-// +kubebuilder:printcolumn:name="InferenceReady",type="string",JSONPath=".status.conditions[?(@.type==\"InferenceReady\")].status",description=""
-// +kubebuilder:printcolumn:name="JobStarted",type="string",JSONPath=".status.conditions[?(@.type==\"JobStarted\")].status",description=""
-// +kubebuilder:printcolumn:name="WorkspaceSucceeded",type="string",JSONPath=".status.conditions[?(@.type==\"WorkspaceSucceeded\")].status",description=""
+// +kubebuilder:printcolumn:name="TargetNodeCount",type="integer",JSONPath=".status.targetNodeCount",description=""
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=".status.status",description=""
```
Fei-Guo: Should we keep the old behavior of reporting condition? This seems to be a UX change that is not backward compatible.

api/v1beta1/workspace_types.go
rambohe-ch: fixed, unified as the `State`

api/v1beta1/workspace_types.go
rambohe-ch: fixed

Fei-Guo (approved):

PR #1747, opened 2026-01-28
fix: trivy unknown severitis for test/kaito-base image

rambohe-ch:
**Reason for Change**:

1. I've met the trivy error in this pull request: https://github.com/kaito-project/kaito/actions/runs/21424380339/job/61690448685, and there are 36 unknown severities for test/kaito-base:v0.0.1 image.

in this pull request, we update Debian package handling in [Dockerfile] to run apt-get install related libs, so security-fixed packages (like OpenSSL) get pulled in during image build. 

2. install opencv related libs for ministral model depends on them.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/docker/presets/models/tfs/Dockerfile
+++ b/docker/presets/models/tfs/Dockerfile
@@ -12,7 +12,15 @@ RUN pip3 install --upgrade pip uv
 WORKDIR /workspace
 
 # Required for torch.compile
-RUN apt-get update -y && apt-get install --no-install-recommends curl gcc libc-dev perl -y && \
+RUN apt-get update -y && \
+    if dpkg -s libssl3t64 >/dev/null 2>&1; then \
+        apt-get install --only-upgrade --no-install-recommends -y openssl libssl3t64 openssl-provider-legacy; \
+    elif dpkg -s libssl3 >/dev/null 2>&1; then \
+        apt-get install --only-upgrade --no-install-recommends -y openssl libssl3; \
+    else \
+        apt-get install --only-upgrade --no-install-recommends -y openssl; \
+    fi && \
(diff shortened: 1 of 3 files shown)


zhuangqh:
please bump the base image tag in kaito/presets/workspace/models/support_models.yaml

rambohe-ch:
> please bump the base image tag in kaito/presets/workspace/models/support_models.yaml

fixed, PTAL

zhuangqh:
Since apt upgrade may bring potential sys lib compability issues, I suggest update problematic lib only.

zhuangqh:
Upgrade Compatibility Test is broken because the base image tag was bumped but the upgrade test try to use it.
Skip this test as this test should be recovered once official base image is released.

docker/presets/models/tfs/Dockerfile
```diff
@@ -12,7 +12,15 @@ RUN pip3 install --upgrade pip uv
 WORKDIR /workspace
 
 # Required for torch.compile
-RUN apt-get update -y && apt-get install --no-install-recommends curl gcc libc-dev perl -y && \
+RUN apt-get update -y && \
+    if dpkg -s libssl3t64 >/dev/null 2>&1; then \
```
andyzhangx, line 16: how did you make this change? @rambohe-ch 
I got similar error like in another project when upgrading python base image .e.g `FROM python:3.12-slim`
```
vllm-router: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

do I need to make the similar change?

andyzhangx: nvm, that's different issue

PR #1748, opened 2026-01-28
chore: force annotation and label values to be strings

t0rr3sp3dr0:
**Reason for Change**:
Due to a limitation of the Extensions RP, we need to force annotations and labels to have string values even when a different type is passed to Helm. This will allow us to set Prometheus annotations that would otherwise be impossible to set when installing KAITO as an extension:

```yaml
prometheus.io/port: "1234"
prometheus.io/scrape: "true"
```


zhuangqh:
please run compatiblity test before merging.

PR #1802, opened 2026-02-27
feat: improve logs of workspace controller

rambohe-ch:
**Reason for Change**:

1. upgrade readme with latest workspace kubectl output
2. improve default-node-image-family startup parameter can be set only when value is not empty in helm chart
3. improve service creation only if sevice doesn't exist, so don't log `create service` in each reconcile.
~~4. add cache for CheckIfNodePluginsReady function, so only output `node plugin` one login in a reconcile.~~
5. add debug logs for syncWorkspaceStatus function when status changes
6. update logs of nodeclaim
7. fix one bug: deepcopy workspace status for syncWorkspaceStatus

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/workspace/controllers/workspace_controller.go
+++ b/pkg/workspace/controllers/workspace_controller.go
@@ -103,8 +103,6 @@ func (c *WorkspaceReconciler) SetDefaultNodeImageFamily(defaultNodeImageFamily s
 }
 
 func (c *WorkspaceReconciler) Reconcile(ctx context.Context, req reconcile.Request) (result reconcile.Result, err error) {
-	ctx = c.nodeResourceManager.PrepareContext(ctx)
-
 	workspaceObj := &kaitov1beta1.Workspace{}
 	if err = c.Client.Get(ctx, req.NamespacedName, workspaceObj); err != nil {
 		if apierrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


pkg/workspace/controllers/workspace_controller.go
```diff
@@ -845,10 +858,78 @@ func (c *WorkspaceReconciler) updateWorkspaceStatusIfChanged(ctx context.Context
 				return nil
 			}
 
+			klog.InfoS("Workspace status changed",
```
zhuangqh: if klog.V(4).Enabled() {
	}

rambohe-ch: fixed

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -845,10 +858,78 @@ func (c *WorkspaceReconciler) updateWorkspaceStatusIfChanged(ctx context.Context
...
+				oldCondition.ObservedGeneration, newCondition.ObservedGeneration))
+		}
+	}
+
+	if len(changes) == 0 {
+		return "status changed"
+	}
+
+	return strings.Join(changes, "; ")
+}
```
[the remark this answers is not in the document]
rambohe-ch: only debug logs

PR #1806, opened 2026-02-27
feat: add Qdrant vector store backend with hybrid search support

bangqipropel:
- Add QdrantVectorStoreHandler with LlamaIndex hybrid search (dense + BM25 sparse)
- Add VectorDBConfig to RAGEngine CRD (type, url, accessSecret)
- Wire VECTOR_DB_TYPE/URL/SECRET env vars from CRD to ragengine pod
- Add HybridRetriever (Python-side BM25 + dense fusion) for non-Qdrant backends
- Add Prometheus metrics: hybrid search mode, encoding latency, vector store stats
- Fix restore: use from_vector_store() + docstore rebuild (zero GPU, avoids CUDA OOM)
- Refactor base.py: extract _create_storage_context_for_load, add _use_async_indexing
- Move FAISS binary-format fallback to FaissVectorStoreHandler
- Add qdrant-client==1.16.2, fastembed>=0.4.0, llama-index-vector-stores-qdrant==0.6.0
- Add Qdrant deployment example and unit tests

**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


api/v1beta1/ragengine_types.go
```diff
@@ -26,11 +26,33 @@ type PersistentVolumeConfig struct {
 	MountPath string `json:"mountPath,omitempty"`
 }
 
+// VectorDBConfig specifies the vector database backend configuration.
+type VectorDBConfig struct {
+	// Type specifies the vector database backend to use.
+	// Supported values: "faiss" (default, in-process), "qdrant" (client-server).
```
Fei-Guo: I don't think there should have a default for FAISS here. If we use FAISS, the entire VectorDB field should be unset. The default can be qdrant, or we make Type as a required field. 

Type seems to indicate if the db is a vector db or a rational DB. How about rename it to `backend engine`/`engine`? In the API comment, we need to clearly mention those engines have to be the llamaindex supported engines. 

I also think URL needs to be a required field.

presets/ragengine/vector_store/retriever/hybrid_retriever.py
```diff
@@ -0,0 +1,235 @@
...
+                return None
+            return BM25Retriever.from_defaults(
+                docstore=self._index.docstore,
+                similarity_top_k=top_k,
+            )
+        except Exception as e:
+            logger.warning(f"Failed to build BM25 retriever: {e}. Falling back to vector-only.")
+            return None
+
+    def _fuse(
```
Fei-Guo: Do you invent the fuse algorithm? or there is a best practice for fuse algorithm?

pkg/ragengine/manifests/manifests.go
```diff
@@ -186,11 +186,41 @@ func RAGSetEnv(ragEngineObj *kaitov1beta1.RAGEngine) []corev1.EnvVar {
...
-		Value: "faiss", // TODO: get storage done
+		Value: vectorDBType,
+	}
+	envs = append(envs, storageEnv)
+
+	// Inject vector DB connection info if configured
+	if ragEngineObj.Spec.Storage != nil && ragEngineObj.Spec.Storage.VectorDB != nil {
+		if ragEngineObj.Spec.Storage.VectorDB.URL != "" {
+			envs = append(envs, corev1.EnvVar{
+				Name:  "VECTOR_DB_URL",
```
bfoley13: pull out the env keys into consts and add some tests for the manifest generation

presets/ragengine/main.py
```diff
@@ -129,13 +135,73 @@ async def track_requests(request: Request, call_next):
...
+    )
 
 # Initialize RAG operations
 rag_ops = VectorStoreManager(vector_store_handler)
 
+# Background task: periodically collect vector store backend metrics
+_metrics_task = None
+
+
+async def _collect_vector_store_metrics():
```
Fei-Guo: Why do we need to proactively pull how many points in the collection? It is a constant value if no change to the indexes, right?

presets/ragengine/metrics/prometheus_metrics.py
```diff
@@ -172,6 +172,30 @@
     "num_requests_running", "Number of requests currently being processed"
 )
 
+# Vector store metrics (Qdrant / backend-specific)
```
Fei-Guo: Are the metrics defined this PR used anywhere in the PR? e.g., I don't see rag_vector_store_operation_latency  is used anywhere.

presets/ragengine/vector_store/base.py
```diff
@@ -775,6 +793,16 @@ async def load(self, index_name: str, path: str, overwrite: bool):
         else:
             await self._load_internal(index_name, path, overwrite)
 
+    def _create_storage_context_for_load(
```
Fei-Guo: I don't see why we need this method. index_name is not used.

presets/ragengine/main.py
```diff
@@ -129,13 +135,73 @@ async def track_requests(request: Request, call_next):
...
+    vector_store_handler = FaissVectorStoreHandler(embedding_manager)
+elif VECTOR_DB_TYPE.lower() == "qdrant":
+    from vector_store.qdrant_store import QdrantVectorStoreHandler
+
+    vector_store_handler = QdrantVectorStoreHandler(
+        embedding_manager,
+        vector_db_url=VECTOR_DB_URL,
+        vector_db_access_secret=VECTOR_DB_ACCESS_SECRET,
+    )
+elif VECTOR_DB_TYPE.lower() == "milvus":
```
Fei-Guo: Let us remove milvus since the package is not included in this PR.

PR #1818, opened 2026-03-02
feat: add multi-platform (amd64/arm64) support for workspace image

sozercan:
## Summary

Add arm64 support to the workspace image by enabling multi-platform builds using Go's native cross-compilation.

## Changes

- **`docker/workspace/Dockerfile`**: Remove `--platform=$BUILDPLATFORM` from the final distroless stage so it uses the target platform instead of the build host platform
- **`Makefile`**: Update `docker-build-workspace` target to build for `linux/amd64,linux/arm64`

## How it works

No QEMU emulation is needed because:
- The Go build stages already use `--platform=$BUILDPLATFORM` and cross-compile via `CGO_ENABLED=0 GOARCH=${TARGETARCH}`
- The final stage only uses `COPY` (no `RUN`), so BuildKit can create layers for each target architecture natively

### DIFF
--- a/.github/workflows/aikit.yaml
+++ b/.github/workflows/aikit.yaml
@@ -62,7 +62,7 @@ jobs:
       - name: Build workspace controller
         run: |
           # Use the docker-build-workspace target with local configuration
-          make docker-build-workspace
+          make docker-build-workspace ARCH=amd64
 
           # Load into kind cluster
           kind load docker-image local/workspace:test --name kaito
@@ -71,7 +71,6 @@ jobs:
           IMG_NAME: "workspace"
           IMG_TAG: "test"
           OUTPUT_TYPE: "type=docker"
-          ARCH: "amd64"
 
       - name: Deploy KAITO workspace controller
         run: |
--- a/.github/workflows/trivy.yaml
+++ b/.github/workflows/trivy.yaml
@@ -36,8 +36,7 @@ jobs:
           export REGISTRY=${{ matrix.registry }}
           export IMG_TAG=latest
           export OUTPUT_TYPE="type=docker"
-          export ARCH="amd64"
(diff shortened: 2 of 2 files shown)


andyzhangx:
have we run the e2e test with this change?

it now reports: https://github.com/kaito-project/kaito/actions/runs/22893103295/job/66420494505?pr=1832
```
Run make docker-build-workspace
img-builder        docker-container                                       
 \_ img-builder0    \_ unix:///var/run/docker.sock   running   0.18.1     linux/amd64 (+4), linux/arm64, linux/arm (+2), linux/ppc64le, (4 more)
docker buildx build \
	--file ./docker/workspace/Dockerfile \
	--output=type=registry \
	--platform="linux/amd64,linux/arm64" \
	--build-arg VERSION=3f3e10ca \
	--build-arg BUILD_DATE=2026-03-10T08:16:04Z \
	--pull \
	 \
	--tag kaitori22893103295rn904ra1rbxz.azurecr.io/workspace:ri22893103295rn904ra1rbxz .
ERROR: failed to build: Multi-platform build is not supported for the docker driver.
Switch to a different driver, or turn on the containerd image store, and try again.
Learn more at https://docs.docker.com/go/build-multi-platform/
make: *** [Makefile:306: docker-build-workspace] Error 1
```

andyzhangx:
at least for workspace controller, it does not need to run on arm64 node

Makefile
```diff
@@ -301,6 +301,7 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	fi
 
 .PHONY: docker-build-workspace
+docker-build-workspace: ARCH = amd64,arm64
```
andyzhangx, line 304: it should be `docker-build-workspace: ARCH = amd64,linux/arm64`

PR #1823, opened 2026-03-03
feat: add NVIDIA MIG support for BYO nodes (#1744)

### LINKED ISSUE #1744: Feature Request: Support for NVIDIA Multi-Instance GPU (MIG)
## Description

Currently, KAITO allocates full GPUs to model workloads. For smaller LLMs (e.g., Phi-3, Mistral 7B with quantization), a full A100 or H100 is often overkill, leading to underutilized hardware and higher costs. We propose adding support for **NVIDIA MIG**, allowing users to partition a single physical GPU into multiple instances to run several small models concurrently on a single node.

## Use Case

**Efficient Multi-Model Hosting:** A user wants to run 4 small model instances (e.g., 4x Phi-4-mini) on a single Azure `Standard_ND96asr_v4` node (which has 8x A100 GPUs). By using MIG, they could theoretically host dozens of small models on that same node by partitioning each GPU.

## Proposed Scope: BYO (Bring Your Own) Node First

To reduce initial complexity, this feature will focus on **BYO Nodes**:

* The infrastructure admin is responsible for enabling MIG mode on the GPUs and creating the partitions (e.g., `1g.10gb`) at the OS/Node level.
* KAITO will focus on **discovery and scheduling** rather than hardware partitioning management.

## Technical Challenges

1. **Resource Specification:** Kubernetes represents MIG instances as unique extended resources (e.g., `nvidia.com/mig-1g.10gb`). KAITO's current `Resource` spec needs to handle these instead of standard `nvidia.com/gpu`.
2. **VRAM Estimator:** The internal KAITO estimator must be updated to understand that a `1g.10gb` profile has specific memory limits, preventing it from trying to schedule a model that requires 20GB of VRAM onto a 10GB MIG partition.
3. **Workflow Decoupling:** This should be decoupled from the main Node Autoprovisioning (NAP) workflow initially, as NAP requires complex integration with cloud-specific APIs to request MIG-enabled pools.

## Proposed API Changes

Add support for MIG resource strings in the `Kaitocontrol` / `Workspace` CRD.

```yaml
apiVersion: kaito.sh/v1alpha1
kind: Workspace
metadata:
  name: phi-3-mig
spec:
  offloading: false
  resource:
    instanceType: "" # Leave empty for BYO / Manual scheduling
    labelSelector:
      matchLabels:
        apps: mig-enabled-node
    # New capability: requesting specific MIG resources
    resourceRequests:
      nvidia.com/mig-1g.10gb: 1 
  inference:
    preset:
      name: "phi-3-mini-4k-instruct"

```

## Implementation Strategy

* **Partition Management:** **Customer Managed.** The user configures the MIG strategy (Single Strategy or Mixed Strategy) on the node. KAITO simply consumes the available resources advertised by the NVIDIA Device Plugin.
* **Scheduler:** KAITO should ensure the `count` in the resource request aligns with how many MIG instances are required per replica.
* **Estimator Logic:** * Enhance the estimator to map `nvidia.com/mig-x.ygb` to a memory capacity of `y` GB.
* If the model's estimated footprint > `y`, the admission controller should reject the Workspace creation with a descriptive error.

## References

* [[NVIDIA MIG Device Enumeration & Names](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/mig-device-names.html)](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/mig-device-names.html)
* [[Promoting MIG Resources in Kubernetes](https://www.google.com/search?q=https://docs.nvidia.com/datacenter/cloud-native/kubernetes/latest/mig-k8s.html)](https://www.google.com/search?q=https://docs.nvidia.com/datacenter/cloud-native/kubernetes/latest/mig-k8s.html)

robert-cronin:
**Reason for Change**:
Add NVIDIA Multi-Instance GPU (MIG) support for BYO nodes. Smaller models (e.g. phi-4-mini) don't need a full A100/H100, MIG lets users partition a single GPU and run multiple models concurrently, reducing cost and improving utilization.

**Requirements**

**Issue Fixed**:
Fixes #1744

**Notes for Reviewers**:
This is gated behind the `enableMIG` feature flag and requires `disableNodeAutoProvisioning=true` (BYO-only for initial scope).

Key changes:
- New `spec.resource.mig` field (profile + count) on Workspace v1beta1 CRD
- Webhook validates MIG profile is a known NVIDIA profile, model VRAM fits the partition (with 2% model overhead and 84% vLLM utilization factor), rejects MIG for tuning, and requires BYO mode
- MIG profile database for A30, A100, H100, H200, B200 from official NVIDIA docs (`pkg/utils/mig/`)
- Advanced nodes estimator validates model fits in MIG slice and returns nodeCount=1 (no multi-node for MIG)
- Inference runtime sets `nvidia.com/mig-<profile>` resource requests/limits on pods, forces tensor-parallel-size=1, and adds MIG-specific toleration
- Example manifest at `examples/mig/workspace-mig.yaml`


Fei-Guo:
https://docs.nvidia.com/datacenter/cloud-native/kubernetes/latest/index.html#the-single-strategy,  have we considered this one?

robert-cronin:
> https://docs.nvidia.com/datacenter/cloud-native/kubernetes/latest/index.html#the-single-strategy, have we considered this one?

I focused on mixed strategy here since thats what needs the nvidia.com/mig-* resource types. Single strategy uses regular nvidia.com/gpu so it hits the existing BYO path.

The issue is that the BYO path doesnt know its dealing with MIG slices, so it sets `tensor-parallel-size` to the full slice count instead of 1. So I think we would need some way to detect or declare single strategy to handle that correctly.

What were you thinking for the single strategy?

Fei-Guo:
@robert-cronin.  Thanks for pointing it out the static slice problem.  I think we need to define the proper UX first for user to use MIG feature. Do you mind starting with a design doc first instead of a direct implementation PR?  Do we expect users to run multiple kaito workspaces in one singe node? How to configure workspace cr to make this happen. Also, how to deal with static partition, is it sufficient to overwrite the tensor-parallel-size parameter via a customized configmap? Basically, I don't want after checking in this pr, users are still blocked by using MIG effectively in KAITO.

robert-cronin:
> @robert-cronin. Thanks for pointing it out the static slice problem. I think we need to define the proper UX first for user to use MIG feature. Do you mind starting with a design doc first instead of a direct implementation PR? Do we expect users to run multiple kaito workspaces in one singe node? How to configure workspace cr to make this happen. Also, how to deal with static partition, is it sufficient to overwrite the tensor-parallel-size parameter via a customized configmap? Basically, I don't want after checking in this pr, users are still blocked by using MIG effectively in KAITO.

Yeah good call, I'll put together a design doc first. The UX around multi-workspace-per-node and single strategy definitely needs more thought before implementation.

Is there a preferred format or location for design docs? I can start with a Google doc if that works.

Fei-Guo:
https://github.com/kaito-project/kaito/blob/main/docs/proposals/YYYYMMDD-proposal-template.md   @robert-cronin, here we just added the template.

robert-cronin:
> https://github.com/kaito-project/kaito/blob/main/docs/proposals/YYYYMMDD-proposal-template.md @robert-cronin, here we just added the template.

Hey @Fei-Guo I've added https://github.com/robert-cronin/kaito/blob/5a9f9c02038bb80b1ba341de43146c5db10d95d7/docs/proposals/20260330-mig-support.md as an initial attempt at a proposal.

PR #1824, opened 2026-03-04
fix: prevent webhook panic on auto-generated vLLM presets with empty GPU params

sozercan:
## Problem

For auto-generated vLLM presets (any HuggingFace model not in KAITO's built-in preset list, e.g. `Qwen/Qwen3-0.6B`), `GPUCountRequirement` is always an empty string because the generator has no logic to derive GPU count from HuggingFace metadata. Calling `resource.MustParse("")` on this empty string caused the webhook to panic with:

```
cannot parse '': quantities must match the regular expression...
```

This crashes the webhook and returns a 500 error to the user on workspace creation.

## Root Cause

- Built-in presets hardcode both `GPUCountRequirement` and `TotalSafeTensorFileSize`
- Auto-generated vLLM presets only populate `TotalSafeTensorFileSize` (from HuggingFace model metadata) but never set `GPUCountRequirement`
- The validation code called `resource.MustParse()` on both fields without checking for empty strings

## Fix

- **Independent field checks**: Instead of a single all-or-nothing OR guard, check each field independently. GPU count validation is skipped only when `GPUCountRequirement` is empty. GPU memory + distributed inference checks are skipped only when `TotalSafeTensorFileSize` is empty.
- **Safe parsing**: Replace `resource.MustParse` with `resource.ParseQuantity` with proper error handling to prevent panics on malformed input.
- **Both API versions**: Applied to both `v1alpha1` and `v1beta1` workspace validation.
- **Warning logs**: Log warnings when skipping individual checks due to empty or unparseable fields.

## Testing

Deployed the fix and validated that creating a workspace with `Qwen/Qwen3-0.6B` (a non-built-in model) no longer panics the webhook.

### DIFF
--- a/api/v1beta1/workspace_validation.go
+++ b/api/v1beta1/workspace_validation.go
@@ -475,42 +475,19 @@ func (r *ResourceSpec) validateCreateWithInference(ctx context.Context, inferenc
 			}
 			params := modelPreset.GetInferenceParameters()
 
-			machineTotalNumGPUs := resource.NewQuantity(int64(machineCount*skuConfig.GPUCount), resource.DecimalSI)
 			machineTotalGPUMem := resource.NewQuantity(int64(machineCount)*skuConfig.GPUMem.Value(), resource.BinarySI) // Total GPU memory
-
-			// GPU count check: only run if GPUCountRequirement is specified
-			if params.GPUCountRequirement == "" {
-				klog.Warningf("Skipping GPU count validation for preset %s: GPUCountRequirement not specified", presetName)
-			} else {
-				modelGPUCount, err := resource.ParseQuantity(params.GPUCountRequirement)
-				if err != nil {
-					klog.Warningf("Skipping GPU count validation for preset %s: failed to parse GPUCountRequirement %q: %v", presetName, params.GPUCountRequirement, err)
-				} else if machineTotalNumGPUs.Cmp(modelGPUCount) < 0 {
-					if bypassResourceChecks {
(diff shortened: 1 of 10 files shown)


zhuangqh:
@andyzhangx PTAL

api/v1beta1/workspace_validation.go
```diff
@@ -478,53 +478,67 @@ func (r *ResourceSpec) validateCreateWithInference(ctx context.Context, inferenc
 			machineTotalNumGPUs := resource.NewQuantity(int64(machineCount*skuConfig.GPUCount), resource.DecimalSI)
 			machineTotalGPUMem := resource.NewQuantity(int64(machineCount)*skuConfig.GPUMem.Value(), resource.BinarySI) // Total GPU memory
 
-			modelGPUCount := resource.MustParse(params.GPUCountRequirement)
-			modelTotalGPUMemory := resource.MustParse(params.TotalSafeTensorFileSize)
```
zhuangqh, line 482: once the generated preset is registered. params.TotalSafeTensorFileSize won't be empty. And the following workflow will assume this value is non-empty.

The validation about `GPUCountRequirement` is unnecessary anymore after introducing mem estimator.
Can you just remove it?

andyzhangx:
@sozercan which kaito version are you trying? `Qwen/Qwen3-0.6B` actually works using `mcr.microsoft.com/aks/kaito/workspace:0.9.0` kaito controller. empty `GPUCountRequirement` value does not matter starting from workspace:0.8.0

PR #1839, opened 2026-03-11
feat: add model TPM benchmark test and log

### LINKED ISSUE #1828: Feature Request: Startup-phase benchmarking to measure workspace TPM

Part of #1775 (TPM-aware Autoscaling Policy). The autoscaler needs a reliable per-instance TPM value to enforce `tokenPerMinuteLimit`, but there is currently no mechanism to measure the actual throughput of a workspace after it starts up. Manual or estimated values are error-prone and vary across GPU SKUs and model configurations.

Once the workspace endpoint is reachable, automatically run a short load-test against it to measure the actual sustained Tokens Per Minute (TPM). The benchmark must complete within 2 minutes. Only after the benchmark finishes is `InferenceReady=True` set, with the measured TPM written to `workspace.status.tokenPerMinute`. This value feeds into `inferenceset.status.workspaceProfile.tokenPerMinute` for use by the TPM-aware autoscaling policy. The benchmark should be re-triggered when the instance type or model preset changes.

N/A

**Additional context**

N/A

2170chm:
**Reason for Change**:

This PR introduces a benchmark stage after the model is up and before the model is marked ready for traffic. This always logs a dynamically benchmarked TPM (Token per minute, including both input and output tokens) in the format `KAITO_BENCHMARK_RESULT YYYY-MM-DDTHH:MM:SSZ {"vllm_total_tpm":<TPM>,"ttft_avg_ms":<TTFT>,"tpot_avg_ms":<TPOT>}`, which can be parsed by the controller as json and write to the custom resource. This feature helps user understand the capacity of the deployed model.
**Requirements**

**Issue Fixed**:

Closes #1828 
**Notes for Reviewers**:

#### Design

1. To enable the feature flag, add `kaito.sh/run-benchmark: "true"` under `metadata.annotations` in the workspace yaml config.

2. The benchmark test itself runs for 60 seconds. Unless specified by the user, the rate of requests is calculated dynamically based on model and GPUs to saturate the GPU. The prompt input size is 2048 and output size is 256.

3. Currently, it will always log one result regardless of any failure and will not block readiness for any errors from the benchmark stage. If anything fails in the benchmark phase (guideLLM, vLLM metrics reading, etc.), the result will be -1 for all three fields. This log is guaranteed to be there once the model is marked ready(in case the logging fails, it will exit with no log, best effort solution). The controller should take care of this edge case.

4. The benchmark test is designed to run for 60 seconds (due to latency in metrics reading, it can take a few more seconds). After that, it will wait for all in-flight benchmark test requests to drain before marking ready, which should take around 10s to 70s. If the latency is very high (something wrong likely happened or the custom concurrency rate is to high), it can take much longer to reach the ready phase to ensure the model is fully functioning when marked ready.

5. The benchmark process is spawned from the startup exec probe process, and it is configured to have a 600s timeout to not exit the benchmark midway. vLLM will remain running as PID1.

#### Issues

1. For Mistral-3-3b, the tokenizer can only be used with transformers >= 5.0.0, therefore guideLLM is installed in a site-package virtual environment to use transformers 5.0.0 without interfering with vLLM's requirements.

2. For Llama-3-3-70b deployed in a distributed manner, there are vLLM cache problems that cause the throughput to be low. It is recommended to turn vLLM cache off. Otherwise, the drain phase after the benchmark result will take much longer.

#### Example results
Example result log with tpm, tftt, and tpot:
Screenshot 2026-03-12 at 7 58 49 PM

The latest printed format is `KAITO_BENCHMARK_RESULT {"vllm_total_tpm":<TPM>}` to make it json-parsable.
gpt-oss-120b-single-node-benchmark-result
Llama-3-3-70b-distributed-benchmark-result
Mistral-3-3b-single-node-benchmark-result
Phi-4-14b-single-node-benchmark-result
Qwen2 5-32b-single-node-benchmark-result

### DIFF
--- a/presets/workspace/inference/vllm/benchmark_entrypoint.py
+++ b/presets/workspace/inference/vllm/benchmark_entrypoint.py
@@ -65,9 +65,6 @@ def _write_to_pid1(line: str, fd: int = 1) -> None:
     This makes output visible in ``kubectl logs`` even though this script runs
     as an exec probe child process whose own stdio is captured by kubelet, not
     by the container log driver.
-
-    Raises ``OSError`` on failure (no fallback — the probe's own stdio is not
-    captured by the container log driver so a fallback write would be invisible).
     """
     with open(f"/proc/1/fd/{fd}", "a") as fh:
         fh.write(line)
@@ -75,8 +72,16 @@ def _write_to_pid1(line: str, fd: int = 1) -> None:
 
 
 def _log(msg: str, tag: str = "KAITO_BENCHMARK") -> None:
+    """Format and emit a log line through PID 1's stdout.
+
+    If the write fails (e.g. ``/proc/1/fd/1`` is inaccessible), the script exits
+    immediately with code 1.
+    """
     ts = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
-    _write_to_pid1(f"{tag} {ts} {msg}\n", fd=1)
+    try:
+        _write_to_pid1(f"{tag} {ts} {msg}\n", fd=1)
+    except Exception:
+        sys.exit(1)
 
 
 # ── vLLM helpers ─────────────────────────────────────────────────────────────
@@ -351,14 +356,10 @@ def main() -> None:
         failed = True
 
     # Always emit the result line so the controller has a parseable record even on failure.
-    try:
-        _log(
-            f'{{"vllm_total_tpm":{tpm},"ttft_avg_ms":{ttft_ms},"tpot_avg_ms":{tpot_ms}}}',
-            tag="KAITO_BENCHMARK_RESULT",
-        )
-    except Exception:
-        # If logging fail, not much we can do - fail the probe
-        sys.exit(1)
+    _log(
+        f'{{"vllm_total_tpm":{tpm},"ttft_avg_ms":{ttft_ms},"tpot_avg_ms":{tpot_ms}}}',
+        tag="KAITO_BENCHMARK_RESULT",
+    )
     sys.exit(1 if failed else 0)
 
 
--- a/presets/workspace/inference/vllm/tests/test_benchmark_entrypoint.py
+++ b/presets/workspace/inference/vllm/tests/test_benchmark_entrypoint.py
@@ -65,6 +65,19 @@ def test_health_check_exception():
         assert bm._health_check() is False
 
 
+# ── _log ──────────────────────────────────────────────────────────────────────
+
+
+def test_log_exits_1_on_write_failure():
+    """If _write_to_pid1 raises, _log calls sys.exit(1) instead of propagating."""
+    with (
+        patch.object(bm, "_write_to_pid1", side_effect=OSError("no /proc/1/fd/1")),
+        pytest.raises(SystemExit) as exc_info,
+    ):
+        bm._log("some message")
+    assert exc_info.value.code == 1
+
+
 # ── _read_counter ─────────────────────────────────────────────────────────────
 
 _METRICS_BODY = b"""\


pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +462,42 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
 			readinessTimeout = defaultStartupProbeTimeout
 		}
 
+		// When benchmarking is enabled, replace the readiness probe with a sentinel-file exec probe.
+		finalCommands := commands
+		readinessProbe := defaultReadinessProbe
+		var benchmarkEnvVars []corev1.EnvVar
+		if ctx.Workspace.Annotations[v1beta1.AnnotationRunBenchmark] == "true" {
+			finalCommands = append([]string{"/bin/bash", "/workspace/vllm/benchmark_entrypoint.sh"}, commands...)
```
zhuangqh: I suggest use a seperate sidecar container to run the benchmark job. 
keep the model server easy.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +462,42 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
...
+			readinessProbe = buildBenchmarkReadinessProbe()
+
+			// Inject GPU/model dimensions so benchmark_entrypoint.sh can compute the
+			// saturation concurrency at runtime.
+			if gpuConfig != nil && inferenceParam.TotalSafeTensorFileSize != "" && inferenceParam.BytesPerToken > 0 {
```
zhuangqh: I suggest calculate the final BENCHMARK_RATE result here unless you cant get the full info here.

presets/workspace/inference/vllm/benchmark_entrypoint.sh
```diff
@@ -0,0 +1,227 @@
+#!/bin/bash
```
zhuangqh: please consider implementing this script in python. This mixed python, bash script is really hard to maintain.
with python, you can leverage hf lib and use guidellm in lib mode.
and you can add some unit test for it.

pkg/workspace/inference/preset_inferences.go
2170chm: Per our discussion, it is now implemented as a subprocess of the startup probe process.

pkg/workspace/inference/preset_inferences.go
2170chm: It is now fully calculated in go and the request rate is passed as an environment variable.

presets/workspace/inference/vllm/benchmark_entrypoint.sh
2170chm: It is now implemented in Python and unit tested.

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,339 @@
...
+
+    # Case 3: HF cache dir  (models--org--name → org/name)
+    cache_dirs = [
+        d for d in weights.iterdir() if d.is_dir() and d.name.startswith("models--")
+    ]
+    if cache_dirs:
+        parts = cache_dirs[0].name[len("models--") :].split("--", 1)
+        if len(parts) == 2:
+            return f"{parts[0]}/{parts[1]}"
+
```
zhuangqh: just read it from hf lib if it's not baked-in mode.
```python
from huggingface_hub import scan_cache_dir

cache_info = scan_cache_dir(cache_dir="/workspace/weights")
print(cache_info)
```

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,339 @@
...
+    --data-num-workers 0 prevents subprocess workers that can fail in containers.
+
+    GUIDELLM__REPORT_GENERATION__SOURCE overrides the URL guidellm uses to fetch its
+    HTML report template.  Without this, guidellm crashes post-benchmark with an httpx
+    HTTPStatusError on a 301 redirect from blog.vllm.ai.  We don't use guidellm's HTML
+    output — our TPM comes from vLLM's Prometheus counters — so a stub file is fine.
+
+    Returns True on success, False on guidellm non-zero exit.
+    """
+    guidellm_bin = "/opt/guidellm-venv/bin/guidellm"
```
zhuangqh: try call the lib instead. like
```python
    import asyncio

    from guidellm.benchmark.entrypoints import benchmark_generative_text
    from guidellm.benchmark import BenchmarkGenerativeTextArgs

    args = BenchmarkGenerativeTextArgs(
        target=VLLM_BASE_URL,
        data=[f"prompt_tokens={BENCHMARK_INPUT_LEN},output_tokens={BENCHMARK_OUTPUT_LEN}"],
        profile="throughput",
        rate=[float(rate)],
        max_seconds=BENCHMARK_DURATION,
        processor=processor or None,
        data_num_workers=0,
        random_seed=random.randint(0, 2**31 - 1),
        outputs=[],  # no report files needed
    )

    try:
        asyncio.run(benchmark_generative_text(args=args))
        return True
    except Exception as exc:
        _log(f"guidellm_failed: {exc}")
        return False
```

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,339 @@
...
+
+    elapsed = t1_epoch - t0_epoch
+    _log(
+        f"benchmark_window elapsed_sec={elapsed:.1f} "
+        f"delta_gen={delta_gen} delta_prompt={t1_prompt - t0_prompt}"
+    )
+    return round((delta_gen + (t1_prompt - t0_prompt)) * 60.0 / elapsed, 2)
+
+
+def _drain() -> None:
```
zhuangqh: I still suggest giving a timeout for it.
```
def _drain(timeout: float = 100.0) -> None:
    """Spin until vllm:num_requests_running reaches zero."""
    _log("drain_start")
    deadline = time.monotonic() + timeout
    while True:
        if _read_counter("vllm:num_requests_running") == 0:
            break
        if time.monotonic() >= deadline:
            raise TimeoutError(f"_drain timed out after {timeout}s")
        time.sleep(2)
```

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,339 @@
...
+        _drain()
+        t_drain_end = time.time()
+        _log(
+            f"drain_done elapsed={t_drain_end - t_bench_end:.1f}s "
+            f"total_phase_elapsed={t_drain_end - t_bench_start:.1f}s"
+        )
+    except Exception as exc:
+        _log(f"benchmark_failed error={exc}")
+
+    _write_to_pid1(f"KAITO_BENCHMARK_RESULT vllm_total_tpm={tpm}\n", fd=1)
```
zhuangqh: a structured output may be better for latter parse.

api/v1beta1/labels.go
```diff
@@ -55,6 +55,16 @@ const (
 
 	// AnnotationNodeImageFamily specifies node image family used by generated NodeClaim.
 	AnnotationNodeImageFamily = KAITOPrefix + "node-image-family"
+
+	// AnnotationRunBenchmark enables the post-load throughput benchmark stage.
+	// When set to "true" on a Workspace, the inference container runs a guidellm
+	// benchmark after the model loads before marking the container as ready.
+	AnnotationRunBenchmark = KAITOPrefix + "run-benchmark"
+
+	// AnnotationBenchmarkRate overrides the auto-computed guidellm concurrency rate.
```
Fei-Guo: benchmark-rate is a vague term. is this just the `max-concurrecy`?

pkg/workspace/inference/preset_inferences.go
```diff
@@ -301,6 +301,61 @@ func buildDistributedStartupProbe(timeout time.Duration, wObj *v1beta1.Workspace
...
+// buildBenchmarkStartupProbe returns an exec startup probe that runs
+// benchmark_entrypoint.py on every kubelet tick.
+//
+// While vLLM is loading, the script exits 1 (/health not yet up), consuming the
+// failureThreshold budget.  Once /health passes, the script runs the full benchmark
+// and drain phase, then exits 0 — which marks the startup probe as passed and
+// activates the readiness probe.
+//
+// timeoutSeconds is set to 300 to prevent kubelet killing the process mid-benchmark.
+func buildBenchmarkStartupProbe(timeout time.Duration) *corev1.Probe {
```
Fei-Guo: Can we merge  buildBenchmarkStartupProbe and buildDistributedBenchmarkStartupProbe? they look similar.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +475,63 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
...
+					if rate < 1 {
+						rate = 1
+					}
+				}
+			}
+			if rate > 0 {
+				benchmarkEnvVars = []corev1.EnvVar{
+					{Name: "BENCHMARK_RATE", Value: strconv.Itoa(rate)},
+				}
+			}
```
Fei-Guo: You should add a comment here to explain what the behavior is if rate is 0.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +475,63 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
...
+				if v, err := strconv.Atoi(rateStr); err == nil && v > 0 {
+					rate = v
+					klog.Infof("benchmark rate overridden by annotation: %d", rate)
+				} else {
+					klog.Warningf("ignoring invalid %s annotation value %q (must be a positive integer); falling back to computed rate", v1beta1.AnnotationBenchmarkRate, rateStr)
+				}
+			}
+			if rate == 0 && gpuConfig != nil && inferenceParam.TotalSafeTensorFileSize != "" && inferenceParam.BytesPerToken > 0 {
+				// Compute saturation concurrency
+				// availableVRAMGiB = (totalVRAM - modelWeights) × 0.9 safety factor for overhead.
```
Fei-Guo: The comment here should not repeat the algorithm. Instead, it should explain the rational or reasoning about this algorithm. And that is the effect if rate is not accurate.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +475,63 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
...
+			// vLLM is PID 1 — the benchmark runs from the exec startup probe instead.
+			// The readiness probe stays as defaultReadinessProbe (no sentinel file needed).
+			startupProbe = buildBenchmarkStartupProbe(readinessTimeout)
+
+			// Determine BENCHMARK_RATE: user annotation takes precedence; fall back to
+			// computing from available VRAM and per-token KV-cache size.
+			var rate int
+			if rateStr := ctx.Workspace.Annotations[v1beta1.AnnotationBenchmarkRate]; rateStr != "" {
+				if v, err := strconv.Atoi(rateStr); err == nil && v > 0 {
+					rate = v
```
Fei-Guo: I would suggest moving rate val valiation into the webhook, can be done in future change. But you should cap the rate to avoid user putting very high value. Btw, do you think anyone wants to controller the rate? can we remove it in the first version?

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+# import so that guidellm's newer huggingface_hub (which exports is_offline_mode
+# and other symbols) is loaded instead of the older system-installed version.
+_sp = glob.glob("/opt/guidellm-venv/lib/python*/site-packages")
+if _sp and _sp[0] not in sys.path:
+    sys.path.insert(0, _sp[0])
+
+from huggingface_hub import scan_cache_dir  # noqa: E402
+
+# ── Configuration ─────────────────────────────────────────────────────────────
+BENCHMARK_DURATION = 60
```
Fei-Guo, line 53: This is a good start. But I would expect our benchmark can take different query pattern. This is encode heavy one, but we can add decode heavy one, or mixed type.

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+
+    The controller calculates:
+        kv_per_req = BENCHMARK_INPUT_LEN * bytesPerToken / 1 GiB
+        rate       = ceil(availableVRAMGiB / kv_per_req)
+
+    Falls back to 512 if the env var is absent (unknown SKU / model).
+    """
+    val = os.environ.get("BENCHMARK_RATE", "")
+    if not val:
+        return 512
```
Fei-Guo: If you want to add a default, I'd suggest adding it in the controller code. Better readability.

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+    try:
+        return max(1, int(val))
+    except ValueError:
```
Fei-Guo, line 134: The rate value is supposed to be validated in the controller code already. If user inputs something invalid by directly changing the env value in statefulset spec, you should just fail the test.

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+    # Case 1: baked-in weights (tokenizer lives at the weights root)
+    if (weights / "config.json").exists():
+        return str(weights)
+
+    # Case 2: DAR or HF cache — ask huggingface_hub for the repo_id
+    try:
+        cache_info = scan_cache_dir(cache_dir=str(weights))
+        repos = sorted(cache_info.repos, key=lambda r: r.repo_id)
+        if repos:
+            return repos[0].repo_id
```
Fei-Guo, line 170: it is not very clear what is returned? what is repo_id? is it a path as well?

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+        target=VLLM_BASE_URL,
+        data=[
+            f"prompt_tokens={BENCHMARK_INPUT_LEN},output_tokens={BENCHMARK_OUTPUT_LEN}"
+        ],
+        profile="throughput",
+        rate=[float(rate)],
+        max_seconds=BENCHMARK_DURATION,
+        processor=processor or None,
+        data_num_workers=0,
+        random_seed=random.randint(0, 2**31 - 1),
```
Fei-Guo: We often use time as random seed (something like gettimeofday()).

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+        t_bench_end = time.time()
+        _log(f"benchmark_done elapsed={t_bench_end - t_bench_start:.1f}s")
+        _drain()
+        t_drain_end = time.time()
+        _log(
+            f"drain_done elapsed={t_drain_end - t_bench_end:.1f}s "
+            f"total_phase_elapsed={t_drain_end - t_bench_start:.1f}s"
+        )
+    except Exception as exc:
+        _log(f"benchmark_failed error={exc}")
```
Fei-Guo, line 352: Should we fail the probe if benchmarking fails? I think so. User should get a clear signal something is wrong, If this is a false alarm, user can add the annotation to bypass the entire benchmark phase.

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -0,0 +1,318 @@
...
+
+    _write_to_pid1(f'KAITO_BENCHMARK_RESULT {{"vllm_total_tpm":{tpm}}}\n', fd=1)
```
Fei-Guo: What if the _write_to_pid1 call fails?

Fei-Guo: And we only log tpm, not TTFT?

pkg/workspace/inference/preset_inferences.go
2170chm: It is removed now per our discussion. User won't be able to find out the correct value anyway.

presets/workspace/inference/vllm/benchmark_entrypoint.py
2170chm: Will defer to future versions.

presets/workspace/inference/vllm/benchmark_entrypoint.py
2170chm: > What if the _write_to_pid1 call fails?

In this case, it is very unrecoverable. The controller should treat a lack of KAITO_BENCHMARK_RESULT as if it is -1.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -301,6 +301,54 @@ func buildDistributedStartupProbe(timeout time.Duration, wObj *v1beta1.Workspace
...
+// failureThreshold budget.  Once /health passes, the script runs the full benchmark
+// and drain phase, then exits 0 — which marks the startup probe as passed and
+// activates the readiness probe.
+//
+// When wObj is non-nil the probe is built for distributed inference: a shell
+// conditional routes the leader (POD_INDEX=0) to benchmark_entrypoint.py and
+// workers to the standard multi-node health check.
+//
+// timeoutSeconds is set to 600 to prevent kubelet killing the process mid-benchmark.
+func buildBenchmarkStartupProbe(timeout time.Duration, wObj *v1beta1.Workspace) *corev1.Probe {
```
Fei-Guo: Using wobj == nil check seems to be hacky, we can add a new bool to switch the code.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +468,63 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
 			readinessTimeout = defaultStartupProbeTimeout
 		}
 
+		// When benchmarking is enabled, run the benchmark from the exec startup probe.
+		// vLLM remains PID 1; benchmark_entrypoint.py is called on each kubelet probe tick.
+		finalCommands := commands
```
zhuangqh: can you make this function an individual modifier

pkg/workspace/inference/preset_inferences.go
```diff
@@ -420,16 +468,63 @@ func GenerateInferencePodSpec(gpuConfig *sku.GPUConfig, numNodes int) func(*gene
...
+		readinessProbe := defaultReadinessProbe
+		startupProbe := buildStartupProbe(readinessTimeout)
+		var benchmarkEnvVars []corev1.EnvVar
+		if ctx.Workspace.Annotations[v1beta1.AnnotationRunBenchmark] == "true" {
```
zhuangqh: Can you move this check to https://github.com/kaito-project/kaito/blob/main/api/v1beta1/labels.go

func IsRunBenchmarkEnabled(ws *Workspace) bool {

pkg/workspace/inference/preset_inferences.go
zhuangqh: Yes. it's very tricky here.

zhuangqh:
do bump base image tag here to let the dockerfile change taking effect https://github.com/kaito-project/kaito/blob/main/presets/workspace/models/supported_models.yaml

PR #1864, opened 2026-03-20
feat: Writes TPM to workspace and inferenceset

### LINKED ISSUE #1849: Controller logic to write benchmarked workspace TPM to inferenceset.status.workspaceProfile
Part of #1775 (TPM-aware Autoscaling Policy), follow up to #1828. The autoscaler needs a reliable per-instance TPM value to enforce tokenPerMinuteLimit. Write the logged TPM result in the pod log to the CR's `inferenceset.status.workspaceProfile`.
Per #1828, the benchmarked TPM result is guaranteed to be in the kubectl log (last occurrence) of Pod 0 once inferenceReady is true. 
Update the controller to parse it and write to `inferenceset.status.workspaceProfile` if IsRunBenchmarkEnabled is true and inferenceReady is true.
Reflect that the result is written via a new condition `BenchmarkCompleted`

**Additional context**

2170chm:
Implemented controller to parse benchmark Tokens per Minute(TPM) results from pod log and write to WS and IS

**Reason for Change**:

This PR writes the loggged TPM to the workspace's status, and writes the sum of each workspace's TPM to Inferenceset's status. It helps user understand the throughput of the Model and can potentially be used for autoscaling.
**Requirements**

**Issue Fixed**:

Closes #1849 
**Notes for Reviewers**:

#### Design Decisions
1. Feature Flag - Adding `kaito.sh/run-benchmark: "true"` under InferenceSet.metadata.annotation enables the benchmark phase. It will also enable it for all child workspaces, each of which runs the benchmark test.

2. An inferenceset's TPM is the sum of all of its workspaces' TPMs. On each inferenceset reconcile loop, the sum is calculated and used to overwrite the current TPM. InferenceSet `BenchmarkCompleted` condition is always recomputed. Unlike the workspace-level guard, the InferenceSet condition is written from scratch on every reconcile by comparing benchmarkedReplicas == desired number of replicas. No guard needed because no I/O is involved.

3. Inferenceset WorkspaceProfile is cleared when no benchmark result exists. When no ready replica has a valid result, the InferenceSet's WorkspaceProfile is set to nil rather than left stale from a previous generation of workspaces. 

4. Workspace log parsing has an idempotency guard. It skips the log streaming and parsing on every reconcile once we have the condition `BenchmarkCompleted=True`, avoiding redundant log reads in steady state. Guard safety via explicit clearing in the not-ready path — when InferenceReady goes false, BenchmarkCompleted condition and BenchmarkResult are cleared immediately. This ensures the guard only holds when the pod has been continuously ready, and benchmark re-runs automatically after recovery.

5. Log parsing only reads the last 500 lines for efficiency. It reads the last occurrence of a valid result line and by the time the log is streamed, if there is a valid result line, it should be well within the last 500 lines.

6. When controller throws an error from stream or log parsing failures such as malformed json log, the benchmark script has successfully finished so the startup probe already passes. In this case, the pod will be marked ready, but the workspace will have a false `BenchmarkCompleted` condition with the "BenchmarkFailed" reason. The `WorkspaceSucceeded` condition will also be false. The workspace will still serve traffic, but controller prioritizes deleting it when scaling down, and the inferenceset TPM calculation will not include it. If no workspace has benchmark results, the inferenceset will also have a false `BenchmarkCompleted` condition.

Example results (GPUs: 2 x Standard_NC48ads_A100_v4)
Results with the latest API:
Latest

Terminal failures (replicas: 2, when controller log parsing logic fails. e.g. from streaming, malformed json log)
Failed_Inferenceset
Failed_ws_1
Failed_ws_2

Legacy results:
Llama-3-3-8b Inferenceset (replicas: 2):
Screenshot 2026-03-23 at 12 57 12 PM
Single Node Workspace:
Screenshot 2026-03-23 at 1 05 45 PM
Distributed Workspace:
Screenshot 2026-03-23 at 1 19 42 PM

Phi-4-mini Inferenceset (replicas: 2)
Phi-4-mini-inferenceset-TPM
Phi-4-mini-inferenceset-ws-TPM-1
Phi-4-mini-inferenceset-ws-TPM-2

Pod restarts (replicas: 2. When one of the pods restart and hasn't finished benchmarking)
After one of the two pods restarted


cmd/workspace/main.go
```diff
@@ -175,11 +176,18 @@ func main() {
 	k8sclient.SetGlobalClient(mgr.GetClient())
 	kClient := k8sclient.GetGlobalClient()
 
+	kubeClient, err := kubernetes.NewForConfig(cfg)
```
Fei-Guo, line 177: https://github.com/kaito-project/kaito/blob/6d94fc5551a71477372d601f689f176025744f50/pkg/k8sclient/client.go#L26

We have a global kubeclient, so this change is not needed.

api/v1alpha1/labels.go
```diff
@@ -102,3 +106,9 @@ func GetInferenceSetRuntimeName(iObj *InferenceSet) model.RuntimeName {
 
 	return runtime
 }
+
+// IsRunBenchmarkEnabled reports whether the InferenceSet has the benchmark
+// annotation set to "true".
+func IsRunBenchmarkEnabled(iObj *InferenceSet) bool {
```
Fei-Guo, line 112: If default behavior is to disable benchmark run, we need to add doc about how to enable it (can be future change).

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -806,14 +810,20 @@ func applyTuningWorkspaceStatus(status *kaitov1beta1.WorkspaceStatus, generation
...
+	generation := wObj.GetGeneration()
 	resourceReady := resourceConditionStatus == metav1.ConditionTrue
 	isInferenceEstablished := status.State == kaitov1beta1.WorkspaceStateReady || status.State == kaitov1beta1.WorkspaceStateNotReady
 
 	if inferenceReady && resourceReady {
 		setWorkspaceCondition(status, generation, appendMessage,
 			kaitov1beta1.WorkspaceConditionTypeInferenceStatus, metav1.ConditionTrue, "WorkspaceInferenceStatusSuccess", "Inference has been deployed successfully")
+
+		if kaitov1beta1.IsRunBenchmarkEnabled(wObj) {
+			applyBenchmarkStatus(ctx, kubeClient, status, wObj, generation, appendMessage)
```
Fei-Guo: This looks wrong. If applybenchmarkStatus has any error, we should not set WorkspaceStateReady.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -824,13 +834,46 @@ func applyInferenceWorkspaceStatus(status *kaitov1beta1.WorkspaceStatus, generat
...
+// clears the condition on any pod restart or rolling update, so this guard only works in
+// steady state when the pod is stable and the result is already recorded.
+func applyBenchmarkStatus(ctx context.Context, kubeClient kubernetes.Interface, status *kaitov1beta1.WorkspaceStatus, wObj *kaitov1beta1.Workspace, generation int64, appendMessage func(string) string) {
+	// Skip once the benchmark is done. Safe because the not-ready path clears BenchmarkCompleted
+	// whenever inference goes down, so we won't get into a stale state.
+	if c := meta.FindStatusCondition(status.Conditions, string(kaitov1beta1.WorkspaceConditionTypeBenchmarkCompleted)); c != nil && c.Status == metav1.ConditionTrue {
+		return
+	}
+
+	result, err := reconcileBenchmarkResult(ctx, wObj, &kubeClientStreamer{kubeClient: kubeClient})
```
Fei-Guo: If the benchmark fails, are we going to flood controller with all error msg?

Fei-Guo (commented):
Make sure to check the controller logs when benchmark run fails (you can inject error manually). 


`The aggregateBenchmarkResults function sums TokensPerMinute values and increments benchmarkedReplicas for any workspace where parsing succeeds. If a benchmark fails and emits a TPM of -1 (as per the Python script), it will be included in the sum and count towards the BenchmarkCompleted condition. This could result in an incorrect total TPM and a misleading "completed" status if any replica failed its benchmark. Validation should ensure TPM > 0 before aggregating.`  

And check the above comment from PR agent.

cmd/workspace/main.go
2170chm: controller-runtime client can't access a pod's log. I need a typed client here from client-go. I will make it global as well.

api/v1alpha1/labels.go
2170chm: Sure, I will add that in the future. This is just a feature flag that can be reverted easily after we are confident it works in production for a while. If you believe we should make it default to enable benchmark, I can change that in the future as well.

api/v1beta1/workspace_types.go
```diff
@@ -210,6 +244,11 @@ type WorkspaceStatus struct {
 	// TargetNodeCount is used for recording the desired number of gpu nodes that needed for the workspace.
 	// This field remains immutable after being set by NodesEstimator.
 	TargetNodeCount int32 `json:"targetNodeCount,omitempty"`
+
+	// BenchmarkResult holds the metrics from the post-load inference benchmark.
+	// Only populated when the kaito.sh/run-benchmark annotation is "true".
+	// +optional
+	BenchmarkResult *BenchmarkResult `json:"benchmarkResult,omitempty"`
```
Fei-Guo: BenchmarkResult -> Performance.

api/v1beta1/workspace_types.go
```diff
@@ -193,6 +193,40 @@ const (
...
+	// InputTokens is the number of prompt tokens per request.
+	InputTokens int32 `json:"inputTokens"`
+	// OutputTokens is the number of generated tokens per request.
+	OutputTokens int32 `json:"outputTokens"`
+	// MaxConcurrency is the maximum number of concurrent requests used during the run.
+	MaxConcurrency int32 `json:"maxConcurrency"`
+}
+
+// BenchmarkMetric holds a single benchmark measurement along with its description and config.
+type BenchmarkMetric struct {
```
Fei-Guo: BenchmarkMetric -> metric

api/v1beta1/workspace_types.go
```diff
@@ -193,6 +193,40 @@ const (
 	WorkspaceStateFailed    WorkspaceState = "Failed"
 )
 
+// BenchmarkConfig holds the parameters used for the benchmark run.
+type BenchmarkConfig struct {
```
Fei-Guo: this needs to be kv map.

api/v1alpha1/inferenceset_types.go
```diff
@@ -69,6 +91,9 @@ type InferenceSetStatus struct {
 	// Conditions report the current conditions of the InferenceSet.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// WorkspaceProfile holds aggregated performance characteristics across all workspace replicas.
+	// +optional
+	WorkspaceProfile *WorkspaceProfile `json:"workspaceProfile,omitempty"`
```
Fei-Guo: WorkspaceProfile-> Performance.

api/v1alpha1/inferenceset_types.go
```diff
@@ -57,6 +57,28 @@ type InferenceSetSpec struct {
 	UpdateStrategy appsv1.StatefulSetUpdateStrategy `json:"updateStrategy,omitempty"`
 }
 
+// WorkspaceMetric holds an aggregated benchmark measurement across workspace replicas.
```
Fei-Guo: We should reuse the struct defined in workspace.

pkg/k8sclient/client.go
```diff
@@ -26,3 +27,13 @@ func SetGlobalClient(c client.Client) {
...
 	return Client
 }
+
+var KubeClient kubernetes.Interface
+
+func SetGlobalKubeClient(c kubernetes.Interface) {
+	KubeClient = c
+}
+
+func GetGlobalKubeClient() kubernetes.Interface {
```
Fei-Guo: GetGlobalKubeClient->GetGlobalClientGoClient

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -824,13 +839,58 @@ func applyInferenceWorkspaceStatus(status *kaitov1beta1.WorkspaceStatus, generat
...
+	if err != nil {
+		// Log at verbose level only — this fires on every reconcile until the
+		// benchmark pod emits its result line, so ErrorS would flood the logs.
+		// The BenchmarkCompleted condition carries the reason for visibility.
+		klog.V(4).InfoS("benchmark result not yet available", "workspace", klog.KObj(wObj), "err", err)
+		setWorkspaceCondition(status, generation, appendMessage,
+			kaitov1beta1.WorkspaceConditionTypeBenchmarkCompleted, metav1.ConditionFalse,
+			"BenchmarkResultUnavailable", err.Error())
+		return false
```
Fei-Guo: This needs to be a clear error log with return `true` since retry would not fix the error.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -806,14 +808,27 @@ func applyTuningWorkspaceStatus(status *kaitov1beta1.WorkspaceStatus, generation
 	}
 }
 
-func applyInferenceWorkspaceStatus(status *kaitov1beta1.WorkspaceStatus, generation int64, appendMessage func(string) string,
+func applyInferenceWorkspaceStatus(ctx context.Context, kubeClient kubernetes.Interface, status *kaitov1beta1.WorkspaceStatus, wObj *kaitov1beta1.Workspace, appendMessage func(string) string,
```
Fei-Guo: Client is a global variable now, you don't need to change method inputs.

api/v1alpha1/inferenceset_types.go
2170chm: Config only applies to workspace's metric

PR #1870, opened 2026-03-24
feat: add GPU spot instance support for Azure node pools #1837

### LINKED ISSUE #1837: Add support for GPU Spot instances

GPU is very expensive with limited quota.

Running models on Spot instances GPU in order to reduce the cost for both dev and production environments.

KQITO doesn't support natively yet Spot instances. But it could work by just adding the required tolerations related to Spot Nodepool.

**Additional context**

What should be done to enable this feature is to add the following Toleration to:

1. workspace StatefulSet: https://github.com/kaito-project/kaito/blob/main/charts/kaito/workspace/templates/kaito.sh_workspaces.yaml
 
2. nvidia-device-plugin-daemonset DaemonSet : https://github.com/kaito-project/kaito/blob/1e723e307fc390ec8dd42bad55b815a59f2f4019/charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml#L38
 
        - key: kubernetes.azure.com/scalesetpriority
          operator: Equal
          value: spot
          effect: NoSchedule

zeel2104:
Reason for Change:
GPU instances are expensive and have limited quota. This PR adds support for running KAITO workloads on Azure Spot GPU node pools to reduce costs in both dev and production environments.
Requirements

[X] added unit tests and e2e tests (if applicable).

Issue Fixed:

Fixes #1837

Notes for Reviewers:

-> Added a spotInstance.enabled flag to values.yaml (defaults to false, no breaking change for existing users)
-> When enabled, adds the kubernetes.azure.com/scalesetpriority=spot:NoSchedule toleration to the nvidia-device-plugin DaemonSet so it can be scheduled on Spot node pools
-> The workspace Deployment already supports tolerations dynamically via Values.tolerations so no changes needed there
-< To enable spot support, users simply set spotInstance.enabled: true in their values override
-> No existing Helm tests found in the chart. Validated via helm lint and helm template --set spotInstance.enabled=true dry-ruN


charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
```diff
@@ -47,6 +47,12 @@ spec:
           operator: "Equal"
           value: "gpu"
           effect: "NoSchedule"
+        {{- if .Values.spotInstance.enabled }}
```
zhuangqh: I think it doesnt hurt to add this toleration without toggle?

Fei-Guo:
If you check the issue, we would also need to add toleration in the sts pod template otherwise, the pod cannot use the spot gpu.

Fei-Guo:
@zeel2104, the toleration we add is specific for azure. Please add a guard in both yaml and controller to add the toleration for azure cloud provider only.

charts/kaito/workspace/templates/nvidia-device-plugin-ds.yaml
HoussemDellai: Agree at the end customer could use node labels/selectors. And Spot instances are legitimate target for running AI models.

HoussemDellai:
> @zeel2104, the toleration we add is specific for azure. Please add a guard in both yaml and controller to add the toleration for azure cloud provider only.

Oh good point. KAITO is meant to be multi-provider and the Spot tolerations might be different across different providers. At the end may be it is good to add an input variable in values.yaml for 'custom'/provider-specific tolerations.

zeel2104:
@Fei-Guo 
Updated the PR to guard the Azure spot toleration by cloud provider in both the Helm chart and controller-generated workloads.

Validation:
- go test ./pkg/workspace/inference ./pkg/workspace/tuning
- helm template checks:
  - cloudProviderName=azure, spotInstance.enabled=true -> toleration present
  - cloudProviderName=aws, spotInstance.enabled=true -> toleration absent
  - cloudProviderName=azure, spotInstance.enabled=false -> toleration absent
Let me know if any modifications need to be done

charts/kaito/workspace/values.yaml
```diff
@@ -47,6 +47,13 @@ nvidiaDevicePlugin:
   daemonsetName: "nvidia-device-plugin-daemonset"
   image: "mcr.microsoft.com/oss/v2/nvidia/k8s-device-plugin:v0.18.2-1"
   imagePullPolicy: "IfNotPresent"
+
+# Spot instance support for Azure GPU node pools.
+# Set enabled: true to allow workloads to be scheduled on Azure Spot nodes.
+# This adds the required toleration for Azure Spot node pools.
+spotInstance:
+  enabled: false
```
andyzhangx, line 55: just set default as true since it would be easier to covered in e2e test and no harm if it's supporting spot by default?

zhuangqh: agree

Issue #1896, opened 2026-03-30
Add curated model catalog in Kaito


Fei-Guo:
**Is your feature request related to a problem? Please describe.**
Currently, Kaito will retrieve the model meta data from HF directly, which requires the http calls during model validation. Kaito also maintains two sources for model weights, MCR and hf.

It would be more efficient to maintain a model catalog list to replace the existing supportedmodel.yaml and unify the model source to be hf. 

The model catalog will be also very useful for UI to list model choices for the users. 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Abhishek9639:
Greetings @Fei-Guo, 
I’d like to work on this issue. It looks interesting and aligns with my experience. 
Could you please assign it to me?
Thanks

zhehli688:
4/6/26 update: initial implementation was merged: https://github.com/kaito-project/kaito/pull/1901. Migration of preset models in progress: https://github.com/kaito-project/kaito/pull/1920.

zhehli688:
4/13/26 update: completed onboarding of phi-4 models in #1920. Onboarding of other preset models in progress: #1945.

PR #1897, opened 2026-03-31
feat: prototype for builtin model catalog

zhehli688:
**Reason for Change**:

This is the first step to resolve issue https://github.com/kaito-project/kaito/issues/1896. I created a new model catalog in presets/workspace/models/builtin_models.yaml and migrated the two existing phi-4 models. For models in builtin_models.yaml, we enforce the following behaviors:

- Load model weights from HF instead of MCR (base image is still loaded from MCR).
- Load model metadata from builtin_models.yaml instead of Go files like presets/workspace/models/phi4/model.go (Some metadata are still fetched from HF and we will hardcode them in builtin_models.yaml in another PR).

**Requirements**

- [x] verified deployment of two existing phi-4 models in an AKS cluster

**Issue Fixed**:

**Notes for Reviewers**:


presets/workspace/models/builtin_models.go
```diff
@@ -0,0 +1,225 @@
...
+var builtinModelsYAML []byte
+
+// BuiltinCatalog holds the list of builtin models parsed from builtin_models.yaml.
+type BuiltinCatalog struct {
+	Models []BuiltinModelSpec `yaml:"models"`
+}
+
+// BuiltinModelSpec defines all metadata for a builtin preset model,
+// fully deserializable from YAML. This replaces per-model Go registration files.
+type BuiltinModelSpec struct {
```
zhuangqh, line 37: can we reuse existing datastructure? seems identical

zhehli688: Assuming you mean reusing PresetParam. It will require non-trivial refactorization because PresetParam will need to support two modes:
1. For the new path, unmarshal builtin_models.yaml directly into PresetParam
2. For the existing path, get metadata from supported_models.yaml and other parameteres from Go files.

The refactorization is also risky because it will impact all existing Preset models. My current plan is to leave PresetParam intact for now. Once we migrate all Preset models to use builtin_models.yaml, we can consolidate PresetParam and BuiltinModelSpec (we should be able to only keep one of them). Let me know if this plan doesn't make sense or you have a simpler solution to reuse PresetParam in mind.

zhehli688:
Discussed with Fei offline. Will take a difference approach instead.

PR #1917, opened 2026-04-04
feat: implemented preset custom pvc for model weights

### LINKED ISSUE #1850: [Feature] Support user-provided PVC for model weights in `inference.preset` mode
### Problem

In `inference.preset` mode, the `model-weights-volume` is always either an EmptyDir or an auto-provisioned NVMe-backed PVC. Users cannot supply their own PersistentVolumeClaim — for example, one pre-populated with model weights from a shared storage system, or backed by a durable storage class so that weights survive pod restarts without re-downloading.

The only workaround today is to switch to `inference.template`, which requires defining the entire pod spec manually and loses all preset benefits: automatic vLLM/Transformers configuration, GPU resource assignment, model download init containers, distributed inference setup, probes, and tolerations.

### Current Behavior

| Volume | Type | Mount Path | Purpose |
|--------|------|------------|---------|
| `model-weights-volume` | EmptyDir *or* NVMe PVC (VolumeClaimTemplate) | `/workspace/weights` | Model weights |

- **EmptyDir path**: `SetDefaultModelWeightsVolume` appends `utils.DefaultModelWeightsVolume` (an `EmptyDir{}`) to `spec.Volumes`.
- **NVMe path**: `GenerateModelWeightsCacheVolume` creates a `PersistentVolumeClaim` with `StorageClassName: localNVMe` and adds it as a StatefulSet VolumeClaimTemplate.
- In both cases the volume is named `model-weights-volume` and mounted at `/workspace/weights` via `utils.DefaultModelWeightsVolumeMount`.
- Model data is populated either by an ORAS init container (`model-weights-downloader`) or at runtime by vLLM (`--download-dir /workspace/weights`).

There is no mechanism for users to provide their own volume source for this mount.

### Use Case

A user has a PVC pre-populated with model weights (e.g., via a batch job, a shared NFS volume, or a CSI snapshot). They want to use it with `inference.preset` without re-downloading the model and without losing preset benefits.

Desired YAML:

```yaml
apiVersion: kaito.sh/v1beta1
kind: Workspace
metadata:
  name: workspace-llama3-8b
resource:
  labelSelector:
    matchLabels:
      apps: llama3-8b
  count: 1
inference:
  preset:
    name: llama-3-8b-instruct
    presetOptions:
      volumeSource:
        persistentVolumeClaim:
          claimName: my-model-weights-pvc
```

With this configuration the controller uses the user-provided PVC as the backing for `model-weights-volume` instead of creating an EmptyDir or NVMe VolumeClaimTemplate. The mount path stays `/workspace/weights`. Everything else — runtime flags, probes, GPU resources, distributed inference setup — continues to be managed by the preset.

Today I have to do something like this:

```yaml
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-model-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Gi                      # Qwen3-0.6B is small, 10Gi is plenty
  # storageClassName: <your-storage-class>  # Optional, uses default if omitted
---
apiVersion: kaito.sh/v1beta1
kind: Workspace
metadata:
  name: workspace-qwen3-0-6b
resource:
  labelSelector:
    matchLabels:
      apps: qwen3-0.6b
  count: 1          # Number of nodes to use
inference:
  template:         # Template mode = you define the pod spec yourself
    spec:
      containers:
      - name: inference-server
        image: vllm/vllm-openai:latest    # or any vLLM image
        command: ["python3", "-m", "vllm.entrypoints.openai.api_server"]
        args:
        - "--model"
        - "Qwen/Qwen3-0.6B"
        - "--host"
        - "0.0.0.0"
        - "--port"
        - "8000"
        ports:
        - containerPort: 8000
        resources:
          requests:
            nvidia.com/gpu: "1"
          limits:
            nvidia.com/gpu: "1"
        volumeMounts:
        - name: model-cache
          mountPath: /root/.cache/huggingface
        - name: dshm
          mountPath: /dev/shm
      volumes:
      - name: model-cache
        persistentVolumeClaim:
          claimName: my-model-pvc        # Your pre-created PVC
      - name: dshm
        emptyDir:
          medium: Memory                 # Shared memory for GPU IPC
```

2170chm:
**Reason for Change**:

This PR enables kaito to support ModelWeightsPVC and ModelWeightsSubPath under PresetOptions, which allows user to specify an existing pvc and optionally a subpath that contains the model weights (or if not, a deployment downloads them automatically in the pvc at the subpath or root if subpath is not provided). It allows users to reuse their model weights without downloading across restarts and it also automatically downloads missing files if the user's provided path does not contain all the necessary files.
**Requirements**

**Issue Fixed**:

Closes #1850 
**Notes for Reviewers**:
1. Download behavior with BYO PVC differs by model type:
DownloadAtRuntime models (e.g., llama-3.1-8b-instruct): vLLM uses HuggingFace Hub's download logic, which checks file hashes and skips already-existing files. This is where BYO PVC provides the most value. After a first download or if the weights are provided by the user, subsequent workspace creations skip the download entirely.
ORAS-based models (e.g., phi-4-mini-instruct): The init container pulls the full OCI artifact from the registry and overwrites existing files every time. BYO PVC still provides persistence, but does not skip the download step.

Happy path:
llama-3-1-8b first time download
llama-3-1-8b first time download
llama-3-1-8b reuse model weights in pvc
llama-3-1-8b reuse model weights

phi-4-mini custom pvcwith subpath
phi-4-mini-custom-pvc-w-subpath
phi-4-mini custom pvc w/o subpath (root)
phi-4-mini-custom-pvc-wo-subpath
models downloaded in pvc after both deployments
models downloaded after both deployments

Edge cases validated by webhook:
webhook rejects tuning workspace
webhook rejects tuning workspace
webhook rejects undersized pvc
webhook rejects undersized pvc
webhook rejects non-existent-PVC
webhook rejects non-existent-PVC
webhook rejects subpath without pvc
webhook rejects subpath without pvc


api/v1beta1/workspace_types.go
```diff
@@ -75,6 +75,17 @@ type PresetOptions struct {
 	// ModelAccessSecret is the name of the secret that contains the huggingface access token.
 	// +optional
 	ModelAccessSecret string `json:"modelAccessSecret,omitempty"`
+	// ModelWeightsPVC is the name of a pre-existing PVC in the same namespace
+	// to use as the model weights volume. When set, the controller uses this PVC
+	// instead of creating an EmptyDir or NVMe VolumeClaimTemplate.
+	// Only valid for inference workspaces; rejected for tuning workspaces.
+	// +optional
+	ModelWeightsPVC string `json:"modelWeightsPVC,omitempty"`
```
zhuangqh, line 83: use `ModelFilesVolume *v1.VolumeSource` for a more generic API? like what we do at adapterSpec https://github.com/kaito-project/kaito/blob/main/api/v1beta1/workspace_types.go#L129

pvc will be
```
corev1.VolumeSource{
	PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
		ClaimName: pvc.Name,
	},
},
```

PR #1919, opened 2026-04-06
fix: support non-streaming /v1/chat/completions in text-generation server

### LINKED ISSUE #1885: Support non-streaming responses in /v1/chat/completions
The HF tfs runtime (kaito-base) always returns SSE regardless of the stream field in the request body. When a client sends "stream": false, it still receives a chunked data: {...} SSE stream. This breaks standard OpenAI-compatible clients that expect a single JSON response object when stream=false.

Root cause: transformers[serving] v4.57.x — the version currently pinned in presets/workspace/dependencies/requirements.txt — does not handle the stream parameter in generate_chat_completion(). It unconditionally returns a Generator[str, None, None].

Bump transformers[serving] to 5.x.x. The new versions gate on req.get("stream") and returns a JSONResponse for non-streaming requests, making the behavior spec-compliant. Also, in [inference_api.py chat_completion](https://github.com/2170chm/kaito/blob/ea90bf84db9923dac16d5a28c16f2f2b9da77356/presets/workspace/inference/text-generation/inference_api.py#L342-L348), instead of wrapping `output` in StreamingResponse, directly return `output`.

Implementing the buffering logic ourselves in inference_api.py (wrapper around the generator). This is viable but duplicates logic that upstream will have on main.

**Additional context**
The fix already exists on the HF transformers main branch in src/transformers/cli/serve.py but has not been released as of v4.57.6.
Current pin: transformers[serving] >= 4.57.1, < 5 (presets/workspace/dependencies/requirements.txt).
vLLM is not affected — it has its own OAI-compatible server that correctly handles both streaming and non-streaming.
Tracking issue: once a new transformers release ships with the fix, bump the lower bound in requirements.txt and validate both stream=true and stream=false against /v1/chat/completions.

officialasishkumar:
**Reason for Change**:
The text-generation wrapper always returned SSE for `/v1/chat/completions`, even when clients sent `"stream": false`.

This change keeps streaming behavior unchanged and aggregates streamed chat-completion chunks into a single OpenAI-compatible JSON response for non-streaming requests. It also preserves streamed tool-call deltas and the served-model alias in the final response.

**Requirements**

**Issue Fixed**:
Fixes #1885

**Notes for Reviewers**:
1. This stays on the current `transformers` 4.x integration path instead of attempting a broader 5.x migration; the wrapper now handles `stream=false` locally and also passes through upstream JSON responses if they become available later.
2. Tests cover plain non-streaming responses, streamed tool-call delta aggregation, and served-model alias preservation.

### DIFF
--- a/presets/workspace/inference/text-generation/inference_api.py
+++ b/presets/workspace/inference/text-generation/inference_api.py
@@ -339,13 +339,142 @@ def _resolve_model_name(body: dict) -> dict:
     return body
 
 
+def _maybe_return_json_response(output: Any) -> JSONResponse | None:
+    """Pass through upstream non-streaming responses when available."""
+    if isinstance(output, JSONResponse):
+        return output
+    if isinstance(output, dict):
+        return JSONResponse(output)
+    if hasattr(output, "model_dump"):
+        return JSONResponse(output.model_dump(exclude_none=True))
+    if hasattr(output, "dict"):
+        return JSONResponse(output.dict(exclude_none=True))
+    return None
+
+
+def _iter_sse_payloads(output: Any):
+    """Yield JSON payloads from the transformers SSE generator."""
+    for raw_chunk in output:
+        if raw_chunk is None:
+            continue
+        if isinstance(raw_chunk, bytes):
+            raw_chunk = raw_chunk.decode()
+
+        for line in str(raw_chunk).splitlines():
+            if not line.startswith("data:"):
+                continue
+
+            payload = line.removeprefix("data:").strip()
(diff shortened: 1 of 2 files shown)


Fei-Guo:
@officialasishkumar thanks for working on this! I'd rather wait for new hf release though for this problem. We want to keep the wrapper layer thin in the base image.

officialasishkumar:
@Fei-Guo understood. I am not planning further wrapper-layer changes here and will leave this branch as-is pending the upstream Hugging Face release path.

PR #1962, opened 2026-04-15
feat: add non-streaming output guardrails hook for chat completions

xiaoqi-7:
**Summary**
This PR adds a minimal non-streaming output guardrails hook to the RAGEngine `/v1/chat/completions` path.

It introduces a small OutputGuardrails post-processing step for chat completion responses, adds the minimum configuration required to enable output scanning, supports basic redact and block behavior, and includes focused API tests.

**Scope**

Included:

- non-streaming output guardrails hook
- regex and banned-substring scanners
- redact / block behavior
- focused API tests
- scoped docs for the initial hook and its current limitations

Not included:

- streaming guardrails
- fail-open behavior
- structured auditing
- remote audit sinks
- metrics and monitoring

**Validation**
Validated with targeted API tests:
5 passed, 19 deselected


presets/ragengine/config.py
```diff
@@ -69,6 +69,27 @@
...
 # LLM_RESPONSE_FIELD = os.getenv("LLM_RESPONSE_FIELD", "result")  # Uncomment if needed in the future
 
+
+def _parse_csv_env(name: str) -> tuple[str, ...]:
+    raw_value = os.getenv(name, "")
+    return tuple(item.strip() for item in raw_value.split(",") if item.strip())
+
+
+OUTPUT_GUARDRAILS_ENABLED = (
+    os.getenv("OUTPUT_GUARDRAILS_ENABLED", "false").lower() == "true"
```
andyzhangx, line 79: so this is enabled by env var? how can end user enable this feature in KAITO? there should be also doc change for this feature.

presets/ragengine/guardrails/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .output_guardrails import OutputGuardrails
```
andyzhangx, line 14: for new python file, you need to add license header

Fei-Guo (approved):

PR #1968, opened 2026-04-16
feat: add karpenter crds dependencies and ensure aksnodeclass resources

### LINKED ISSUE #1950: Add Karpenter CRD dependencies, feature gate, and global AKSNodeClass bootstrap
## Description

Add Karpenter CRD dependencies to the KAITO controller, introduce a feature gate to enable Karpenter as the node provisioner, and bootstrap global AKSNodeClass resources at controller startup.

The global AKSNodeClass resources are created by the KAITO controller at startup (not via Helm chart), because the `AKSNodeClass` CRD is installed by Azure Karpenter and we cannot guarantee the installation order between KAITO and Azure Karpenter.

**Ref:** [New CRD Dependencies](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#new-crd-dependencies) | [Global AKSNodeClass](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#global-aksnodeclass) | [AKSNodeClass Configuration](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#aksnodeclass-configuration)

## Tasks

- [ ] Add Karpenter and Azure Karpenter Go module dependencies
- [ ] Generate or vendor client-go clients, informers, and listers for NodePool, NodeClaim, AKSNodeClass CRDs
- [ ] Register the new CRD schemes in the controller's scheme
- [ ] Add RBAC rules for NodePool, NodeClaim, AKSNodeClass (create, get, list, watch, update, patch, delete)
- [ ] Add feature gate `KarpenterProvisioner` (default `false`) to enable Karpenter as the node provisioner
- [ ] When feature gate is `true`, at controller startup:
  - Verify `AKSNodeClass` CRD exists in the cluster (i.e., Azure Karpenter is installed)
  - If CRD does not exist, fail startup with a clear error message
  - Create global AKSNodeClass resources (`image-family-ubuntu`, `image-family-azure-linux`) with `osDiskSizeGB: 300` and `karpenter.kaito.sh/managed-by: kaito` label (create-if-not-exists)
- [ ] When feature gate is `false`, skip all Karpenter-related initialization
- [ ] Add unit tests for feature gate logic, CRD existence check, and AKSNodeClass bootstrap

rambohe-ch:
**Reason for Change**:

1. Upgrade Karpenter dependencies: sigs.k8s.io/karpenter v1.6.2→v1.10.0, azure-karpenter v1.6.3→v1.10.2, aws-karpenter v1.6.3→v1.10.0, k8s.io/* v0.34→v0.35.  

2. Upgrade golang from 1.25 to 1.26.2, because karpenter v1.10.0 depends on it.

3. Add Start(ctx) error to NodeProvisioner interface for provisioner initialization at controller startup

4. Implement AzureKarpenterProvisioner: verifies Karpenter CRDs (NodePool, NodeClaim, AKSNodeClass), bootstraps global AKSNodeClasses, and runs a background goroutine to recreate them if accidentally deleted

5. Update factory to support ensureNodeClass feature gate with directClient for pre-cache CRD verification

6. Remove NodeManager.EnsureNodesReady — readiness checks are now encapsulated within each NodeProvisioner implementation

**Requirements**

**Issue Fixed**:

Fixes #1950 

**Notes for Reviewers**:


rambohe-ch:
@Fei-Guo  @zhuangqh  @2170chm PTAL

cmd/workspace/main.go
```diff
@@ -182,10 +185,22 @@ func main() {
 	}
 	k8sclient.SetGlobalClientGoClient(kubeClient)
 
-	// Create the node provisioner based on feature gates.
+	// Create a direct (non-cached) client for provisioner initialization,
```
Fei-Guo: There needs a clearer reasoning for using the direct client:
1) It is only used for query CRD existence for now;
2) The overhead is low and there is no need to cache entire CRD objects in controller; 

If the above are not the reasons, please explain. It is not clear why we cannot wait until Manager's cache is ready.

2170chm:
I think the Start() method needs RBAC.
apiextensions.k8s.io/customresourcedefinitions — get/list/watch — needed by verifyRequiredCRDs()
karpenter.sh/nodepools CRUD. Currently only nodeclaims is listed
The karpenter.azure.com/aksnodeclasses rule is gated on cloudProviderName "azure" but this provisioner uses "azure-karpenter", so the condition never matches

cmd/workspace/main.go
rambohe-ch: The reason for direct client is: we should prepare the global nodeclasses before manager(client.Client) started.

rambohe-ch:
> I think the Start() method needs RBAC. apiextensions.k8s.io/customresourcedefinitions — get/list/watch — needed by verifyRequiredCRDs() karpenter.sh/nodepools CRUD. Currently only nodeclaims is listed The karpenter.azure.com/aksnodeclasses rule is gated on cloudProviderName "azure" but this provisioner uses "azure-karpenter", so the condition never matches

RBAC has been updated, and the current status of node provisioner selection is verycomplicated:
 for example:
 - feature gate: FeatureFlagDisableNodeAutoProvisioning to select BYO
 - feature gate: FeatureFlagEnsureNodeClass + env [CLOUD_PROVIDER](https://github.com/kaito-project/kaito/blob/a60bcf1db373dbcaf9ffb913d07a8f8ad9dff41f/pkg/utils/nodeclaim/nodeclaim.go#L329) = "azure" to select azure karpenter.
 - feature gate: FeatureFlagEnsureNodeClass + env [CLOUD_PROVIDER](https://github.com/kaito-project/kaito/blob/a60bcf1db373dbcaf9ffb913d07a8f8ad9dff41f/pkg/utils/nodeclaim/nodeclaim.go#L329) = "aws" to select aws karpenter.
 - otherwise, select gpu-provisioner.

@Fei-Guo  @2170chm  I'd like to add a new startup parameter for configuring the specified node provisoner to replace current feature gates and envs. and I'll keep the consistency with current selection methods instead of replacing them directly.

cmd/workspace/main.go
rambohe-ch: I've updated the comments for the direct client.

pkg/nodeprovision/manager/factory.go
```diff
@@ -28,9 +29,14 @@ import (
 
 // NewNodeProvisioner creates and returns a NodeProvisioner based on feature gates.
 //
+//   - ensureNodeClass enabled: AzureKarpenterProvisioner (uses directClient for
+//     CRD verification and global AKSNodeClass bootstrap at Start time).
 //   - NAP disabled (BYO mode): BYOProvisioner (all provisioning ops are no-ops).
 //   - Default (NAP enabled): AzureGPUProvisioner (creates/deletes NodeClaims).
-func NewNodeProvisioner(kClient client.Client, recorder record.EventRecorder, defaultNodeImageFamily string) nodeprovision.NodeProvisioner {
+func NewNodeProvisioner(kClient, directClient client.Client, recorder record.EventRecorder, defaultNodeImageFamily string) nodeprovision.NodeProvisioner {
+	if featuregates.FeatureGates[consts.FeatureFlagEnsureNodeClass] {
```
zhuangqh, line 37: if FeatureFlagEnsureNodeClass, use karpenter? This is so weird. 

This flag only guard whether you check node class only I think.

rambohe-ch: As the response in https://github.com/kaito-project/kaito/pull/1968#issuecomment-4264085451, I will add a new startup parameter for choosing node provisioner.

zhuangqh (approved):

Fei-Guo (approved):

PR #1975, opened 2026-04-20
feat: Migrate default EPP from GWIE to llm-d inference scheduler

andyzhangx:
Replace the default Endpoint Picker (EPP) image from the upstream Gateway API Inference Extension EPP to llm-d-inference-scheduler (`mcr.microsoft.com/oss/v2/llm-d/llm-d-inference-scheduler:v0.7.1`).

The **InferencePool chart remains** from GWIE (oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool), **only the EPP container image is overridden** to use llm-d. This follows the GIE to llm-d migration plan where EPP implementation and plugins are consolidated into llm-d-inference-scheduler.

llm-d extends the GWIE EPP with advanced scheduling plugins including KV cache-aware routing, P/D disaggregation, and pluggable filters/scorers.

Changes:
- Replace GatewayAPIInferenceExtensionImageRepository with EPPImageHub, EPPImageName, EPPImageTag consts to properly set hub/name/tag for the InferencePool helm chart image override
- Update HelmRelease values to pass all three image fields
- Update tests accordingly
- Update gateway-api-inference-extension.md docs to reflect llm-d EPP

Ref: https://github.com/llm-d/llm-d-inference-scheduler

**Reason for Change**:
Migrate default EPP from GWIE to llm-d inference scheduler for advanced scheduling capabilities (KV cache-aware routing, P/D disaggregation). Uses MCR-hosted image for production reliability.

**Requirements**

- [ ] added unit tests and e2e tests (if applicable).

PR #1979, opened 2026-04-21
docs: add design proposal for migrating EPP to llm-d inference scheduler

andyzhangx:
## What type of PR is this?
/kind documentation

## What this PR does / why we need it:

Adds a design proposal for migrating KAITO's default Endpoint Picker (EPP) from the upstream Gateway API Inference Extension (GWIE) EPP to the [llm-d inference scheduler](https://github.com/llm-d/llm-d-inference-scheduler).

**I have added optional EPPPluginsConfig in InferenceSetSpec for "Advanced: Custom EPP Plugin Configuration", this is the only API change during migration. And by default, we don't even need to set this EPPPluginsConfig , this is for better scheduling with advanced features.**

### Key points:
- **Drop-in replacement**: llm-d uses the same `EndpointPickerConfig` API — no breaking changes
- **Zero config**: Default plugins (queue-scorer, kv-cache-utilization-scorer, prefix-cache-scorer) work out of the box with KAITO's vLLM metrics
- **Advanced plugins**: Enables access to llm-d-specific plugins (precise prefix cache, P/D disaggregation, label-based filtering)
- **Upstream alignment**: Follows the [GIE to llm-d migration plan](https://github.com/kubernetes-sigs/gateway-api-inference-extension/issues/2430)

### Related PRs:
- Implementation: #1975
- MCR image build: [Azure/dalec-build-defs#6817](https://github.com/Azure/dalec-build-defs/pull/6817)

## Which issue(s) this PR fixes:
N/A (design proposal)

## Special notes for your reviewer:
This is a design-only proposal with no code changes.

### DIFF
--- a/docs/proposals/20260421-migrate-epp-to-llm-d-inference-scheduler.md
+++ b/docs/proposals/20260421-migrate-epp-to-llm-d-inference-scheduler.md
@@ -131,54 +131,111 @@ All three default scorers work out of the box with KAITO's vLLM inference server
 
 The EPP scrapes metrics from the InferencePool's `targetPorts` (port 5000, the same port as KAITO's vLLM API endpoint). KAITO's vLLM exposes the `/metrics` endpoint on port 5000 with all required Prometheus metrics.
 
-### Advanced llm-d Plugins
+### Advanced: Custom EPP Plugin Configuration
 
-Users can access llm-d-specific plugins by providing custom `pluginsCustomConfig` in the InferencePool Helm values:
+For advanced plugins like `precise-prefix-cache-scorer` or P/D disaggregation, users provide a custom `EndpointPickerConfig` via a **ConfigMap reference** in the InferenceSet spec. This keeps the InferenceSet clean — the plugin configuration lives in a separate ConfigMap.
 
-#### Precise Prefix Cache Scorer
+#### API Changes
+
+Add an optional `eppPluginsConfigRef` field to InferenceSetSpec:
+
+```go
+type InferenceSetSpec struct {
+    // ...existing fields...
+
+    // EPPPluginsConfigRef references a ConfigMap containing custom EPP plugins configuration.
+    // The ConfigMap must contain a key "config.yaml" with the EndpointPickerConfig content.
+    // If not specified, the default llm-d plugins (queue-scorer, kv-cache-utilization-scorer,
+    // prefix-cache-scorer) are used.
+    // +optional
+    EPPPluginsConfigRef *corev1.LocalObjectReference `json:"eppPluginsConfigRef,omitempty"`
+}
+```
+
+#### User Experience
+
+```yaml
+apiVersion: kaito.sh/v1alpha1
+kind: InferenceSet
+metadata:
+  name: phi-4-mini
+spec:
+  eppPluginsConfigRef:
+    name: phi-4-mini-epp-plugins   # references a ConfigMap in the same namespace
+  # ...other fields unchanged...
+```
+
+#### Controller Behavior
+
+When `eppPluginsConfigRef` is set:
+
+1. Controller reads the referenced ConfigMap's `config.yaml` key
+2. Injects the content into the InferencePool HelmRelease values:
+   ```yaml
+   inferenceExtension:
+     pluginsConfigFile: "custom-plugins.yaml"
+     pluginsCustomConfig:
+       custom-plugins.yaml: <content from ConfigMap>
+   ```
+3. Flux reconciles the HelmRelease → EPP deployment picks up the custom config
+4. Controller watches the referenced ConfigMap for changes and triggers re-reconciliation
+
+#### Example: Precise Prefix Cache Scorer
 
 Token-level KV cache matching using KV events from the routing sidecar:
 
 ```yaml
-inferenceExtension:
-  pluginsCustomConfig:
-    custom-plugins.yaml: |-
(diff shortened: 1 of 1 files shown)


docs/proposals/20260421-migrate-epp-to-llm-d-inference-scheduler.md
```diff
@@ -0,0 +1,272 @@
...
+The llm-d EPP binary is fully compatible with this config format (same `EndpointPickerConfig` API).
+
+### Default Plugin Metrics Compatibility
+
+All three default scorers work out of the box with KAITO's vLLM inference server:
+
+| Plugin | Required Metric | vLLM Metric Name | Available on Port 5000 |
+|--------|----------------|-------------------|----------------------|
+| `queue-scorer` | Request queue depth | `vllm:num_requests_waiting` + `vllm:num_requests_running` | ✅ Yes |
+| `kv-cache-utilization-scorer` | KV cache usage | `vllm:kv_cache_usage_perc` | ✅ Yes |
```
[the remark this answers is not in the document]
andyzhangx: vllm:kv_cache_usage_per is correct

docs/proposals/20260421-migrate-epp-to-llm-d-inference-scheduler.md
```diff
@@ -0,0 +1,345 @@
...
+```go
+type InferenceSetSpec struct {
+    // ...existing fields...
+
+    // EPPPluginsConfigRef references a ConfigMap containing custom EPP plugins configuration.
+    // The ConfigMap must contain a key "config.yaml" with the EndpointPickerConfig content.
+    // If not specified, the default llm-d plugins (queue-scorer, kv-cache-utilization-scorer,
+    // prefix-cache-scorer) are used.
+    // +optional
+    EPPPluginsConfigRef *corev1.LocalObjectReference `json:"eppPluginsConfigRef,omitempty"`
```
rambohe-ch: 1. If the configmap about eppPluginsConfigRef is changed, should we upgrade the epp pod?

2. Shoud we verify the format and the contents of config.yaml in configmap?

andyzhangx: Good questions!

**Q1 (ConfigMap change → EPP pod restart?):**
The current llm-d EPP (v0.7.1) loads the plugin configuration once at startup via `--configFile` and does **not** support hot-reload. So yes, if the `eppPluginsConfigRef` ConfigMap is updated, the EPP pod needs to be restarted.

In practice, KAITO handles this automatically:
- The InferenceSet controller watches the referenced ConfigMap for changes
- On change, the controller updates the HelmRelease values (e.g., a checksum annotation on the EPP Deployment pod template)
- Flux reconciles the HelmRelease → triggers a rolling restart of the EPP pod

Users just update the ConfigMap, and KAITO handles the restart transparently. I will add a "Configuration Update Behavior" section to the proposal to document this.

**Q2 (Validate ConfigMap format/contents?):**
Two levels of validation:
1. **KAITO controller-level**: The InferenceSet reconciler validates that the referenced ConfigMap exists and contains the `config.yaml` key. If the ConfigMap is missing or the key is absent, the controller sets a condition on the InferenceSet (e.g., `ConfigInvalid`) and does not proceed with the HelmRelease update.
2. **EPP-level**: The llm-d EPP validates the `EndpointPickerConfig` schema on load ([configloader.go](https://github.com/kubernetes-sigs/gateway-api-inference-extension/blob/main/pkg/epp/config/loader/configloader.go)). If the config is invalid, the EPP fails to start and the pod enters CrashLoopBackOff, which is visible via standard K8s monitoring.

Deep validation of plugin names/parameters should be left to the EPP — KAITO should not hardcode knowledge of valid llm-d plugin types, as that would create tight coupling and break when new plugins are added upstream.

I will update the proposal with "Configuration Update Behavior" and "Validation" sections.

andyzhangx: Updated in latest commits (46cade4, dfd8903):

**Renamed `eppPluginsConfigRef` → `eppPluginsConfig`** (plain `string` type instead of `*corev1.LocalObjectReference`) for consistency with `runtimeConfig` in the MRI proposal.

1. **ConfigMap change handling**: Simplified — the controller only tracks the `eppPluginsConfig` field value (ConfigMap name). When the name changes, the HelmRelease is updated and Flux reconciles the EPP Deployment. If the content changes but the name stays the same, the user can either create a new ConfigMap with a different name or manually `kubectl rollout restart` the EPP pod. No ConfigMap watches or checksum annotations needed.

2. **Validation**: The controller validates that the referenced ConfigMap exists and contains the required `config.yaml` key. Format validation of the config content is left to the EPP itself at startup — this avoids duplicating llm-d's validation logic in KAITO and keeps the controller simple.

mikemorris:
Let me know if you encounter any difficulties with this approach, especially if you intend to use llm-d's EPP implementation in a standalone manner with serving backends other than llm-d.

A few impactful changes to note are that the API groups for the InferenceModelRewrite and InferenceObjective CRDs will move to llm-d's API group (and I expect would be installed with the llm-d CRDs rather than with GAIE) and that the merged EPP implementation in llm-d may _remove_ the plugin extensibility model in favor of just implementing all existing upstream and llm-d-specific plugins directly in-tree.

Issue #1952, opened 2026-04-13
Integrate KarpenterProvisioner into Workspace controller reconciliation loop


rambohe-ch:
## Description

Update the Workspace controller to support both `GpuProvisioner` and `KarpenterProvisioner` paths, selected at runtime based on the `KarpenterProvisioner` feature gate and existing NodeClaim state. At this stage the feature gate defaults to `false` — all workspaces continue using `GpuProvisioner` and existing E2E tests pass without modification.

When the feature gate is enabled in the future, new workspaces use Karpenter while legacy workspaces with live gpu-provisioner NodeClaims remain on the old path. Legacy workspaces whose nodes crash (NodeClaims gone) are automatically migrated to Karpenter.

**Ref:** [Workload Scheduling](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#workload-scheduling) | [Replicas Calculation](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#replicas-calculation) | [Deletion Strategy](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#deletion-strategy) | [Resource Ownership and Lifecycle](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#resource-ownership-and-lifecycle) | [Handling Existing (Legacy) Workspaces](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#handling-existing-legacy-workspaces)

## Workspace Controller Reconcile Logic

```
reconcile(workspace):
  // Step 1: Feature gate check
  if feature gate KarpenterProvisioner == false:
    use GpuProvisioner (existing behavior, no change)
    return

  // Step 2: Feature gate is true — determine provisioner path
  legacyNodeClaims    = list NodeClaims with label "kaito.sh/workspace=<name>"
  karpenterNodeClaims = list NodeClaims with label "karpenter.kaito.sh/workspace=<name>"

  if len(legacyNodeClaims) > 0:
    // Legacy workspace with live gpu-provisioner NodeClaims
    // Keep using GpuProvisioner — do not create duplicate nodes
    use GpuProvisioner
    return

[...]

PR #1982, opened 2026-04-22
feat: karpenter provisioner

### LINKED ISSUE #1947: Implement `KarpenterProvisioner` — NodePool lifecycle management (create/delete)
## Description

Implement the `KarpenterProvisioner` for `ProvisionNodes()` and `DeprovisionNodes()` methods, which create and delete per-Workspace NodePools.

**Ref:** [Per-Workspace NodePool](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#per-workspace-nodepool) | [NodePool Lifecycle Management](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#nodepool-lifecycle-management) | [Resource Definitions](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#resource-definitions) | [Naming Conventions](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#naming-conventions)

## Tasks

- [ ] Implement `ProvisionNodes()`: create per-Workspace NodePool with:
  - Name: `<workspace-namespace>-<workspace-name>` (truncated with hash suffix if >253 chars)
  - `replicas` = `workspace.Status.TargetNodeCount`
  - `nodeClassRef` pointing to global AKSNodeClass (`image-family-ubuntu` by default)
  - Instance type from `workspace.Resource.InstanceType`
  - Labels: `karpenter.kaito.sh/workspace: <workspace-name>`, `karpenter.kaito.sh/managed-by: kaito`
  - Taints: `karpenter.kaito.sh/workspace=<workspace-name>:NoSchedule`
  - Disruption budget: `nodes: "0"` for InferenceSet workspaces, `nodes: "1"` for standalone
- [ ] Implement `DeprovisionNodes()`: delete per-Workspace NodePool (Karpenter cascades cleanup via OwnerReference)
- [ ] `EnableDrift()`: patch NodePool budget `nodes: "0"` → `nodes: "1"`
- [ ] `DisableDrift()`: patch NodePool budget `nodes: "1"` → `nodes: "0"`
- [ ] Add unit tests for NodePool creation, deletion, and budget patching

2170chm:
Replace AzureKarpenterProvisioner with a generic KarpenterProvisioner that receives cloud-specific NodeClass config (group, kind, name mapping) at construction time. Azure-specific startup logic (CRD verification, AKSNodeClass bootstrap) is removed — Helm/installer handles NodeClass resources.

- Add NodeClassConfig struct and KarpenterProvisioner in pkg/nodeprovision/karpenter/
- Add CLI flags for karpenter node class configuration
- Add karpenter workspace nodeSelector and toleration (gated by provisioner type)
- Add kaito.sh/workspace label to NodePool template for controller watch
- Add karpenter workspace toleration to NVIDIA device plugin daemonset
- Add AKSNodeClass Helm template
- Delete pkg/nodeprovision/azure-karpenter/

**Reason for Change**:

Migrate node provisioning to using karpenter.

**Requirements**

**Issue Fixed**:

Closes #1947 
**Notes for Reviewers**:
A. Currently, kaito still defaults to using gpu provisioner since not all required features (drift controller) are implemented.

B. To use static provisioning, we need to enable `settings.featureGates.staticCapacity=true` when installing karpenter.

C. Right now still blocked by Karpenter's regional provisioning API. Forked azure karpenter and hacked it to support regional provisioning for local GPU provisioning tests. 

D. Tested the following cases end-to-end:

1. Happy path end-to-end on Cilium cluster with A100 (create → serve → delete cascade)
2. Idempotency — apply same workspace twice
3. Multi-node provisioning — 70B model gets 3 NodeClaims
4. Image family annotation — azurelinux resolves to correct NodeClass
5. InferenceSet workspace — drift budget blocked, inferenceset labels present
6. Multi-workspace isolation — same matchLabels, no cross-scheduling
7. Deletion while provisioning — cascade cleanup mid-provision
8. GPU-provisioner backward compatibility — no karpenter artifacts on pods, full happy path works


pkg/nodeprovision/karpenter/nodepool.go
```diff
@@ -0,0 +1,146 @@
...
+	if isInferenceSetWorkspace(ws) {
+		driftBudgetNodes = "0"
+	}
+
+	// Template labels propagated to NodeClaims and Nodes.
+	// Include the workspace scheduling label plus the user's matchLabels
+	// so that inference pods' nodeAffinity (built from matchLabels) is satisfied.
+	templateLabels := map[string]string{
+		consts.KarpenterWorkspaceKey:         ws.Name,
+		kaitov1beta1.LabelWorkspaceName:      ws.Name,      // Required by the controller's NodeClaim watch predicate.
```
rambohe-ch, line 83: gpu-provisioner will use this label to watch nodeclaims, this means that nodeclaims created by karpenter will affect gpu-provisioner.

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	return nil
+}
+
+// DeleteNodes deletes the NodePool for the Workspace. Idempotent — NotFound is ignored.
+// Karpenter cascades deletion: NodePool → NodeClaim → Node → VM.
+func (p *KarpenterProvisioner) DeleteNodes(ctx context.Context, ws *kaitov1beta1.Workspace) error {
+	nodePoolName := NodePoolName(ws.Namespace, ws.Name)
+	np := &karpenterv1.NodePool{}
+	np.Name = nodePoolName
```
rambohe-ch, line 80: np := &karpenterv1.NodePool{Name: nodePoolName}

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,228 @@
...
+
+	if err := p.client.Delete(ctx, np); err != nil {
```
rambohe-ch, line 82: If nodepool.DeletionTimestamp is not nil, we should skip the delete operation. because it means nodepool is under deleting.

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,228 @@
...
+	if int32(len(nodeClaimList.Items)) < ws.Status.TargetNodeCount {
+		return false, false, nil
+	}
+
+	for i := range nodeClaimList.Items {
+		if !nodeclaim.IsNodeClaimReadyNotDeleting(&nodeClaimList.Items[i]) {
+			return false, false, nil
+		}
+	}
+
```
rambohe-ch, line 114: Maybe we should verify all nodes are ready or not after nodeclaim verification. because gpu resource may not be ready even node is ready.

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,228 @@
...
+			}
+			if found {
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("NodePool %q has no budget entry with Drifted reason", nodePoolName)
+		}
+
+		return p.client.Update(ctx, np)
```
rambohe-ch, line 156: if budget nodes is not changed, we should not update the nodepool.

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,228 @@
...
+
+	targetCount := int(ws.Status.TargetNodeCount)
+	if readyCount >= targetCount {
+		nodeClaimCond.Status = metav1.ConditionTrue
+		nodeClaimCond.Reason = "NodeClaimsReady"
+		nodeClaimCond.Message = "Enough NodeClaims are ready"
+
+		// For karpenter, NodeClaim readiness implies node readiness because
+		// karpenter only marks a NodeClaim Ready after the underlying node
+		// is registered and ready.
```
rambohe-ch, line 208: we can not recognize a node is ready even the node is ready, because the gpu resource may not ready for the node.

rambohe-ch:
There is another question about deleting global nodeclass from karpenter. Because we can not ensure the installation orders of karpenter and kaito. If kaito is installed before karpenter, the NodeClass in the helm chart can not be created because related CRDs resources are not installed by Kaprenter.

pkg/nodeprovision/karpenter/nodepool.go
```diff
@@ -0,0 +1,146 @@
...
+	driftBudgetNodes := "1"
+	if isInferenceSetWorkspace(ws) {
+		driftBudgetNodes = "0"
+	}
+
+	// Template labels propagated to NodeClaims and Nodes.
+	// Include the workspace scheduling label plus the user's matchLabels
+	// so that inference pods' nodeAffinity (built from matchLabels) is satisfied.
+	templateLabels := map[string]string{
+		consts.KarpenterWorkspaceKey:         ws.Name,
```
Fei-Guo: If this label is used by pod selector, the value needs to be ws's full name (ns + name).

charts/kaito/workspace/templates/aksnodeclass.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and (eq .Values.nodeProvisioner "karpenter") (eq .Values.karpenterNodeClass.group "karpenter.azure.com") }}
```
Fei-Guo: If this is the general node class template, should we rename it? aksnodeclass.yaml -> nodeclass.yaml

2170chm: This is not a general node class template. It is gated for aks only. Other provider will have different node class yamls, and when we introduce new providers in the future, we can add new yamls for each.

2170chm:
> There is another question about deleting global nodeclass from karpenter. Because we can not ensure the installation orders of karpenter and kaito. If kaito is installed before karpenter, the NodeClass in the helm chart can not be created because related CRDs resources are not installed by Kaprenter.

Yes that's a valid concern. But to support cloud-agnostic provisioner, controller shouldn't know about the nodeclass. One solution is to have helm store the nodeclass in a config map and controller can create it in Start() after verifying that karpenter CRDs exist. What do you think?

pkg/nodeprovision/karpenter/nodepool.go
2170chm: Good point, I added this to make this PR testable and planned to remove this and change the event handler in #1952, but it is confusing so I will change the event handler in this PR as well.

pkg/workspace/inference/preset_inferences.go
```diff
@@ -105,6 +105,17 @@ func defaultTolerations() []corev1.Toleration {
...
 	}
 
+	// Tolerate the karpenter workspace taint so inference pods can schedule
+	// on karpenter-provisioned GPU nodes.
+	if consts.IsKarpenterProvisioner() {
+		tolerations = append(tolerations, corev1.Toleration{
+			Effect:   corev1.TaintEffectNoSchedule,
+			Key:      consts.KarpenterWorkspaceKey,
+			Operator: corev1.TolerationOpEqual,
+			Value:    ws.Namespace + "-" + ws.Name,
```
zhuangqh: Should be NodePoolName(xx, xx)?

pkg/workspace/controllers/workspace_event_handler.go
```diff
@@ -37,15 +39,20 @@ type nodeClaimEventHandler struct {
...
+			return &client.ObjectKey{Namespace: namespace, Name: name}
+		}
 	}
-	namespace, ok := nc.Labels[kaitov1beta1.LabelWorkspaceNamespace]
-	if !ok {
-		return nil
+	// Karpenter path — label value is "namespace-name" (nodePoolName format).
+	if nodePoolName, ok := nc.Labels[consts.KarpenterWorkspaceKey]; ok {
+		if namespace, ok := nc.Labels[consts.KarpenterWorkspaceNamespaceKey]; ok {
+			name := strings.TrimPrefix(nodePoolName, namespace+"-")
```
zhuangqh: what if ns-name exceed max length?

pkg/nodeprovision/manager/factory.go
```diff
@@ -18,24 +18,29 @@ import (
...
 
 // NewNodeProvisioner creates and returns a NodeProvisioner based on the provisionerType parameter.
 //
-//   - azure-karpenter: AzureKarpenterProvisioner (uses directClient for
-//     CRD verification and global AKSNodeClass bootstrap at Start time).
+//   - karpenter: KarpenterProvisioner (cloud-agnostic karpenter NodePool CRUD).
 //   - byo: BYOProvisioner (all provisioning ops are no-ops).
 //   - azure-gpu-provisioner (default): AzureGPUProvisioner (creates/deletes NodeClaims).
-func NewNodeProvisioner(kClient, directClient client.Client, recorder record.EventRecorder, defaultNodeImageFamily string, provisionerType string) nodeprovision.NodeProvisioner {
+func NewNodeProvisioner(kClient, directClient client.Client, recorder record.EventRecorder, defaultNodeImageFamily string, provisionerType string, nodeClassGroup, nodeClassKind, nodeClassDefaultName string, imageFamilyNames map[string]string) nodeprovision.NodeProvisioner {
```
zhuangqh: nit: I would suggest introducing a struct for that. too many arguments here.

```go
type ProvisionerConfig struct {
    KClient                client.Client
    DirectClient           client.Client
    Recorder               record.EventRecorder
    DefaultNodeImageFamily string
    ProvisionerType        string
    NodeClassGroup         string
    NodeClassKind          string
    NodeClassDefaultName   string
    ImageFamilyNames       map[string]string
}
```

pkg/nodeprovision/karpenter/nodepool.go
rambohe-ch: no problem

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,260 @@
...
+func NewKarpenterProvisioner(c client.Client, cfg NodeClassConfig) *KarpenterProvisioner {
+	return &KarpenterProvisioner{client: c, nodeClassConfig: cfg, nodeResourceManager: resource.NewNodeManager(c)}
+}
+
+// Name returns the provisioner name.
+func (p *KarpenterProvisioner) Name() string { return "KarpenterProvisioner" }
+
+// Start is a no-op. The installer/Helm chart is responsible for ensuring
+// NodeClass resources exist before KAITO starts.
+func (p *KarpenterProvisioner) Start(_ context.Context) error { return nil }
```
rambohe-ch: We should check related NodeClass existence and readiness  before provisioning nodes.

rambohe-ch:
> > There is another question about deleting global nodeclass from karpenter. Because we can not ensure the installation orders of karpenter and kaito. If kaito is installed before karpenter, the NodeClass in the helm chart can not be created because related CRDs resources are not installed by Kaprenter.
> 
> Yes that's a valid concern. But to support cloud-agnostic provisioner, controller shouldn't know about the nodeclass. One solution is to have helm store the nodeclass in a config map and controller can create it in Start() after verifying that karpenter CRDs exist. What do you think?

No problem, and we should check the nodeclass existence and readiness before creating these NodeClass.

zhuangqh (approved):

charts/kaito/workspace/values.yaml
```diff
@@ -43,6 +43,24 @@ featureGates:
   enableInferenceSetController: false
 defaultNodeImageFamily: ""
 nodeProvisioner: ""
+karpenterProvider: "karpenterAzure"
```
rambohe-ch: karpenterAzure --> azure

charts/kaito/workspace/values.yaml
```diff
@@ -43,6 +43,24 @@ featureGates:
...
+karpenterAzure:
+  group: "karpenter.azure.com"
+  kind: "AKSNodeClass"
+  apiVersion: "v1beta1"
```
rambohe-ch: apiVersion --> version

charts/kaito/workspace/values.yaml
```diff
@@ -43,6 +43,24 @@ featureGates:
...
+  resourceName: "aksnodeclasses.karpenter.azure.com"
+  nodeClasses:
+    - name: "image-family-ubuntu"
+      default: true
+      annotationValue: "ubuntu"
```
rambohe-ch: It seems that annotationValue is unused.

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,423 @@
...
+
+// NodeClassConfig holds cloud-specific NodeClass reference info.
+// Group, Kind, and ResourceName are injected via CLI flags.
+// DefaultName and AnnotationMap are derived by Start() from ConfigMap labels.
+type NodeClassConfig struct {
+	Group         string            // e.g. "karpenter.azure.com"
+	Kind          string            // e.g. "AKSNodeClass"
+	ResourceName  string            // full CRD resource name (e.g. "aksnodeclasses.karpenter.azure.com")
+	DefaultName   string            // populated by Start(): name of entry with karpenter.kaito.sh/default=true
+	AnnotationMap map[string]string // populated by Start(): annotationValue → name
```
rambohe-ch: It seems that AnnotationMap is unused.

charts/kaito/workspace/values.yaml
2170chm: It is read and put as a label on the nodeclass, which is used to construct the annotation map.

Then, in the resolveNodeClassName(), if an nodeimagefamily annotation is provided by the user, it will use that map to find the right nodeclass. That way for other providers, if they don't have a nodeimagefamily, user won't supply this value and it will just use the default.

However, this still means the controller is coupled with the nodeImageFamily annotation, and in the future, if other providers need nodeclass selection through workspace config, it will require a controller code change. But I still keep it there because I don't want to change the existing API. To generalize it, since we have decided to put each image family as a single nodeclass, we can make this annotation's name more generic, something like nodeclass. But that will change the API. What do you think?

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,423 @@
...
+
+// checkNodeClassReady performs a single point-in-time check that the named
+// NodeClass exists and has a Ready=True condition. Unlike waitForNodeClassReady,
+// it does not poll — it returns an error immediately if the resource is missing
+// or not ready.
+func (p *KarpenterProvisioner) checkNodeClassReady(ctx context.Context, name string) error {
+	obj := &unstructured.Unstructured{}
+	obj.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   p.nodeClassConfig.Group,
+		Version: "v1beta1",
```
rambohe-ch: Version shoud be passed from the startup parameter, and can not be fixed to v1beta1.

pkg/nodeprovision/karpenter/provisioner.go
```diff
@@ -0,0 +1,423 @@
...
+	return fmt.Errorf("NodeClass %q exists but is not Ready", name)
+}
+
+// waitForNodeClassReady polls until the NodeClass has a Ready=True condition.
+func (p *KarpenterProvisioner) waitForNodeClassReady(ctx context.Context, name string) error {
+	return wait.PollUntilContextTimeout(ctx, 5*time.Second, 120*time.Second, true, func(ctx context.Context) (bool, error) {
+		obj := &unstructured.Unstructured{}
+		obj.SetGroupVersionKind(schema.GroupVersionKind{
+			Group:   p.nodeClassConfig.Group,
+			Version: "v1beta1",
```
rambohe-ch: ditto

rambohe-ch (approved):

zhuangqh (approved):
Good job!

Fei-Guo (approved):

PR #1992, opened 2026-04-24
feat: wire default guardrails policy configmap for ragengine

xiaoqi-7:
## Summary
This PR wires the default output guardrails policy ConfigMap into RAGEngine deployments.

When `spec.guardrails.enabled` is `true`, the controller now ensures that a guardrails policy ConfigMap is available for the workload:
- if `spec.guardrails.configMapRef.name` is provided, use that ConfigMap
- otherwise, copy the default `ragengine-guardrails-policy-template` ConfigMap from the release namespace into the target namespace

The deployment then mounts the policy file and exposes its path to the RAG runtime through environment variables.

## What Changed
- added guardrails policy template constants in `api/v1beta1/ragengine_types.go`
- updated guardrails validation in `api/v1beta1/ragengine_validation.go` to:
  - validate a user-provided ConfigMap in the workload namespace
  - validate the default guardrails template in the release namespace when guardrails are enabled without `configMapRef`
- added controller logic in `pkg/ragengine/controllers/preset_rag.go` to:
  - resolve user-provided or default guardrails ConfigMaps
  - copy the default template into the target namespace when needed
  - mount the ConfigMap into the RAG pod
- updated `pkg/ragengine/manifests/manifests.go` to inject:
  - `OUTPUT_GUARDRAILS_ENABLED=true`
  - `OUTPUT_GUARDRAILS_POLICY_PATH=/etc/kaito/guardrails/guardrails.yaml`
- added the default Helm template:
  - `charts/kaito/ragengine/templates/guardrails-policy-configmap.yaml`

## Behavior
When `spec.guardrails.enabled` is `true`:
- if `spec.guardrails.configMapRef.name` is set, the deployment mounts that ConfigMap
- if `spec.guardrails.configMapRef` is omitted, the controller copies `ragengine-guardrails-policy-template` from the release namespace into the RAGEngine namespace and mounts it
- the pod mounts the policy at `/etc/kaito/guardrails/guardrails.yaml`
- the manifest sets `OUTPUT_GUARDRAILS_POLICY_PATH` to that mounted path

## Validation
Ran:

```bash
go test ./pkg/ragengine/controllers ./pkg/ragengine/manifests ./api/v1beta1 -run 'TestCreatePresetRAGGuardrails|TestRAGSetEnv|TestRAGEngineValidateGuardrails|TestCreatePresetRAG|TestGenerateRAGDeploymentManifestDifferentConfigurations'

### CODE REVIEW

File: api/v1beta1/ragengine_types.go [hunk 1]
```diff
@@ -17,6 +17,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
```
Fei-Guo, line 20: DefaultGuardrailsPolicyConfigMapTemplate -> DefaultGuardrailsPolicyConfigMapName

api/v1beta1/ragengine_validation.go
```diff
@@ -99,34 +97,48 @@ func (w *RAGEngine) validateGuardrails(ctx context.Context) (errs *apis.FieldErr
...
+	field := "configMapRef.name"
+	if guardrails.ConfigMapRef != nil && guardrails.ConfigMapRef.Name != "" {
+		cmName = guardrails.ConfigMapRef.Name
+	} else {
+		releaseNamespace, err := utils.GetReleaseNamespace()
+		if err != nil {
+			return apis.ErrGeneric(fmt.Sprintf("Failed to determine release namespace: %v", err), "namespace")
+		}
+		cmNamespace = releaseNamespace
+		field = "enabled"
```
Fei-Guo, line 123: Does it make more sense to choose field = "configMapRef" in this case?

charts/kaito/ragengine/templates/guardrails-policy-configmap.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: ragengine-guardrails-policy-template
+  namespace: {{ .Release.Namespace }}
+data:
+  guardrails.yaml: |
+    action: block
+    blockMessage: The model output was blocked by output guardrails.
+    scanners: []
```
Fei-Guo: Is this intended the default scanner is empty? If user enables guardrial only without providing a custom profiles, they are supposed to get some default protections, right?

docs/proposals/20260416-ragengine-guardrails-ux-api.md
```diff
@@ -87,8 +87,14 @@ in ConfigMap YAML, not in the CRD.
...
-Follow-up implementation may provide a default ConfigMap and default mount path so that
-guardrail policy can be enabled without introducing a broad CRD surface in the same step.
+If `spec.guardrails.enabled` is `true` and `configMapRef` is not set, the
+controller copies the default guardrails policy ConfigMap
+(`ragengine-guardrails-policy-template`) into the RAGEngine namespace and mounts
+it into the Pod.
+
+- Auto-copied ConfigMaps carry an `OwnerReference` to the RAGEngine.
+- User-provided ConfigMaps are not modified or owned by the controller.
+- Hot reload is not part of this PR.
```
Fei-Guo, line 100: You probably should mention the default scanners here.

pkg/ragengine/controllers/preset_rag.go
```diff
@@ -140,16 +141,18 @@ func configStorageVolume(storageSpec *v1beta1.StorageSpec) (corev1.Volume, corev
...
 			Name: manifests.GuardrailsPolicyVolumeName,
 			VolumeSource: corev1.VolumeSource{
 				ConfigMap: &corev1.ConfigMapVolumeSource{
 					LocalObjectReference: corev1.LocalObjectReference{
-						Name: configMapName,
+						Name: cmName,
 					},
+					Items: []corev1.KeyToPath{{
+						Key:  manifests.GuardrailsPolicyFileName,
+						Path: manifests.GuardrailsPolicyFileName,
```
Fei-Guo: Should this be manifests.GuardrialsPolicyFilePath?

pkg/ragengine/controllers/preset_rag.go
```diff
@@ -159,6 +162,55 @@ func configGuardrailsVolume(configMapName string) (corev1.Volume, corev1.VolumeM
...
+		client.ObjectKey{Name: v1beta1.DefaultGuardrailsPolicyConfigMapTemplate},
+	)
+	if err != nil || cm == nil {
+		return cm, err
+	}
+
+	// User-provided ConfigMaps belong to the user; never patch them.
+	// Only the auto-copied default template should follow the RAGEngine
+	// lifecycle so it is garbage-collected when the RAGEngine is deleted.
+	if userOwned {
```
Fei-Guo: It is hard to add an owner here because the same configmap can be shared by multiple ragengine CRs in the same namespace. i.e., you just need to copy once. 

If you remove the configmap in any of ragengine cr cleanup process, it will likely break other ragengine crs. 

To fix the problem, we need to rename the configmap to use a ragengine CR specific name, which may introduce other problems. Hence, we should not add owner to the copied configmap and yes, we may leave this configmap uncleaned even if we delete the ragengine.

pkg/ragengine/controllers/preset_rag.go
xiaoqi-7: No, this is correct.
get the key from ConfigMap: guardrails.yaml
the file under the dir of volume mount: guardrails.yaml

api/v1beta1/ragengine_validation.go
xiaoqi-7: No. I prefer to keeping it on enabled makes more sense here: in the default-policy path, configMapRef was not provided by the user, and the actual failure is that guardrails was enabled but no default policy ConfigMap could be resolved. The change is only to make that reason explicit in the error message. I changed the error message info to make this clearer.

Fei-Guo (approved):

PR #2003, opened 2026-04-28
feat: karpenter drift controller

### LINKED ISSUE #1948: Implement InferenceSet-level drift orchestration controller
## Description

Implement the drift orchestrator that controls the pace of node upgrades across Workspaces within an InferenceSet by toggling NodePool disruption budgets one Workspace at a time.

**Ref:** [Drift Orchestration](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#drift-orchestration)

## Tasks

- [ ] Watch NodeClaims across all Workspaces in an InferenceSet for `Drifted=True` condition
- [ ] Select ONE Workspace at a time for drift upgrade
- [ ] Call `EnableDrift()` to patch NodePool budget from `nodes: "0"` → `nodes: "1"`
- [ ] Wait until all NodeClaims in that Workspace are no longer drifted
- [ ] Call `DisableDrift()` to restore `nodes: "0"`
- [ ] Move to next Workspace, repeat until all Workspaces are upgraded
- [ ] Add unit and integration tests for the orchestration flow

2170chm:
**Reason for Change**:

- Built a new controller that orchestrates rolling drift upgrades for InferenceSet workspaces. It upgrades one workspace's nodepool at a time per InferenceSet while allowing inter-InferenceSet concurrency.

- Uses Karpenter's NodePool disruption budgets as the control mechanism: budget "0" blocks drift replacement, budget "1" allows it.

- Watches NodeClaim events filtered by InferenceSet labels, so standalone workspaces are completely ignored (they use budget "1" by default and Karpenter handles them directly)

- Added a workload readiness gate. After a node is replaced, the controller waits for WorkspaceSucceeded=True before proceeding to the next workspace, preventing service downtime when all workloads are not ready.

**Requirements**

**Issue Fixed**:

Closes #1948 
**Notes for Reviewers**:
Verified e2e:
Test 1 (Single Workspace Drift): budget cycled 0→1→0, NodeClaim replaced, workload came up on the new node
Test 2 (Serial Upgrade, 3 replicas): upgrades were one-at-a-time, workload readiness gate held before each transition
Test 3 (Inter-InferenceSet Concurrency): is-a and is-b upgraded in parallel, while is-b still serialized internally
Test 4 (No-Op): no drifted NodeClaims, budget stayed at "0", controller did nothing
Test 5 (Controller Restart): killed the controller mid-upgrade, it restarted and resumed where it left off without duplicating or skipping work
Test 6 (Standalone Workspace): standalone NodePool had budget "1" from creation, Karpenter handled drift directly, drift controller never touched it

### DIFF
--- a/cmd/workspace/main.go
+++ b/cmd/workspace/main.go
@@ -47,6 +47,7 @@ import (
 
 	kaitov1alpha1 "github.com/kaito-project/kaito/api/v1alpha1"
 	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
+	drift "github.com/kaito-project/kaito/pkg/controllers/drift"
 	"github.com/kaito-project/kaito/pkg/featuregates"
 	"github.com/kaito-project/kaito/pkg/inferenceset"
 	"github.com/kaito-project/kaito/pkg/k8sclient"
@@ -276,6 +277,16 @@ func main() {
 
 		if err = inferenceSetReconciler.SetupWithManager(mgr); err != nil {
 			klog.ErrorS(err, "unable to create controller", "controller", "InferenceSet")
+			exitWithErrorFunc()
+		}
+
+		driftReconciler := drift.NewDriftReconciler(
+			kClient,
+			mgr.GetScheme(),
+			mgr.GetEventRecorderFor("drift-controller"),
+		)
+		if err = driftReconciler.SetupWithManager(mgr); err != nil {
+			klog.ErrorS(err, "unable to create controller", "controller", "Drift")
 			exitWithErrorFunc()
 		}
 	}
--- a/cmd/workspace/main.go
+++ b/cmd/workspace/main.go
@@ -280,15 +280,17 @@ func main() {
 			exitWithErrorFunc()
 		}
 
-		driftReconciler := drift.NewDriftReconciler(
-			kClient,
-			mgr.GetScheme(),
-			mgr.GetEventRecorderFor("drift-controller"),
-			nodeProvisioner,
-		)
-		if err = driftReconciler.SetupWithManager(mgr); err != nil {
-			klog.ErrorS(err, "unable to create controller", "controller", "Drift")
-			exitWithErrorFunc()
+		if consts.IsKarpenterProvisioner() {
+			driftReconciler := drift.NewDriftReconciler(
+				kClient,
+				mgr.GetScheme(),
+				mgr.GetEventRecorderFor("drift-controller"),
+				nodeProvisioner,
+			)
+			if err = driftReconciler.SetupWithManager(mgr); err != nil {
+				klog.ErrorS(err, "unable to create controller", "controller", "Drift")
+				exitWithErrorFunc()
+			}
 		}
 	}
 
--- a/cmd/workspace/main.go
+++ b/cmd/workspace/main.go
@@ -284,6 +284,7 @@ func main() {
 			kClient,
 			mgr.GetScheme(),
 			mgr.GetEventRecorderFor("drift-controller"),
+			nodeProvisioner,
 		)
 		if err = driftReconciler.SetupWithManager(mgr); err != nil {
 			klog.ErrorS(err, "unable to create controller", "controller", "Drift")
--- a/pkg/controllers/drift/controller.go
+++ b/pkg/controllers/drift/controller.go
@@ -19,6 +19,7 @@ import (
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -33,6 +34,7 @@ import (
 	karpenterv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 
 	kaitov1alpha1 "github.com/kaito-project/kaito/api/v1alpha1"
+	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
 	"github.com/kaito-project/kaito/pkg/nodeprovision"
 	karpenterpkg "github.com/kaito-project/kaito/pkg/nodeprovision/karpenter"
 	"github.com/kaito-project/kaito/pkg/utils/consts"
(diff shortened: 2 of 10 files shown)


pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,289 @@
...
+			break // Only one should be upgrading at a time
+		}
+
+		// Check if this workspace has drifted NodeClaims.
+		hasDrifted, err := nodeclaim.HasDriftedNodeClaims(ctx, r.Client, nodePoolName)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("checking drifted NodeClaims for NodePool %q: %w", nodePoolName, err)
+		}
+		if hasDrifted {
+			driftedCandidates = append(driftedCandidates, workspaceDriftInfo{
```
Fei-Guo: You only need one driftcandidate, so you can stop the loop as long as you find one and you don't need an array of candidates.

2170chm: Good point. We still need to continue the loop to scan everything because we might find an actively upgrading nodepool later.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,289 @@
...
+				nodePoolName:       nodePoolName,
+			}
+			break // Only one should be upgrading at a time
+		}
+
+		// Check if this workspace has drifted NodeClaims.
+		hasDrifted, err := nodeclaim.HasDriftedNodeClaims(ctx, r.Client, nodePoolName)
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("checking drifted NodeClaims for NodePool %q: %w", nodePoolName, err)
+		}
```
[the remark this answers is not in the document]
2170chm: Practically, the number of workspaces and nodeclaims per nodepool is very small. Also, the current implementation skips the call once we find the first candidate.

zhuangqh (approved):

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,294 @@
...
+		if err != nil {
+			return ctrl.Result{}, fmt.Errorf("checking drifted NodeClaims for upgrading NodePool %q: %w",
+				upgrading.nodePoolName, err)
+		}
+		if hasDrifted {
+			// Still drifted — requeue and wait for karpenter to finish replacement.
+			klog.V(2).InfoS("Workspace still has drifted NodeClaims, waiting",
+				"workspace", klog.KRef(upgrading.workspaceNamespace, upgrading.workspaceName),
+				"nodePool", upgrading.nodePoolName)
+			return ctrl.Result{RequeueAfter: driftRequeueInterval}, nil
```
rambohe-ch: I think the best parctice for reconcile requeue is: we have enough reason not to monitor the trigger resource, so we need to requeue explictly. because fixed time requeue is not responsive, we should base on the resouce changes to trigger reconcile.

So i recommend to add a Workspace watch in SetupWithManager. so one of them(nodeclaim, workspace, inferenceset) changes, the reconcile will be triggered, this means we can return `ctrl.Result{}, nil` instead of adding driftRequeueInterval.

cmd/workspace/main.go
```diff
@@ -278,6 +279,17 @@ func main() {
...
 			exitWithErrorFunc()
 		}
+
+		driftReconciler := drift.NewDriftReconciler(
+			kClient,
+			mgr.GetScheme(),
+			mgr.GetEventRecorderFor("drift-controller"),
+			nodeProvisioner,
+		)
+		if err = driftReconciler.SetupWithManager(mgr); err != nil {
```
rambohe-ch: we should register this controller when nodeprovisioner is karpenter. otherwise this controller will dump error logs.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,294 @@
...
+//  2. For each Workspace, get its NodePool and read the drift budget
+//  3. If one has budget "1" (upgrade in progress):
+//     - If it still has drifted NodeClaims: requeue (wait)
+//     - If no drifted NodeClaims but workload not ready: requeue (wait)
+//     - If no drifted NodeClaims and workload ready: set budget back to "0", requeue
+//  4. If none has budget "1" (no upgrade in progress):
+//     - Find next workspace with drifted NodeClaims
+//     - Set its budget to "1" (enable drift), requeue
+//     - If no drifted workspaces: done (no requeue)
+func (r *DriftReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
rambohe-ch, line 97: Please consider consolidating the NodeClaim List calls. HasDriftedNodeClaims is invoked once per workspace inside the loop, and again for the upgrading workspace afterwards — up to N+1 List calls per Reconcile. Since all NodeClaims managed by this InferenceSet already carry KarpenterInferenceSetKey / KarpenterInferenceSetNamespaceKey labels, we can List them once at the top of Reconcile filtered by those labels, group by karpenterv1.NodePoolLabelKey into a driftedByNodePool map[string]bool, and then look up drift status in O(1) for both the loop and the upgrading-workspace check.

pkg/controllers/drift/controller.go
```diff
@@ -0,0 +1,294 @@
...
+	inferenceSet := &kaitov1alpha1.InferenceSet{}
+	if err := r.Get(ctx, req.NamespacedName, inferenceSet); err != nil {
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// 2. List Workspaces for this InferenceSet.
+	wsList, err := inferencesetutil.ListWorkspaces(ctx, inferenceSet, r.Client)
```
rambohe-ch: Reconcile should be NodePool-centric; Workspace is only needed for the readiness check

Looking at how ws is actually used in Reconcile, only isWorkspaceReady truly needs the Workspace object — every other use (NodePoolName(ns, name), log/event fields, EnableDriftRemediation / DisableDriftRemediation arguments) just needs the (namespace, name) pair. If NodePools created by karpenter pkg already carry the InferenceSet labels (KarpenterInferenceSetKey / KarpenterInferenceSetNamespaceKey) and the workspace ns/name labels, we could drive the entire reconcile from a single List NodePools (filtered by InferenceSet labels) plus a single List NodeClaims (same filter), and only fetch the Workspace once — for the upgrading one — when we reach the readiness check.

rambohe-ch (approved):

PR #2004, opened 2026-04-28
feat: support deployment name as served-mode-name for inferenceset

### LINKED ISSUE #2001: Use InferenceSet name as served-model-name for workspaces created by InferenceSet

When a Workspace is created by an InferenceSet, the `served-model-name` parameter is set to the preset model name (e.g., `p.VLLM.ModelName` or `p.Transformers.ModelName`). This means all workspace replicas managed by the same InferenceSet expose a model name that is the raw model identifier rather than the InferenceSet name. For users routing requests through the InferenceSet abstraction, having the served model name match the InferenceSet name would provide a more consistent and intuitive experience.

- If a Workspace is created by an InferenceSet (identifiable via the `inferenceset.kaito.sh/created-by` label), the `served-model-name` should be set to the InferenceSet's name instead of the default preset model name.
- If a Workspace is standalone (not created by an InferenceSet), the current behavior should be preserved — `served-model-name` continues to use the preset model name.

- Users could manually override `served-model-name` via runtime parameters in the InferenceSet template, but this adds unnecessary configuration burden and is error-prone.

**Additional context**

Currently, `served-model-name` is set in `pkg/model/interface.go`:
- For vLLM: `p.VLLM.ModelRunParams["served-model-name"] = p.VLLM.ModelName`

The InferenceSet controller creates workspaces with label `inferenceset.kaito.sh/created-by=<inferenceset-name>`, which can be used to determine the origin of a workspace and override the served model name accordingly.

rambohe-ch:
**Reason for Change**:

support deployment name(InferenceSet.Name) as served-mode-name for inferenceset

**Requirements**

**Issue Fixed**:

Fixes #2001 

**Notes for Reviewers**:

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/kaito-project/kaito/pkg/sku"
 	"github.com/kaito-project/kaito/pkg/utils"
+	"github.com/kaito-project/kaito/pkg/utils/consts"
 )
 
 type Model interface {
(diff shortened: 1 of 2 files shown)


zhuangqh (approved):
LGTM

PR #2007, opened 2026-04-29
feat: add AWQ quantization support

zhehli688:
**Reason for Change**:

This PR added support for AWQ-quantized models in Kaito. Changes include:
- Supported parsing quantization_config in config.json in HuggingFace.
- Introduced fields quantMethod and quantBits to model catalog and onboarded Qwen/Qwen3-8B-AWQ as an example.
- Set vllm dtype parameter to 'auto' for quantized models. This is needed because in vLLM quantizations usually only support specific dtypes (e.g. only float16 is supported when quantization=awq).
- Fix model license loading for Nvidia models.

NOTE:
- We are not passing flag 'quantization' to vLLM. This is because vLLM can auto-select the optimal quantization method based on hardware and kernel availability. For example, vLLM will choose between awq_marlin (fastest), awq and awq_triton when a model's quantization_method is awq. Passing 'quantization=awq' to vLLM will enforce awq and affect inference speed.
- There is no need to update the KV cache calculation in Kaito's memory estimator, because AWQ only affects the size of weights instead of KV cache. However, quantization for KV cache is available in vLLM: https://docs.vllm.ai/en/latest/features/quantization/quantized_kvcache/. We may add support for it in future.

**Requirements**

- [x] tested AWQ models on both catalog and non-catalog paths:
Screenshot 2026-04-28 at 9 10 04 PM
- [x] verified numbers produced by memory estimator are correct:
Screenshot 2026-04-28 at 10 39 17 PM

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -153,6 +153,18 @@ type Metadata struct {
 	// This field is only for best effort supported vLLM models.
 	// +optional
 	AllowRemoteFiles bool `yaml:"allowRemoteFiles,omitempty"`
+
+	// QuantMethod specifies the weight quantization method used by the model
+	// (e.g., "awq", "gptq"). Maps to quantization_config.quant_method in the
+	// model's HuggingFace config.json.
+	// +optional
+	QuantMethod string `yaml:"quantMethod,omitempty"`
+
+	// QuantBits specifies the number of bits used for weight quantization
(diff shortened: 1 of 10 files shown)


Fei-Guo (approved):

zhuangqh (approved):

zhuangqh:
Nice work!

PR #2009, opened 2026-04-30
feat: legacy nodeprovisioner nodeclaim garbage collection controller

### LINKED ISSUE #1949: Implement Legacy NodeClaim garbage collection controller
## Description

Add a dedicated GC controller that periodically scans for and cleans up leaked legacy gpu-provisioner NodeClaims that have been superseded by Karpenter-managed nodes.

**Ref:** [Legacy NodeClaim GC Controller](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#legacy-nodeclaim-gc-controller) | [Handling Existing (Legacy) Workspaces](https://github.com/kaito-project/kaito/blob/main/docs/proposals/20260407-migrate-node-provisioner-to-azure-karpenter.md#handling-existing-legacy-workspaces)

## Tasks

- [ ] Implement periodic scan (e.g., every 10 seconds)
- [ ] For each Workspace with legacy NodeClaims (`kaito.sh/workspace` label):
  - Check if all replacement Karpenter NodeClaims (`karpenter.kaito.sh/workspace` label) are Ready
  - If yes, delete the legacy NodeClaims
- [ ] Only target NodeClaims with `kaito.sh/workspace` label — never touch `karpenter.kaito.sh/*` labeled ones
- [ ] Add unit tests and integration tests

2170chm:
**Reason for Change**:

Adds a LegacyNodeClaimGCController that periodically scans for legacy gpu-provisioner NodeClaims (kaito.sh/workspace label) and deletes them once safely replaced.

**Requirements**

**Issue Fixed**:

Closes #1949 
**Notes for Reviewers**:

Important Design Decisions that I made:

GC Controller (this PR)
Four conditions must all hold before deletion when the workspace exists:

1. All Karpenter replacement NodeClaims are Ready and not Deleting
2. Karpenter NodeClaim count matches status.targetNodeCount
3. Workspace has WorkspaceSucceeded=True
4. All workspace pods are Running on Karpenter nodes (not legacy nodes)
Condition 4 prevents the race where GC deletes legacy NodeClaims while the workspace is "Ready" from pods still running on those nodes.

When the workspace is deleted, legacy NodeClaims are orphaned, so GC deletes them unconditionally (only condition 1 required). gpu-provisioner may not be running to clean these up, which is why GC exists as a separate controller.

Future: Workspace Controller Migration Patch (not in this PR)
When legacy NodeClaims exist and karpenter mode is active, the workspace controller will:

1. Detect migration state (legacy NodeClaims + karpenter mode + NodePool exists)
2. Wait for all Karpenter NodeClaims to be Ready
3. Patch the StatefulSet pod template's nodeSelector from affinity-based (legacy) to karpenter.kaito.sh/workspace.
4. Rolling update moves pods to Karpenter nodes (zero downtime since nodes are pre-provisioned)
5. The workspace controller does NOT delete legacy NodeClaims. GC will do it. Once pods land on Karpenter nodes, GC's condition 4 is satisfied and it cleans up the legacy NodeClaims.

The reason for this design is that it maintains service liveness during the transition if there are more than one pod and avoids pod eviction for a shorter downtime for single-pod service.

Note that GC controller only runs when karpenter is set as the nodeprovisioner in helm, so it won't run until karpenter migration is fully released.

### DIFF
--- a/cmd/workspace/main.go
+++ b/cmd/workspace/main.go
@@ -48,6 +48,7 @@ import (
 	kaitov1alpha1 "github.com/kaito-project/kaito/api/v1alpha1"
 	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
 	drift "github.com/kaito-project/kaito/pkg/controllers/drift"
+	gc "github.com/kaito-project/kaito/pkg/controllers/gc"
 	"github.com/kaito-project/kaito/pkg/featuregates"
 	"github.com/kaito-project/kaito/pkg/inferenceset"
 	"github.com/kaito-project/kaito/pkg/k8sclient"
@@ -289,6 +290,17 @@ func main() {
 			)
 			if err = driftReconciler.SetupWithManager(mgr); err != nil {
 				klog.ErrorS(err, "unable to create controller", "controller", "Drift")
+				exitWithErrorFunc()
+			}
+		}
+
+		if consts.IsKarpenterProvisioner() {
+			gcController := gc.NewLegacyNodeClaimGCController(
+				kClient,
+				mgr.GetEventRecorderFor("legacy-gc-controller"),
+			)
+			if err = mgr.Add(gcController); err != nil {
+				klog.ErrorS(err, "unable to add controller", "controller", "LegacyNodeClaimGC")
 				exitWithErrorFunc()
 			}
 		}
--- a/pkg/controllers/gc/controller.go
+++ b/pkg/controllers/gc/controller.go
@@ -1,0 +1,288 @@
+// Copyright (c) KAITO authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package gc
+
+import (
+	"context"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	karpenterv1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+
+	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
+	"github.com/kaito-project/kaito/pkg/utils/consts"
+	"github.com/kaito-project/kaito/pkg/utils/nodeclaim"
+)
+
+const (
+	gcInterval = 10 * time.Second
+)
+
+// LegacyNodeClaimGCController periodically scans for and deletes orphaned
+// legacy gpu-provisioner NodeClaims that have been superseded by Ready
+// Karpenter-managed NodeClaims for the same workspace.
+type LegacyNodeClaimGCController struct {
+	client   client.Client
+	recorder record.EventRecorder
+	interval time.Duration
+}
+
+// NewLegacyNodeClaimGCController creates a new GC controller.
+func NewLegacyNodeClaimGCController(c client.Client, recorder record.EventRecorder) *LegacyNodeClaimGCController {
+	return &LegacyNodeClaimGCController{
+		client:   c,
+		recorder: recorder,
+		interval: gcInterval,
+	}
+}
+
+type workspaceKey struct {
+	name      string
+	namespace string
+}
+
+func (c *LegacyNodeClaimGCController) Start(ctx context.Context) error {
+	ticker := time.NewTicker(c.interval)
+	defer ticker.Stop()
+
+	klog.InfoS("Legacy NodeClaim GC controller started", "interval", c.interval)
+
+	for {
+		select {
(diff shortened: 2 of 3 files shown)


2170chm:
> ## PR Reviewer Guide 🔍
> Here are some key observations to aid the review process:
> 
> ⏱️ **Estimated effort to review**: 3 🔵🔵🔵⚪⚪
> 🧪 **PR contains tests**
> 🔒 **No security concerns identified**
> ⚡ **Recommended focus areas for review**
> 
> [**Leader Election**](https://github.com/kaito-project/kaito/pull/2009/files#diff-622797ccaf108961818663aa1074497d35dd2e5cc1356b1e2c5214a268a02b6eR297-R306)
> The `LegacyNodeClaimGCController` is added to the manager via `mgr.Add(gcController)` without implementing the `LeaderElectionRunnable` interface. This means the GC cycle will run on all controller replicas, not just the leader, leading to duplicate API calls and event recording. It should implement `NeedLeaderElection() bool` returning true to ensure only the leader runs the background loop.
> 
> ```go
> if consts.IsKarpenterProvisioner() {
> 	gcController := gc.NewLegacyNodeClaimGCController(
> 		kClient,
> 		mgr.GetEventRecorderFor("legacy-gc-controller"),
> 	)
> 	if err = mgr.Add(gcController); err != nil {
> 		klog.ErrorS(err, "unable to add controller", "controller", "LegacyNodeClaimGC")
> 		exitWithErrorFunc()
> 	}
> }
> ```
> 
> [**GC Interval Frequency**](https://github.com/kaito-project/kaito/pull/2009/files#diff-f01a7638b171fe4f1f2d3f404afa77d52ce4afb803bf989ab0ffc1361a0849c6R35-R37)
> The `gcInterval` is set to 10 seconds. For a controller that lists NodeClaims, Pods, and Workspaces on every cycle, this frequency may generate significant load on the Kubernetes API server, especially in larger clusters. Consider increasing this interval (e.g., to 1 or 5 minutes) to reduce API pressure while still ensuring timely cleanup.
> 
> ```go
> const (
> 	gcInterval = 10 * time.Second
> )
> ```
> 
> [**No Pods Logic**](https://github.com/kaito-project/kaito/pull/2009/files#diff-f01a7638b171fe4f1f2d3f404afa77d52ce4afb803bf989ab0ffc1361a0849c6R173-R175)
> In `arePodsOnKarpenterNodes`, if `len(podList.Items) == 0`, the function returns `false`, preventing GC even if the Workspace is `Succeeded`. While conservative, this may leave legacy NodeClaims orphaned if a Workspace object persists but has no running pods (e.g., completed Job). Verify if this behavior aligns with the desired cleanup strategy for such edge cases.
> 
> ```go
> if len(podList.Items) == 0 {
> 	return false, nil
> }
> ```

1. Leader-only is addressed. 

2. I think the load is small and 10s is agreed in the proposal.

3. If we return `true` on workspace on 0-pods case, we risk deleting the legacy pods when the pods are just being recreated since 0-pods can be a transient state. The orphaned nodeclaims will be GCed once the workspace is deleted.

cmd/workspace/main.go
```diff
@@ -292,6 +293,17 @@ func main() {
 				exitWithErrorFunc()
 			}
 		}
+
+		if consts.IsKarpenterProvisioner() {
```
zhuangqh, line 297: is gc controller also guarded by FeatureFlagEnableInferenceSetController?

pkg/controllers/gc/controller.go
```diff
@@ -0,0 +1,294 @@
+// Copyright (c) KAITO authors.
```
zhuangqh, line 1: `pkg/controllers/gc/` this package is too generic. prefer a more specific naming, like nodeclaimgc.
and I suggest move `pkg/controllers` to `pkg/nodeprovision/controllers` @rambohe-ch WDYT

pkg/controllers/gc/controller.go
```diff
@@ -0,0 +1,294 @@
...
+	// 1. List all legacy NodeClaims (labeled kaito.sh/workspace).
+	legacyList := &karpenterv1.NodeClaimList{}
+	if err := c.client.List(ctx, legacyList,
+		client.HasLabels{kaitov1beta1.LabelWorkspaceName},
+	); err != nil {
+		return err
+	}
+
+	if len(legacyList.Items) == 0 {
+		return nil
```
rambohe-ch, line 99: It seems if there is no legacy nodeclaims, we can exit this controller.

zhuangqh (changes requested):
This is a dangerous code path. Honestly, many production incidents were caused by this kind of GC.
I suggest implement in this way: 
maintain a tobedeleted queue, use 1 thread to collect gc items, use a sperate thread to do real deletion, apply strict rate limit to these deletion requests.

pkg/controllers/gc/controller.go
```diff
@@ -0,0 +1,294 @@
...
+
+	// 3. For each workspace group, check Karpenter replacements and delete if ready.
+	var firstErr error
+	for wsKey, legacyNodeClaims := range grouped {
+		if err := c.processWorkspace(ctx, wsKey, legacyNodeClaims); err != nil {
+			klog.ErrorS(err, "Failed to process workspace for GC",
+				"workspace", klog.KRef(wsKey.namespace, wsKey.name))
+			if firstErr == nil {
+				firstErr = err
+			}
```
rambohe-ch, line 129: we can include all errors in a slice, and combine them to return.

pkg/controllers/gc/controller.go
```diff
@@ -0,0 +1,294 @@
...
+		Namespace: wsKey.namespace,
+	}, ws); err != nil {
+		if !errors.IsNotFound(err) {
+			return err
+		}
+		// Workspace deleted — legacy NodeClaims are definitively orphaned.
+		wsFound = false
+	}
+
+	if wsFound {
```
rambohe-ch, line 248: if workspace.DeletionTimestamp is not nil, we can also delete legacy nodeclaims directly.

pkg/controllers/gc/controller.go
Fei-Guo: +1.

PR #2020, opened 2026-05-04
feat: support vllm inference for GGUF models

### LINKED ISSUE #2002: Support quantized models (GGUF and AWQ formats)

Currently, KAITO lacks first-class support for quantized models. Users who want to run smaller, more efficient model variants (e.g., GGUF or AWQ quantized models) cannot easily do so through the existing workflow. This limits KAITO's ability to serve models on lower-end GPU SKUs and increases cost for workloads that don't require full-precision inference.

Add comprehensive quantization support across the KAITO stack, covering the following areas:

### 1. Auto-detect quantization format from model's config.json

- When a user provides a quantized model (e.g., via HuggingFace model ID or local path), KAITO should automatically parse the model's `config.json` to detect the quantization format (GGUF, AWQ, etc.) and bit-width.
- Use the parsed quantization metadata throughout the stack (estimator, runtime) without requiring users to manually specify it.

### 2. Model estimator should account for quantization when calculating KV cache size

- Update the model resource estimator to factor in quantization format and bit-width when computing KV cache memory requirements.
- Quantized models have smaller weight memory footprints, which leaves more GPU memory available for KV cache; the estimator should reflect this.
- Ensure GPU SKU recommendation logic accounts for the reduced memory requirements of quantized models.

### 3. vLLM runtime should support GGUF and AWQ formats

- Update the vLLM runtime configuration to correctly load and serve GGUF-format models (via `--model` flag or equivalent).
- Update the vLLM runtime configuration to correctly load and serve AWQ-format models (e.g., passing `--quantization awq`).
- Ensure the runtime entrypoint selects the correct arguments based on the quantization format detected from the model's config.json.

**Additional context**

- GGUF is widely used in the llama.cpp ecosystem and is increasingly supported by vLLM.
- AWQ is a popular weight-only quantization method with strong vLLM support.
- This feature would significantly reduce the GPU memory and cost requirements for serving large language models in KAITO.

zhehli688:
**Reason for Change**:

GGUF (GPT-Generated Unified Format) is a binary file format for storing large language models (LLMs), introduced by the llama.cpp project and designed for efficient local inference. Currently it has experimental support in vLLM: https://docs.vllm.ai/en/stable/features/quantization/gguf/. **Caveat: there is an ongoing disucssion about whether to keep GGUF support in vLLM: https://github.com/vllm-project/vllm/issues/39583**.

This PR enabled GGUF model deployment in Kaito through the non-catalog path (model catalog support for GGUF models will be added in another PR). Changes include:
- Added preset options `hfConfigPath` and `tokenizer` and passed them to vLLM when available. `hfConfigPath` supports user-provided path to HF model config and is used when a GGUF model is missing config.json. `tokenizer`supports user-provided path to base model's tokenizer as suggusted by vLLM (https://docs.vllm.ai/en/stable/features/quantization/gguf/). 
- Allowed quantization type of GGUF models to be provided as model name suffix, e.g. Qwen/Qwen3-4B-GGUF:Q4_K_M. This is consistent with how model name and quantization type should be passed to vLLM.
- Refactored GetModelByName to preserve the original case of model repo ID for GGUF models (i.e. Qwen/Qwen3-4B-GGUF:Q4_K_M instead of qwen/qwen3-4b-gguf:q4_k_m) and avoid model resolvement failures (this may be a bug in vLLM given GGUF support is experimental).
- Updated _resolve_processor() in benchmark_entrypoint.py to scan default HF cache as well. This is needed when the original HF repo doesn't contain tokenizer.json (e.g. unsloth/Qwen3-8B-GGUF) and the model needs to use a tokenizer from another repo. Since benchmark_entrypoint.py is part of Kaito's base image, we need to bump base image version to 0.3.1.

**Requirements**

- [x] tested deployment of models Qwen/Qwen3-4B-GGUF:Q4_K_M (missing config.json) and unsloth/Qwen3-8B-GGUF:Q4_K_M:
Screenshot 2026-05-01 at 4 37 41 PM
  

**Issue Fixed**:

Fixes #2002

**Notes for Reviewers**:

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -377,10 +377,14 @@ func (p *PresetParam) buildVLLMInferenceCommand(rc RuntimeContext) []string {
 		p.VLLM.ModelRunParams["enable-lora"] = ""
 	}
 	if p.DownloadAtRuntime {
-		repoId, revision, _ := utils.ParseHuggingFaceModelVersion(p.Version)
-		p.VLLM.ModelRunParams["model"] = repoId
-		if revision != "" {
-			p.VLLM.ModelRunParams["code-revision"] = revision
+		// Only set --model from Version URL if not already set by the generator
+		// (e.g., GGUF models set --model to repo_id:quant_type in FinalizeParams).
+		if _, hasModel := p.VLLM.ModelRunParams["model"]; !hasModel {
+			repoId, revision, _ := utils.ParseHuggingFaceModelVersion(p.Version)
+			p.VLLM.ModelRunParams["model"] = repoId
+			if revision != "" {
+				p.VLLM.ModelRunParams["code-revision"] = revision
+			}
 		}
 		p.VLLM.ModelRunParams["download-dir"] = utils.DefaultWeightsVolumePath
 	}
--- a/presets/workspace/generator/generator.go
+++ b/presets/workspace/generator/generator.go
@@ -42,6 +42,7 @@ var (
 	safetensorRegex = regexp.MustCompile(`.*\.safetensors`)
 	binRegex        = regexp.MustCompile(`.*\.bin`)
 	mistralRegex    = regexp.MustCompile(`consolidated.*\.safetensors`)
+	ggufRegex       = regexp.MustCompile(`.*\.gguf`)
 	// source: https://github.com/vllm-project/vllm/blob/main/vllm/reasoning/__init__.py
 	reasoningParserModeNamePrefixMap = map[string]string{
 		"deepseek-r1":  "deepseek_r1",
@@ -251,10 +252,13 @@ var (
 
 type Generator struct {
 	ModelRepo      string
+	QuantType      string // GGUF quantization type (e.g., "Q4_K_M"); empty for non-GGUF models
 	Token          string
 	Param          model.PresetParam
 	CatalogData    []byte // Optional embedded catalog YAML
+	IsGGUFModel    bool
 	IsMistralModel bool
+	BaseModel      string
 
 	// Analyzed params
 	LoadFormat    string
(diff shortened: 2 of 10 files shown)


api/v1beta1/workspace_types.go
```diff
@@ -75,6 +75,16 @@ type PresetOptions struct {
 	// ModelAccessSecret is the name of the secret that contains the huggingface access token.
 	// +optional
 	ModelAccessSecret string `json:"modelAccessSecret,omitempty"`
+	// HFConfigPath specifies an alternate HuggingFace repo ID to fetch config.json from.
+	// This is needed for GGUF model repos that do not include a config.json file.
+	// When set, --hf-config-path is passed to vLLM at runtime. Example: "Qwen/Qwen3-4B".
+	// +optional
+	HFConfigPath string `json:"hfConfigPath,omitempty"`
```
andyzhangx: do you have example of this field? in L80, Example  "Qwen/Qwen3-4B" is the hf model card id, which is same as  ModelName

api/v1beta1/workspace_types.go
```diff
@@ -75,6 +75,16 @@ type PresetOptions struct {
...
+	// Tokenizer specifies a HuggingFace repo ID to use as the tokenizer source.
+	// This is useful for GGUF models whose embedded tokenizer may not load correctly.
+	// When set, --tokenizer is passed to vLLM at runtime. Example: "Qwen/Qwen3-4B".
+	// +optional
+	Tokenizer string `json:"tokenizer,omitempty"`
```
andyzhangx: do you have example of this field? `Example: "Qwen/Qwen3-4B"` is wrong here?

PR #2021, opened 2026-05-05
feat: onboard Gemma 4 models to model catalog

zhehli688:
**Reason for Change**:

Added the following models to Kaito's model catalog:
- google/gemma-4-31B-it
- google/gemma-4-26B-A4B-it
- google/gemma-4-E4B-it
- google/gemma-4-E2B-it

These models are not supported by LMCache yet so added their architecture to isVLLMHybridKVCacheManagerRequired().

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


presets/workspace/generator/generator.go
```diff
@@ -142,6 +143,7 @@ var (
 		"Glm4MoeForCausalLM":                     "glm45",
 		"Glm47MoeForCausalLM":                    "glm47",
 		"Gemma3ForCausalLM":                      "functiongemma",
+		"Gemma4ForConditionalGeneration":         "functiongemma",
```
zhuangqh: I think it's gemma4. https://github.com/vllm-project/vllm/blob/main/vllm/tool_parsers/__init__.py#L181C6-L181C12

presets/workspace/generator/generator.go
zhuangqh, line 81: add update reasoning parser as well. https://github.com/vllm-project/vllm/blob/main/vllm/reasoning/__init__.py#L51

zhuangqh (approved):

PR #2022, opened 2026-05-05
feat: Add guardrails fail-open config and policy ConfigMap mounting for RAGEngine

xiaoqi-7:
## PR Title

Wire RAGEngine guardrails policy ConfigMap and hardcode fail-closed behavior

## PR Description

This PR updates the v1beta1 RAGEngine guardrails wiring to mount the policy
ConfigMap into the pod and to keep guardrails failure handling fail-closed for
now.

### Changes

- Keep the v1beta1 guardrails CRD surface minimal with:
  - `spec.guardrails.enabled`
  - `spec.guardrails.configMapRef`
- Inject guardrails-related environment variables into RAGEngine pods:
  - `OUTPUT_GUARDRAILS_ENABLED` from `spec.guardrails.enabled`
  - `OUTPUT_GUARDRAILS_POLICY_PATH` when `spec.guardrails.configMapRef` is configured
- Mount the referenced guardrails policy ConfigMap as a read-only volume when
  `spec.guardrails.configMapRef` is set.
- Defer a user-facing `failOpen` API for now and hardcode output guardrails to
  fail closed when guardrails execution fails.
- Update tests and generated CRD/schema artifacts to match the current behavior.

### Scope

This change is intentionally limited to v1beta1; v1alpha1 is not modified.


api/v1beta1/ragengine_types.go
```diff
@@ -29,6 +29,12 @@ type GuardrailsSpec struct {
 	// ConfigMapRef points to a ConfigMap that contains a guardrails.yaml policy document.
 	// +optional
 	ConfigMapRef *ConfigMapReference `json:"configMapRef,omitempty"`
+	// FailOpen controls runtime behavior when the guardrails subsystem itself fails
+	// (e.g. scanner model load error, GPU OOM). When true (default), the response is
+	// returned unscanned. When false, the request fails with HTTP 500. This has no
+	// effect on normal redact/block decisions made by a healthy guardrails pipeline.
```
Fei-Guo: "This has no effect on normal redact/block decisions made by a healthy guardrails pipeline."   Don't quite understand this. What are normal redact/block decisons, and what is health guardrails pipeline.

api/v1beta1/ragengine_types.go
Fei-Guo, line 21: Btw, if there is only one Name field, should we just change line 31 to ConfigMapName or ConfigMap with *string?

Fei-Guo (approved):

PR #2025, opened 2026-05-05
feat: hot-reload output guardrails policy via watchfiles

xiaoqi-7:
## Summary

Hot-reload the RAG runtime's output guardrails policy on ConfigMap changes, without requiring a pod restart.

## How

- [x] Add `GuardrailsReloader` to watch the policy file via `watchfiles`.
- [x] Watch the parent directory to catch kubelet's atomic `..data` symlink swap for ConfigMap updates.
- [x] Wire the reloader into FastAPI startup and shutdown, and read guardrails through `guardrails_reloader.current`.
- [x] Make `OutputGuardrails` immutable at the outer object level so reload publishes a new config snapshot instead of mutating the active one in place.

## Semantics

- [x] Failed reload keeps the previous policy active.
- [x] Unchanged policy results in a no-op reload.
- [x] Reload uses a 60-second debounce window.
- [x] Hot reload can be disabled via `OUTPUT_GUARDRAILS_HOT_RELOAD_ENABLED`.

## Config

- [x] `OUTPUT_GUARDRAILS_HOT_RELOAD_ENABLED` (default: `true`)

## Metrics

- [x] `guardrails_policy_reload_total{result="success"|"failure"|"noop"}`
- [x] `guardrails_policy_loaded_timestamp_seconds`

## Tests

- [x] Added reloader unit coverage and updated guardrails/API tests.
- [x] Verified with `pytest presets/ragengine/tests/guardrails/test_reload.py presets/ragengine/tests/guardrails/test_output_guardrails.py presets/ragengine/tests/api/test_chat_completions.py -q`
- [x] Verified with `ruff check .`
- [x] Verified with `ruff format --check .`

## Compatibility

- [x] Backward compatible.
- [x] Policy updates now apply automatically after the debounce window instead of requiring a restart.


presets/ragengine/guardrails/reload.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+            return
+
+        if new_instance == self._current:
+            guardrails_policy_reload_total.labels(
+                **{"result": RELOAD_RESULT_NOOP}
+            ).inc()
+            return
+
+        self._current = new_instance
```
Fei-Guo: Do you need a lock to protect _current?  what if a client is referencing _current and at the same time it is overwritten by reload?

xiaoqi-7: I updated `OutputGuardrails` to be immutable at the outer object level by using `@dataclass(frozen=True)` and storing `scanner_configs` as a tuple instead of a list.

Since hot reload publishes a new instance via whole-object swap, this makes the design safer by preventing accidental in-place mutation after publish.

Fei-Guo (approved):

Fei-Guo (approved):

Issue #1980, opened 2026-04-22
Feature Request: Auto-Upgrade Support for Base Serving Image in InferenceSet


zhuangqh:
## Background

KAITO users running inference workloads in production currently have no built-in mechanism to upgrade the underlying serving stack (base image, vLLM runtime) without manual intervention and downtime.

This matters because:

1. **CVE remediation** — Security vulnerabilities in base images (e.g., Ubuntu, Python, CUDA) or the vLLM runtime require timely patching. Today, users must manually redeploy to address CVEs, leaving a window of exposure.
2. **Runtime improvements** — New vLLM releases bring performance optimizations (better batching, memory management, kernel improvements) that directly translate to higher throughput and lower latency. Users should be able to benefit from these improvements seamlessly.

> Note: This proposal is scoped to the **base serving image** (vLLM runtime + system dependencies), **not** model weights. Model weights are managed separately and are not part of this upgrade path.

### `autoUpgrade` Field in InferenceSetSpec

Add an `autoUpgrade` field to `InferenceSetSpec` that enables the controller to automatically detect and roll out new base image versions using a canary release strategy:

```go
type AutoUpgradePolicy struct {
    // Enabled controls whether the controller automatically upgrades
    // Workspace replicas when a new base image version is available.
    // +optional
    // +kubebuilder:default:=false
    Enabled bool `json:"enabled"`

    // Channel specifies the upgrade channel.
    // "stable" — only stable, tested releases
    // "latest" — most recent release
    // +optional
    // +kubebuilder:default:="stable"
    // +kubebuilder:validation:Enum=stable;latest
    Channel string `json:"channel,omitempty"`

    // PollInterval specifies how often the controller checks for new versions.
    // Defaults to 24h.
    // +optional
    PollInterval *metav1.Duration `json:"pollInterval,omitempty"`
}

type InferenceSetSpec struct {
    // ... existing fields ...

    // AutoUpgrade configures automatic base image upgrade behavior.
    // When enabled, the controller periodically checks for new base image versions
    // and performs a canary rollout to upgrade Workspace replicas with zero downtime.
    // +optional
    AutoUpgrade *AutoUpgradePolicy `json:"autoUpgrade,omitempty"`
}
```

When `autoUpgrade.enabled` is `true`, the InferenceSet controller:

1. Periodically checks for new base image versions based on the configured channel and poll interval.
[...]

PR #2036, opened 2026-05-08
proposal: support auto upgrade for Kaito base image 

zhehli688:
**Reason for Change**:

https://github.com/kaito-project/kaito/issues/1980

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/docs/proposals/20260507-auto-upgrade-base-image.md
+++ b/docs/proposals/20260507-auto-upgrade-base-image.md
@@ -6,7 +6,7 @@ reviewers:
   - "@Fei-Guo"
   - "@zhuangqh"
 creation-date: 2026-05-07
-last-updated: 2026-05-15
+last-updated: 2026-05-19
 status: provisional
 ---
 
@@ -334,10 +334,11 @@ auto-upgrade feature adds the following changes:
 
 **Key Implementation Notes:**
 
-1. **Parallel Pod Management:** The StatefulSet uses `podManagementPolicy: Parallel`,
-   which means all pods are updated simultaneously during a rolling update — not
-   sequentially by ordinal. This eliminates the need to update existing readiness
-   probes for Worker nodes in distributed inference.
+1. **Sequential Pod Updates:** Despite `podManagementPolicy: Parallel`, the
+   StatefulSet controller still updates pods **one at a time in reverse ordinal
+   order** during a `RollingUpdate`. The `Parallel` policy only affects initial
+   creation and scaling — not rolling updates. This means the worker (pod-1) is
+   updated before the leader (pod-0), which requires the worker probe fix below.
 
 2. **Crash Safety:** The rollout state is fully recoverable from existing state:
    - Workspaces with `kaito.sh/upgrade-to-version` set → upgrade in progress.
@@ -346,6 +347,49 @@ auto-upgrade feature adds the following changes:
    - Workspaces whose StatefulSet image tag differs → needs upgrade.
    If the controller crashes mid-rollout, the next reconcile reconstructs the state
    and resumes.
+
+### Probe Changes
+
+For multi-node distributed inference, the worker pod's startup
+and readiness probes must **not** depend on the leader's `/health` endpoint. During a
+rolling update, the worker (pod-1) is updated first. If the worker's probes check the
+leader's health, a permanent deadlock occurs:
+
+- Worker (new image, potentially new Ray version) cannot join leader (old image, old Ray version)
+- Leader cannot serve `/health` without a worker in its placement group
+- Worker never becomes Ready → StatefulSet never updates leader
+
+**Fix:** Worker probes exit 0 unconditionally. This is safe because:
+- The Service routes traffic only to pod-0 (via `statefulset.kubernetes.io/pod-name` selector)
+- Workers don't expose an HTTP endpoint; their readiness is irrelevant to traffic routing
+- The leader's probes still gate actual service availability
+
+We will update the existing probe commands with a `POD_INDEX` conditional:
+
+```bash
+# Startup probe (benchmark variant)
+if [ "$POD_INDEX" = "0" ]; then
+  python3 /workspace/vllm/benchmark_entrypoint.py
+else
+  true
+fi
+
+# Readiness/liveness probes (non-benchmark)
+if [ "$POD_INDEX" = "0" ]; then
+  python3 /workspace/vllm/multi-node-health-check.py readiness --leader-address=... --vllm-port=5000
+else
+  true
+fi
+```
+
+The `multi-node-health-check.py` script will also enforce this at runtime:
+
+```python
+elif args.probe == "readiness":
+    if not is_leader:
+        sys.exit(0)  # Worker: always ready
+    readiness(args)  # Leader: check /health
+```
 
 ### Failure Handling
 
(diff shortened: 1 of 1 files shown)


docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,810 @@
...
+
+## Motivation
+
+### Why Auto-Upgrade?
+
+1. **CVE remediation** — Security vulnerabilities in base images (Ubuntu, Python, CUDA)
+   or the vLLM runtime require timely patching. Today, users must manually rebuild and
+   redeploy to address CVEs, leaving a window of exposure.
+
+2. **Runtime improvements** — New vLLM releases bring performance optimizations (better
```
andyzhangx: this would easily break if we only upgrade the base image since the new kaito-base image would also require other config changes. and for CVE remediation, actually we only need to auto upgrade kaito-base image patch version, e.g. from 0.3.1 to 0.3.2, that would be safer, while upgrade from 0.3.1 to 0.7.0 may introduce inference service break risk.

zhehli688: Our plan is to have two channels: stable (default) and latest. For `stable`, auto-upgrade only applies to patch version (e.g. from 0.3.1 to 0.3.2) as you mentioned. For `latest`, auto-upgrade will halt if a new workspace fails to reach the InferenceReday state. Will this be sufficient to prevent inference service breaks?

andyzhangx: that works for me, and I think you need to mention these details in the design doc PR, e.g. stable is more recommended, and latest may break

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,782 @@
...
+
+The controller queries the OCI registry directly using the [OCI Distribution Spec
+tags/list API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#content-discovery)
+to enumerate available tags for the `kaito-base` image.
+
+**How it works:**
+1. Controller calls `GET /v2/kaito-base/tags/list` against the configured registry.
+2. Tags are parsed as semver (e.g., `0.3.0`, `0.3.1`, `0.4.0`).
+3. Based on the channel:
+   - `stable`: selects the latest tag matching the current major version (e.g., `0.x.y`).
```
andyzhangx: should only allow 0.x.y --> 0.x.z version (patch version) upgrade in stable channel?

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,529 @@
...
+```
+
+#### New Status Conditions
+
+| Condition Type | Status | Reason | Meaning |
+|---|---|---|---|
+| `AutoUpgradeAvailable` | `True` | `NewVersionDetected` | Base image version mismatch detected between controller and Workspaces |
+| `AutoUpgradeInProgress` | `True` | `RollingUpgrade` | One or more Workspaces are being upgraded |
+| `AutoUpgradeInProgress` | `False` | `UpgradeComplete` | All Workspaces upgraded successfully |
+| `AutoUpgradeFailed` | `True` | `WorkspaceUpgradeFailed` | A Workspace failed to become ready after upgrade |
```
Fei-Guo: Only keep AutoUpgradFailed condition.

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,529 @@
...
+
+#### New Status Fields
+
+```go
+type AutoUpgradeStatus struct {
+    // CurrentVersion is the base image tag hardcoded in the running controller.
+    CurrentVersion string `json:"currentVersion,omitempty"`
```
Fei-Guo: Only keep

NumDriftedWorkspace 
LastSuccessfulUpgradeTime

in status.

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,529 @@
...
+Workspaces by the `inferenceset.kaito.io/name` label, creates new Workspaces if
+below `spec.replicas`, and deletes excess Workspaces (non-ready first). The
+auto-upgrade feature adds the following responsibilities:
+
+1. **Version mismatch detection (every reconcile):**
+   After listing Workspaces, the controller fetches each Workspace's StatefulSet and
+   extracts the container image tag. It compares this against the controller's
+   embedded version (`metadata.MustGet("base").Tag`). This check runs unconditionally
+   — even when `autoUpgrade` is not enabled — so that the `AutoUpgradeAvailable`
+   condition always reflects the true state.
```
Fei-Guo: — even when `autoUpgrade` is not enabled — so that the `AutoUpgradeAvailable`
   condition always reflects the true state.

This is not needed, controller only needs to update numofdriftedworkspace.

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,529 @@
...
+
+2. **Sequential Workspace tagging:**
+   The controller selects **one** un-upgraded Workspace that does not already have
+   the `kaito.sh/upgrade-to-version` label and adds it. It does not tag the next
+   Workspace until the current one completes (label removed by Workspace controller)
```
Fei-Guo: I think we should purse not deleting the label in workspace.

In the workspace controller, the logic should be 
if label.AutoUpdate == true && label.upgrade-to-version == GetImageVersion() {
    upgrade()
)

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,529 @@
...
+This means a controller upgrade (which changes the embedded base image tag but not
+the Workspace spec revision) does **not** trigger a StatefulSet update. The
+auto-upgrade feature adds the following changes:
+
+1. **Upgrade label detection:**
+   Before the revision-match early return, `applyInference()` checks for the
+   `kaito.sh/upgrade-to-version` label on the Workspace. If present, the method
+   proceeds with the StatefulSet update regardless of whether the revision matches.
+
+2. **Pre-download new base image:**
```
Fei-Guo: I think prewarming should be done in the inferenceset controller.

docs/proposals/20260507-auto-upgrade-base-image.md
```diff
@@ -0,0 +1,515 @@
...
+- **All up-to-date**: `numDriftedWorkspaces == 0`
+
+#### New Status Fields
+
+```go
+type AutoUpgradeStatus struct {
+    // NumDriftedWorkspaces is the number of Workspaces whose base image version
+    // differs from the controller's embedded version. When 0, all Workspaces are
+    // up-to-date.
+    NumDriftedWorkspaces int `json:"numDriftedWorkspaces"`
```
Fei-Guo, line 191: NumDriftedWorkspaces int, maybe we can make it a pointer type. 
NumDriftedWorkspaces *int, so if autoupgrade is disabled we clear this field instead of assigning a 0 value which can be ambiguous. LastSuccessfulUpgradeTime is still useful even when autoupgrade is disabled so we can alway keep the AutoUpgradeStatus once it is enabled.

Fei-Guo (approved):
Super clear proposal. Thanks for working on this.

PR #2058, opened 2026-05-21
docs: model mirror proposal

2170chm:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


docs/proposals/20260520-model-mirror.md
```diff
@@ -0,0 +1,463 @@
...
+- `spec.source` is generic — not tied to any specific model registry
+- `status.storageURI` is provider-specific (az://, s3://) and is what vLLM receives as `--model=`
+- `status.pvcName` is auto-generated — consumers read from status, not spec
+- Phase never reaches "Failed" — CR stays in `Downloading` and keeps retrying indefinitely
+
+### ModelPrefetch Controller
+
+A new controller that watches `ModelPrefetch` CRs and manages the download lifecycle:
+
+1. **On CR creation:**
```
andyzhangx, line 216: this would depend on enabling blob csi driver first? would you clarify somewhere in the doc?

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,463 @@
...
+- `status.storageURI` is provider-specific (az://, s3://) and is what vLLM receives as `--model=`
+- `status.pvcName` is auto-generated — consumers read from status, not spec
+- Phase never reaches "Failed" — CR stays in `Downloading` and keeps retrying indefinitely
+
+### ModelPrefetch Controller
+
+A new controller that watches `ModelPrefetch` CRs and manages the download lifecycle:
+
+1. **On CR creation:**
+   - Create StorageClass (provider-specific, if not exists)
```
andyzhangx: on AKS managed blob csi driver (disable by default), it would be `blob-fuse`

docs/proposals/20260520-model-mirror.md
```diff
@@ -0,0 +1,463 @@
...
+| Extra config | None | `--model-loader-extra-config '{"distributed": true}'` (TP>1) |
+
+**Environment variables on inference pod:**
+- `AZURE_STORAGE_ACCOUNT_NAME` — resolved dynamically from PVC → PV → volumeHandle
+
+**Pre-requisite:** `runai-model-streamer` and `runai-model-streamer-azure` Python packages pre-installed in the KAITO base image.
+
+### Workload Identity
+
+The inference pod needs to authenticate to blob storage for RunAI streamer's direct `az://` access. This uses AKS Workload Identity.
```
andyzhangx, line 273: this is for inference pod, while for creating blob pvc by blob csi driver, the csi driver controller needs permission to create a storage account in specific resource group, by default, the AKS managed blob csi driver have write permission on AKS node resource group, while for open source driver install, user needs to configure that permission by themselves.

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,506 @@
...
+spec:
+  source:
+    registry: huggingface               # "huggingface", "oci" (future)
+    name: "qwen/qwen2.5-coder-32b-instruct"
+    accessSecret: "hf-token-secret"     # optional, Secret in kaito-workspace namespace
+  storage:
+    storageSize: ""                     # auto-computed from model metadata if empty
+    storageClassName: "blob-nfs"        # required: name of the pre-created StorageClass
+status:
+  phase: Pending | Downloading | Ready
```
Fei-Guo: I think we only need pending and Ready phase.

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,506 @@
...
+
+```yaml
+apiVersion: kaito.sh/v1alpha1
+kind: ModelPrefetch
+metadata:
+  name: qwen--qwen2.5-coder-32b-instruct  # derived from model ID
+spec:
+  source:
+    registry: huggingface               # "huggingface", "oci" (future)
+    name: "qwen/qwen2.5-coder-32b-instruct"
```
Fei-Guo: name -> modelID

This is a cluster scoped resource, hence accessSecret cannot be a string, it needs to be an ObjectReference struct with namespace and name information.

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,506 @@
...
+- `status.pvcName` is auto-generated — consumers read from status, not spec
+- Phase never reaches "Failed" — CR stays in `Downloading` and keeps retrying indefinitely
+
+### ModelPrefetch Controller
+
+A new controller that watches `ModelPrefetch` CRs and manages the download lifecycle:
+
+1. **On CR creation:**
+   - Validate StorageClass exists (error if not — user prerequisite)
+   - Create PVC (auto-sized from model metadata)
```
Fei-Guo: Controller needs to add a finanlizer to the pvc so that it keeps alive even after the download pod is gone.

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,523 @@
...
+  name: qwen--qwen2.5-coder-32b-instruct  # derived from model ID
+spec:
+  source:
+    registry: huggingface               # "huggingface", "oci" (future)
+    name: "qwen/qwen2.5-coder-32b-instruct"
+    accessSecret: "hf-token-secret"     # optional, Secret in kaito-workspace namespace
+  storage:
+    storageSize: ""                     # auto-computed from model metadata if empty
+    storageClassName: "blob-nfs"        # required: name of the pre-created StorageClass
+  idleTTL: ""                           # optional: e.g. "72h". Auto-delete after last workspace removed. Empty = keep forever.
```
Fei-Guo: After a 2nd thought, I think TTL may be misleading here because this is not a Job. It is a delegation resource that control the blob object lifecycle.  We can remove this for now. We keep this resource forever unless user manually deletes it.

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,523 @@
...
+metadata:
+  name: qwen--qwen2.5-coder-32b-instruct  # derived from model ID
+spec:
+  source:
+    registry: huggingface               # "huggingface", "oci" (future)
+    name: "qwen/qwen2.5-coder-32b-instruct"
+    accessSecret: "hf-token-secret"     # optional, Secret in kaito-workspace namespace
+  storage:
+    storageSize: ""                     # auto-computed from model metadata if empty
+    storageClassName: "blob-nfs"        # required: name of the pre-created StorageClass
```
andyzhangx: set `storageClassName: "blob-nfs" ` as default is ok since nfs performs better in this scenario, and I think we could also allow `blob-fuse` or other storage class name as another option since the account key authentication is more flexible, it even works in non Azure environment (while blob-nfs does not).

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,523 @@
...
+### Inference Pod Configuration
+
+When mirror + streaming is active, the inference pod is configured differently from the original path:
+
+| Aspect | Original Path | Streaming Path |
+|--------|---------------|----------------|
+| `--model=` | Local path or HF ID | `az://<container>/<model-path>` (from CR status) |
+| `--load-format=` | `auto` | `runai_streamer` |
+| Model weights volume | Mounted at `/workspace/weights` | Not mounted |
+| Init container | ORAS puller | None |
```
Fei-Guo: Is init container needed for downloading from hf?

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,523 @@
...
+**Pre-requisite:** `runai-model-streamer` and `runai-model-streamer-azure` Python packages pre-installed in the KAITO base image.
+
+### Workload Identity
+
+The inference pod needs to authenticate to blob storage for RunAI streamer's direct `az://` access. This uses AKS Workload Identity.
+
+**User responsibilities (manual, one-time setup per cluster):**
+1. Create or identify a Managed Identity with blob read access
+2. Create a Federated Credential linking the identity to the KAITO ServiceAccount
+3. Grant the identity `Storage Blob Data Reader` role on the storage account
```
Fei-Guo: I think it should be the user's responsibility to create the SA in EVERY namespace they plan to run the workpsace, and create FIC for every such SAs, and add proper annotations before starting the workspace.

docs/proposals/20260520-model-prefetch-streaming.md
```diff
@@ -0,0 +1,523 @@
...
+
+A model download fails repeatedly (e.g. network issues, invalid HF token). The `ModelMirror` CR stays in `Downloading` with `failureMessage` updated. The workspace shows `ModelMirrorInProgress` condition with the error. The CR keeps retrying with exponential backoff. Once the user fixes the issue (e.g. corrects the HF token Secret), the next retry succeeds and both the CR and workspace proceed.
+
+## Implementation Details
+
+### CR Naming Convention
+
+HuggingFace model IDs contain slashes (e.g. `qwen/qwen2.5-coder-32b-instruct`) which are invalid in Kubernetes resource names. The CR name is derived by:
+
+1. Lowercase the model ID
```
Fei-Guo: How about just use a hash of hf modelID as cr name. It seems to be easier. In the CRD printcolumn, we can show the name of the modelID. Hence the UX is not affected too much when user GET all CRDs to find which models have been mirrored.

docs/proposals/20260520-model-mirror.md
```diff
@@ -0,0 +1,523 @@
...
+
+### Feature Gate and Opt-Out
+
+**Feature gate:**
+- Name: `ModelStreaming`
+- Default: `false` (opt-in initially)
+- Passed via `--feature-gates=ModelStreaming=true` on the controller
+
+**Workspace annotation opt-out:**
+- `kaito.sh/model-streaming: "disabled"`
```
Fei-Guo, line 312: Let us add a webhook to make this annotation immutable.  We cannot disable/enable model-streaming during the workspace lifecycle. If model-streaming does not work, user needs to recreate the workspace.

docs/proposals/20260520-model-prefetch-streaming.md
2170chm: No, this is comparing against the original path, just for documenting purpose

Fei-Guo (approved):

Fei-Guo (approved):

PR #2087, opened 2026-06-09
feat: expose model download bandwidth metrics in vLLM pod

zhuangqh:
**Reason for Change**:

Add two Prometheus gauges — kaito_model_download_speed_bytes_per_second and kaito_model_download_remaining_seconds — that are visible at vLLM's /metrics endpoint throughout the pod lifetime.

Key design:
- Pre-download phase (HF repo ID): the pod pre-binds vLLM's port and serves /metrics from a thread-based HTTPServer that runs independently of vLLM's asyncio loop. This ensures Prometheus scrapes are answered even when the event loop is blocked by CUDA init or weight loading. A patched setup_server hands the already-bound socket to vLLM so the port is never released. A patched build_and_serve stops the thread and zeros both gauges just before uvicorn starts accepting connections.

- Local model paths (abs path): the monitor reports 0 immediately so the same metrics are always exposed regardless of model source.

- Bandwidth tracking: ModelDownloadMonitor polls the model's HF Hub cache subdirectory (discovered via scan_cache_dir, not path mangling) every 2 s. It counts only non-symlink files so blobs/ and snapshots/ symlinks are not double-counted. Total expected size is fetched once from HfFileSystem.ls. ETA measures bytes downloaded since monitoring started, not total on disk, to stay correct for partial re-downloads.

Tests (tests/test_download_monitor.py, fully offline via sys.modules stubs and tmp_path) cover:
1. Absolute model path (local weights) — both metrics stay at 0.
2. Model ID downloaded from scratch — speed tracks rolling average, ETA derived from expected_bytes.
3. Interrupted download resumed mid-loop — partial bytes already on disk, completes during monitoring, loop breaks and ETA flips to 0.
4. Model already fully downloaded on startup (no *.incomplete blobs) — ETA set to 0 on the first iteration even when expected_bytes over-counts (e.g. duplicate *.bin + *.safetensors weights).
5. Download stalled (speed stays at 0) with unknown total — remaining time set to -1.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/presets/workspace/inference/vllm/inference_api.py
+++ b/presets/workspace/inference/vllm/inference_api.py
@@ -12,9 +12,15 @@
 # limitations under the License.
 
 import argparse
+import collections
 import logging
 import os
+import socket
+import threading
+import time
 from dataclasses import dataclass
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from pathlib import Path
 from typing import Any
 
 import psutil
@@ -22,8 +28,11 @@ import pynvml
 import uvloop
 import vllm.entrypoints.openai.api_server as api_server
 import yaml
+from huggingface_hub import HfFileSystem, scan_cache_dir
+from prometheus_client import CONTENT_TYPE_LATEST, Gauge, generate_latest
 from vllm.entrypoints.openai.models.protocol import LoRAModulePath
 from vllm.utils.argparse_utils import FlexibleArgumentParser
+from vllm.v1.metrics.prometheus import get_prometheus_registry
 
 # Initialize logger
 logger = logging.getLogger(__name__)
@@ -32,6 +41,21 @@ logging.basicConfig(
     level=logging.DEBUG if debug_mode else logging.INFO,
     format="%(levelname)s %(asctime)s %(filename)s:%(lineno)d] %(message)s",
     datefmt="%m-%d %H:%M:%S",
+)
+
+# Prometheus metrics for model download monitoring.
+# Explicitly registered in vLLM's prometheus registry so they appear at the
+# /metrics endpoint served by vLLM's FastAPI app.
+_registry = get_prometheus_registry()
+kaito_model_download_speed_bytes_per_second = Gauge(
+    "kaito_model_download_speed_bytes_per_second",
+    "Current model download speed in bytes per second",
+    registry=_registry,
+)
+kaito_model_download_remaining_seconds = Gauge(
+    "kaito_model_download_remaining_seconds",
+    "Estimated remaining time for model download in seconds; -1 when unknown",
+    registry=_registry,
 )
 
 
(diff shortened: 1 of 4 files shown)


presets/workspace/inference/vllm/inference_api.py
```diff
@@ -147,6 +171,227 @@ def to_yaml(self) -> str:
...
+    model download, before vLLM's own HTTP server is ready.
+
+    Running in a dedicated thread — completely separate from vLLM's asyncio
+    event loop — ensures Prometheus scrapes are answered even when the event
+    loop is blocked by heavy engine initialisation (CUDA setup, weight loading).
+    """
+
+    def __init__(self, sock_dup: socket.socket) -> None:
+        class _Handler(BaseHTTPRequestHandler):
+            def do_GET(self) -> None:
```
rambohe-ch, line 363: During the download window the proble hits /health and return 200, and K8s marks the pod Ready while the model is still downloading and vLLM isn't serving. Maybe we should restrict the handler to the /metrics endpoint only and return 404 for endpoints else(especially /health).

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -147,6 +171,227 @@ def to_yaml(self) -> str:
...
+) -> int | None:
+    """Return the total size in bytes of all files in *model_id* on HuggingFace Hub.
+
+    Uses HfFileSystem so the result reflects the actual remote file sizes
+    regardless of what is already cached locally.
+    Returns None when the size cannot be determined.
+    """
+    try:
+        fs = HfFileSystem(token=token)
+        entries = fs.ls(model_id, detail=True)
```
rambohe-ch, line 236: Should we consider the files under sub directories?

zhuangqh: model files will be placed in root dir only.

presets/workspace/inference/vllm/inference_api.py
zhuangqh: good catch.

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -12,18 +12,27 @@
...
 import psutil
 import pynvml
 import uvloop
 import vllm.entrypoints.openai.api_server as api_server
 import yaml
+from huggingface_hub import HfFileSystem, scan_cache_dir
+from prometheus_client import CONTENT_TYPE_LATEST, Gauge, generate_latest
 from vllm.entrypoints.openai.models.protocol import LoRAModulePath
 from vllm.utils.argparse_utils import FlexibleArgumentParser
+from vllm.v1.metrics.prometheus import get_prometheus_registry
```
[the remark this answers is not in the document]
zhuangqh: vllm will be ensured to be installed.

presets/workspace/inference/vllm/tests/test_download_monitor.py
```diff
@@ -0,0 +1,462 @@
...
+    "vllm.v1.metrics.prometheus": MagicMock(),
+    "pynvml": MagicMock(),
+    "uvloop": MagicMock(),
+    "psutil": MagicMock(),
+    "prometheus_client": MagicMock(),
+    "yaml": MagicMock(),
+    "huggingface_hub": _HF_MOCK,
+}
+for _name, _mock in _STUBS.items():
+    sys.modules.setdefault(_name, _mock)
```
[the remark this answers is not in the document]
zhuangqh: mock these func here to save time and avoid heavy server launch process.

Fei-Guo (approved):

PR #2088, opened 2026-06-09
docs: proposal for distributed cache integration

croomes:
Introduces a design proposal for pluggable distributed caching in KAITO covering both model weight and KV caching. The design supports independent provider selection per concern, enabling configurations such as Fluid for model weights alongside FlexKV for attention tensors, or a unified provider serving both through shared infrastructure.

**Reason for Change**:

Model loading from cloud storage takes 5-10 minutes for large models (70B+), creating unacceptable cold-start latency during scale-out. Repeated fetches incur significant egress costs. KAITO currently has no caching abstraction. This proposal adds a provider-neutral framework that integrates with ModelMirror to warm caches during download, enabling near-instant model loading from local/distributed cache on subsequent starts.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


zhuangqh (commented):
Thanks for raising up this idea!

several questions come up in my mind

1 When introducing the provider mechanism, what differences do you expect between providers besides storage media?
2 Kaito recently integrated a model streaming feature to achieve GB/s-level model download speeds. Would you recommend using both model streaming and model cache, or should we select one?
3 The kvcache targets a different interface compared to model files and requires integration with the inference engine. Currently, Kaito uses an L1 kvcache, leveraging lmcache to offload kvcache to CPU memory. Are you considering adding an L2 kvcache? Will you provide a client? As I understand it, like lmcache, communication with the backend requires a specific protocol, and direct mounting of the filesystem to pods is not enough. https://docs.lmcache.ai/kv_cache/storage_backends/index.html

croomes:
Hi @zhuangqh, thanks for the questions!

> 1. When introducing the provider mechanism, what differences do you expect between providers besides storage media? 

They'll likely have different deployment topologies and access methods. But I think the design is abstract enough that these should be hidden by their operator. For example, fluid already abstracts multiple cache implementations in a similar way.

> 2. Kaito recently integrated a model streaming feature to achieve GB/s-level model download speeds. Would you recommend using both model streaming and model cache, or should we select one? 

We would still recommend using the run:ai model streamer, providing it has cache integration. We recently added cache support for Azure in https://github.com/run-ai/runai-model-streamer/pull/139, but the same approach can be used for other backends. The current implementation allows loading models from with the cache over RDMA, and we're working on another PR for GPU Direct support that will avoid a copy into CPU memory.

> 3. The kvcache targets a different interface compared to model files and requires integration with the inference engine. Currently, Kaito uses an L1 kvcache, leveraging lmcache to offload kvcache to CPU memory. Are you considering adding an L2 kvcache? Will you provide a client? As I understand it, like lmcache, communication with the backend requires a specific protocol, and direct mounting of the filesystem to pods is not enough. 

Most KV caches will support integration with LMCache or implement the LMCache API as a drop-in replacement. They would use API-level integration, rather than storage-level. I'd expect KV caches to be L2 providers.

docs/proposals/20260609-distributed-cache-integration.md
```diff
@@ -0,0 +1,658 @@
...
+  - Merge all PodMutations (deduplicate env vars if same provider used for both)
+  - Inject merged mutations into model pod specs
+- Add validation webhook rules:
+  - Each sub-config's `mode` must be a valid enum value
+  - Each sub-config's `provider` must be a registered provider
+- Add conditions to `WorkspaceStatus`: `ModelCacheReady`, `KVCacheReady`.
+
+#### Phase 4: ModelMirror Integration (Cache Warming)
+
+Cache warming is a side-effect of ModelMirror's model download, not a separate lifecycle phase. When a cache provider is installed, downloads write through to both persistent storage and the distributed cache simultaneously.
```
zhuangqh, line 596: IIUC, you will allow runaistreamer to download from this distributed cache.
Then blob storage will be hidden by this cache layer, right?

docs/proposals/20260609-distributed-cache-integration.md
```diff
@@ -0,0 +1,658 @@
...
+    // and the provider can operate (e.g., CRD exists, operator running).
+    IsAvailable(ctx context.Context) (bool, error)
+
+    // IsReady reports whether the cache is warmed and ready to serve.
+    // Returns (ready, reason, error).
+    IsReady(ctx context.Context) (bool, string, error)
+
+    // PodMutations returns the pod-level changes needed for a specific cache
+    // concern (ModelWeights or KVCache).
+    PodMutations(ctx context.Context, concern CacheConcern, workspace *kaitov1beta1.Workspace, modelName, modelRevision string) (*PodMutations, error)
```
zhuangqh, line 453: Besides `PodMutations`, will the provider need to provision some resources, such as the Cache CRs you mentioned, configmap?

zhuangqh (commented):
Can you document about the runtime integration details beside k8s API?
like
- modelcache: integrate with runaistreamer client. Not a Huggface API
- kvcache: 
  1. integrate with LMCache, acts as a LMCache custom L2 storage.
  2. replace the entire vllm connector
what do you think? And probably a plugin specifc configmap is required.

docs/proposals/20260609-distributed-cache-integration.md
```diff
@@ -0,0 +1,658 @@
...
+1. The **Cache Backend Operator** (external, pre-installed) watches provider-specific Cache CRs and reconciles cache infrastructure (server pods, discovery service). Installed independently of KAITO.
+2. The **Cache Controller** creates and reconciles the backend's Cache CR from provider configuration (Helm values), and monitors its status for readiness. Scope: cluster-level cache infrastructure only.
+3. The **Workspace Controller** owns the per-workspace cache lifecycle: resolves providers per concern, queries readiness, gates inference pod creation on cache mode, and injects `PodMutations` into model pods.
+4. **Cache warming** occurs during the ModelMirror download — either transparently via a cache-aware CSI driver (StorageClass), or via webhook injection into the download pod. No separate warming Jobs are required.
+5. **Model pods** use injected configuration (env vars, volumes, labels, or KV connector config) to access cache infrastructure.
+
+
+### API Changes
+
+#### CacheSpec (New Top-Level Workspace Field)
```
Fei-Guo: This is going to be the major debate point of the design:

We need to define the sharing scope of the cache (model weights/kv cache). 

1) Only used by a single workspace, the API changes can be done in workspace crd;
2) Shared by multiple workspaces within one InferenceSet, the API changes can be done in InferenceSet crd;
3) Can be shared by multiple InferenceSet; We need to introduce a cluster scoped CRD, and add reference in InferensetSet CRD; 

I think case 2 and case 3 are both valid; It is possible cx use different inferencesets to run the same model. 
If not a cluster scoped crd, I would think cache should be part of InferenceSet API.

docs/proposals/20260609-distributed-cache-integration.md
```diff
@@ -0,0 +1,658 @@
...
+    // +kubebuilder:validation:Enum=Required;Opportunistic;Disabled
+    Mode CacheMode `json:"mode,omitempty"`
+
+    // CleanupOnDelete invalidates cached model data when workspace is deleted.
+    // +optional
+    CleanupOnDelete bool `json:"cleanupOnDelete,omitempty"`
+}
+
+// KVCacheSpec controls how attention KV tensors are cached.
+type KVCacheSpec struct {
```
Fei-Guo, line 350: What are the knobs that user can tune with the KVcache? Intuitively, user should have a way to configure the size of the cache? Basically, I think user needs to provide the desire about how many resources they plan to use for caching the kv.

How does this KVcache interact with the llmcache's kvcache offloading to CPU/local nvme feature? Can they work together or they are completely exclusive.

hasethuraman:
> Hi @zhuangqh, thanks for the questions!
> 
> > 1. When introducing the provider mechanism, what differences do you expect between providers besides storage media?
> 
> They'll likely have different deployment topologies and access methods. But I think the design is abstract enough that these should be hidden by their operator. For example, fluid already abstracts multiple cache implementations in a similar way.
> 
> > 2. Kaito recently integrated a model streaming feature to achieve GB/s-level model download speeds. Would you recommend using both model streaming and model cache, or should we select one?
> 
> We would still recommend using the run:ai model streamer, providing it has cache integration. We recently added cache support for Azure in [run-ai/runai-model-streamer#139](https://github.com/run-ai/runai-model-streamer/pull/139), but the same approach can be used for other backends. The current implementation allows loading models from with the cache over RDMA, and we're working on another PR for GPU Direct support that will avoid a copy into CPU memory.
> 
> > 3. The kvcache targets a different interface compared to model files and requires integration with the inference engine. Currently, Kaito uses an L1 kvcache, leveraging lmcache to offload kvcache to CPU memory. Are you considering adding an L2 kvcache? Will you provide a client? As I understand it, like lmcache, communication with the backend requires a specific protocol, and direct mounting of the filesystem to pods is not enough.
> 
> Most KV caches will support integration with LMCache or implement the LMCache API as a drop-in replacement. They would use API-level integration, rather than storage-level. I'd expect KV caches to be L2 providers.

Thanks @zhuangqh on that LMCache storage backend link. Have pushed a new commit

In this [commit](https://github.com/kaito-project/kaito/pull/2088/changes/3a50608b8a744f34e7532bf586cc9c505a616b59), I've added a Runtime Integration Details section on model streaming behaviour & 2 modes on path to KVCache [mode 1 - https://docs.lmcache.ai/developer_guide/extending_lmcache/storage_plugins.html, mode 2 - VLLM_KV_TRANSFER_CONFIG)

docs/proposals/20260609-distributed-cache-integration.md
hasethuraman: Thats correct @zhuangqh. Have pushed a new section in https://github.com/kaito-project/kaito/pull/2088/changes/64c65395f600a956f4324bbb0e6ce9d3a3cf417f

docs/proposals/20260609-distributed-cache-integration.md
hasethuraman: Agree — in [commit ](https://github.com/kaito-project/kaito/pull/2088/changes/3a50608b8a744f34e7532bf586cc9c505a616b59) added a Cache Scope section comparing the three options. We've chosen Option 2:  CacheSpec  on both  InferenceSetTemplate  and  Workspace , with InferenceSet propagating cache config to child Workspaces at creation time. This covers the dominant use case (all replicas of the same model sharing one cache config) without introducing a new CRD.

For cross-InferenceSet sharing (case 3), a cluster-scoped CRD can be introduced as a non-breaking future addition. For MultiRoleInference (KV cache shared across prefill/decode roles), the controller propagates consistent  kvCache  config to both role InferenceSets.

docs/proposals/20260609-distributed-cache-integration.md
hasethuraman: In the [commit](https://github.com/kaito-project/kaito/pull/2088/changes/3a50608b8a744f34e7532bf586cc9c505a616b59) added a Config  field (ConfigMap reference) to  KVCacheSpec, following the same pattern as  inference.config . Users can supply specific settings via a ConfigMap; the provider validates and merges with its Helm defaults. This keeps the API provider-neutral while giving users to desired control over tuning knobs.

On LMCache part - documented the 2 modes in the Runtime Integration section in the above mentioned commit: 
- In Mode 1, the new L2 distributed KV cache is designed to work together ( https://docs.lmcache.ai/developer_guide/extending_lmcache/storage_plugins.html) with the existing L1 offloading.
- In Mode 2, vLLM uses a different KV connector, so the current L1 offloading path (LMCache) is no longer used by vLLM. The new connector would need to implement similar offloading behavior if it wants to preserve that functionality.

docs/proposals/20260609-distributed-cache-integration.md
hasethuraman: @zhuangqh yes cache CRs will be created by the respective provider. The config map based on the defaults provided during KAITO installation through values.yaml. Provider also create one more config map per https://github.com/kaito-project/kaito/pull/2088#discussion_r3437044174

docs/proposals/20260609-distributed-cache-integration.md
andyzhangx: For reference on LMCache connector configuration, see: https://docs.lmcache.ai/getting_started/quickstart/offload_kv_cache.html

KAITO already supports KV cache offloading to CPU RAM today. This proposal appears to extend that to local filesystem storage. My concern is that filesystem-based KV cache offloading would introduce additional latency compared to GPU memory (and even CPU RAM), potentially degrading inference performance. Before implementing this, we'd need baseline performance benchmarks to validate that the trade-off is worthwhile.

To keep the scope manageable, I'd suggest focusing on ModelCache only in Stage 1 and deferring filesystem-based KV cache offloading until perf data justifies it.

docs/proposals/20260609-distributed-cache-integration.md
```diff
@@ -0,0 +1,844 @@
...
+
+## Motivation
+
+Model loading and inference latency are significant operational pain points for KAITO users:
+
+- **Cold-start times**: Loading a 70B parameter model from cloud storage can take 5-10 minutes. For autoscaled inference workloads, this creates unacceptable latency during scale-out events.
+- **Storage egress costs**: Repeatedly fetching multi-GB model weights from cloud storage incurs significant egress charges, especially across availability zones.
+- **Scale-out penalty**: When new nodes join the cluster, models must be re-downloaded from remote storage, creating a "thundering herd" effect during scaling events.
+- **Redundant computation**: Without KV caching, common prompt prefixes (system prompts, few-shot examples) are recomputed on every request, wasting GPU cycles and increasing time-to-first-token.
+- **Disaggregated inference**: Prefill/decode disaggregation requires a shared KV cache layer to transfer attention state between prefill and decode pods.
```
andyzhangx, line 82: For context, I've already implemented disaggregated prefill/decode (P/D) in KAITO. That feature uses the Nixl connector (--kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_both","kv_load_failure_policy":"fail"}') for GPU-to-GPU KV cache transfer between prefill and decode roles — which is distinct from the LMCache connector approach that offloads KV cache to local disk.

e.g. you could only use one `kv-transfer-config` config, you cannot use both NixlConnector and LMCacheConnectorV1, that's not supported in vllm.

docs/proposals/20260609-distributed-cache-integration.md
```diff
@@ -0,0 +1,844 @@
...
+- **Cache Controller**: A new KAITO controller that bridges workspace semantics with cache infrastructure, managing cache lifecycle and readiness signaling.
+- **PodMutations**: The set of changes (environment variables, volumes, volume mounts, init containers, labels) that a cache provider injects into model pods to enable cache access.
+- **Cache Warming**: The process of populating a cache with model weights, reducing cold-start latency. Occurs when ModelMirror downloader uses a cache-aware storage backend.
+
+## Summary
+
+This proposal adds a pluggable distributed caching framework to KAITO that accelerates both model loading and inference for AI workloads. It addresses two complementary caching concerns:
+
+1. **Model weight caching** — Caches static model files on fast local storage (NVMe, ramdisk) or in a distributed cache layer, reducing cold-start load times from minutes to seconds.
+2. **KV caching** — Caches attention key/value tensors across requests, enabling prefix reuse, faster time-to-first-token, and disaggregated prefill/decode architectures.
```
andyzhangx, line 70: I'd suggest scoping Stage 1 to model weight caching only. That's a well-defined, self-contained goal with clear implementation boundaries and measurable impact.

zhuangqh (approved):
LGTM. Looking forward to this exciting feature!

Fei-Guo (approved):

PR #2100, opened 2026-06-11
feat: model streaming

### LINKED ISSUE #2073: Integrate workspace and inference pod streaming with model mirror

**Additional context**

2170chm:
**Reason for Change**:

1. Integrates ModelMirror (Phase 1) into the workspace controller to enable automated model streaming from blob storage
2. When ModelStreaming=true, the workspace controller auto-creates ModelMirror CRs, downloads models to blob PVCs in parallel with GPU provisioning, then configures inference pods to stream weights directly from az:// via RunAI Model Streamer
3. Adds CloudProvider interface for provider-specific streaming logic (Azure first, extensible to S3/GCS)
4. Adds runai-model-streamer to base image dependencies

Key Changes

- Workspace controller: ensureModelMirror() before reconcileNodes() (parallel download + GPU provisioning), waitForModelMirror() gates inference pod creation

- Inference pod: --model=az://, --load-format=runai_streamer, distributed streaming for TP>1, weights volume skipped, WI label + SA set

- CloudProvider interface: ResolveStreamingConfig(), ValidateServiceAccount(), CSIDriverName().

- Validations: Webhook validates CSI driver on create. Controller validates SC provisioner + SA existence + WI annotation (these validations need helm flag context)

- Benchmark probe: KAITO_PROCESSOR env var (Case 0) resolves tokenizer from HF when streaming because there is no local tokenizer.

- Base image: Added runai-model-streamer==0.16.0 + runai-model-streamer-azure==0.16.0, bumped tag to 0.3.21. 

**Requirements**

**Issue Fixed**:

Closes #2073 
**Notes for Reviewers**:

CI E2E tests will be added in a separate PR after this is merged. The feature is manual + unit tested and guarded by feature flag, safe to merge.

Manual E2E verifications:

- phi-4-mini-instruct streaming (single-node)
- mistral-7b-instruct streaming (single-node. It contains a custom load_format parameter and is correctly overriden)
- llama-3.3-70b-instruct streaming (multi-node TP=2 PP=2, distributed streaming)
- Non-streaming DAR path regression (unchanged behavior)
- Benchmark probe with KAITO_PROCESSOR (via inject script)
- Webhook/controller negative paths (CSI driver missing, wrong SC, missing SA, annotation immutability)
- ModelMirror lifecycle (CR deletion cleanup, workspace re-creates deleted CR)


andyzhangx (commented):
where did you specify the storage class name? I think you need to add some docs for this config

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -81,20 +82,30 @@ type WorkspaceReconciler struct {
...
-		nodeProvisioner: provisioner,
+		Client:                         client,
+		Scheme:                         scheme,
+		Log:                            log,
+		klogger:                        klog.NewKlogr().WithName("WorkspaceController"),
+		Recorder:                       Recorder,
+		expectations:                   expectations,
+		Estimator:                      &nodesestimator.NodeEstimator{},
+		nodeProvisioner:                provisioner,
+		DefaultModelMirrorStorageClass: defaultStorageClass,
```
andyzhangx: what are the default values of `defaultStorageClass` and `defaultServiceAccount`?

api/v1beta1/condition_types.go
```diff
@@ -58,4 +58,8 @@ const (
 	// True means the benchmark completed and results are stored in status.benchmarkResult.
 	// Set by default; omitted when kaito.sh/disable-benchmark is "true".
 	WorkspaceConditionTypeBenchmarkCompleted = ConditionType("BenchmarkCompleted")
+
+	// WorkspaceConditionTypeModelMirrorInProgress indicates the ModelMirror download is in progress.
+	// Set when ModelStreaming feature gate is enabled and the workspace's ModelMirror CR is not yet Ready.
+	WorkspaceConditionTypeModelMirrorInProgress = ConditionType("ModelMirrorInProgress")
```
Fei-Guo: If you check our convention, we don't add a ConditionType for progress report. Instead, we can add a condition like ModelMirrorReady.

api/v1beta1/workspace_types.go
```diff
@@ -215,6 +215,13 @@ type Performance struct {
 	Metrics map[string]Metric `json:"metrics,omitempty"`
 }
 
+// ModelMirrorReference holds a reference to the cluster-scoped ModelMirror CR
+// associated with this workspace's model.
+type ModelMirrorReference struct {
```
Fei-Guo: I think the model mirror name can be derived from the model_ID. We probably do not need this in CRD since the info is duplicated. We should avoid API changes for a feature whose enablement switch is an annotation.

pkg/workspace/inference/modelstreaming.go
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) KAITO authors.
```
Fei-Guo, line 1: Can we rename all files starting with streaming*.go to modelstreaming*.go.

pkg/workspace/controllers/workspace_controller.go
2170chm: There aren't default values. Users are responsible for creating the storage account and service account based on the guide in KAITO doc which I will add after release and pass in the name here. When streaming is enabled, the storage class is required because it is a cluster resource. The service account is not because it is a namespace resource (though if the user is consistent with the naming in all namespace s they can specify it too. They are called defaults because workspace annotation can override both.

It is similar to the gpu labels that are up to the user to decide the naming.

2170chm:
> where did you specify the storage class name? I think you need to add some docs for this config

I haven't specified it yet. This is feature gated so I am thinking of updating the KAITO doc after this is scheduled to release. This feature will require a lot of manual setups from the user.

api/v1beta1/workspace_validation.go
```diff
@@ -93,6 +94,9 @@ func (w *Workspace) Validate(ctx context.Context) (errs *apis.FieldError) {
 			errs = errs.Also(w.Resource.validateCreateWithTuning(w.Tuning).ViaField("resource"),
 				w.Tuning.validateCreate(ctx, w.Namespace).ViaField("tuning"))
 		}
+		if featuregates.FeatureGates[consts.FeatureFlagModelStreaming] {
```
zhuangqh: move this validation to L90? it's inference only.

pkg/model/interface.go
```diff
@@ -376,7 +382,16 @@ func (p *PresetParam) buildVLLMInferenceCommand(rc RuntimeContext) []string {
 	if !p.VLLM.DisallowLoRA && rc.AdaptersEnabled {
 		p.VLLM.ModelRunParams["enable-lora"] = ""
 	}
-	if p.DownloadAtRuntime {
+	// Model source: streaming (az://) vs download-at-runtime (HF repo).
+	if rc.StreamingModelPath != "" {
+		p.VLLM.ModelRunParams["model"] = rc.StreamingModelPath
+		if rc.StreamingLoadFormat != "" {
```
Fei-Guo: Is it an assert or if condition? Can the rc.StreamingLoadFormat be "" at all if rc.StreamingModelPath!="".

pkg/utils/consts/consts.go
```diff
@@ -50,14 +50,27 @@ const (
...
+
+	// CSI driver names for model streaming (workspace controller + webhook scope).
+	CSIDriverNameAzureBlob = "blob.csi.azure.com"
 )
 
+// CSIDriverNameForCloud returns the expected CSI driver name for the given cloud provider.
+// Returns "" for unsupported providers.
+func CSIDriverNameForCloud(cloud string) string {
+	switch cloud {
+	case AzureCloudName, "":
```
Fei-Guo: Should this be `case AzureCloudName:` ?

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -153,6 +164,130 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
...
+		return fmt.Errorf("StorageClass %q not found: %w", storageClass, err)
+	}
+	expectedCSIDriver := c.StreamingCloudProvider.CSIDriverName()
+	if sc.Provisioner != expectedCSIDriver {
+		return fmt.Errorf("StorageClass %q uses provisioner %q, but model streaming requires %q; "+
+			"create a StorageClass with the correct provisioner",
+			storageClass, sc.Provisioner, expectedCSIDriver)
+	}
+
+	// Validate ServiceAccount exists and has workload identity configured.
```
Fei-Guo: A better approach is to make ValidateServiceAccount interface to take the wObj as the input. Then line 209 to 219 can all moved to cloud specific implementation, improving the readability. This also avoids adding cloud specific WI implementation dependency in main controller code.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -153,6 +164,130 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
...
+		if apierrors.IsNotFound(err) {
+			// CR was deleted externally; ensureModelMirror will recreate on next reconcile
+			return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+		}
+		return &reconcile.Result{}, fmt.Errorf("failed to get ModelMirror CR %s: %w", crName, err)
+	}
+
+	if cr.Status.Phase != kaitov1alpha1.ModelMirrorPhaseReady {
+		klog.InfoS("ModelMirror CR not ready, gating inference", "name", crName, "phase", cr.Status.Phase)
+		return &reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
Fei-Guo: If you add a watcher for modelmirror resource, you don't need to requeue every 10 seconds.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -153,6 +164,130 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
 	return nil
 }
 
+// ensureModelMirror creates the ModelMirror CR for the workspace's model if it doesn't exist.
+// Returns nil if the CR exists (any phase) or was created successfully.
+func (c *WorkspaceReconciler) ensureModelMirror(ctx context.Context, wObj *kaitov1beta1.Workspace, model pkgmodel.Model) error {
+	if !inference.ModelStreamingEnabled(wObj) {
```
Fei-Guo: I would suggest moving checks from 170 to line 175 to the caller side.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -183,10 +318,27 @@ func (c *WorkspaceReconciler) addOrUpdateWorkspace(ctx context.Context, wObj *ka
 		return reconcile.Result{}, nil
 	}
 
+	// Ensure ModelMirror CR exists (starts download in parallel with node provisioning).
+	if wObj.Inference != nil && wObj.Inference.Preset != nil {
+		presetName := string(wObj.Inference.Preset.Name)
+		model, err := models.GetModelByName(ctx, presetName, wObj.Inference.Preset.PresetOptions.ModelAccessSecret, wObj.Namespace, c.Client)
```
Fei-Guo: Can you double check? I think when the code reaches here, the model name validity has been checked and line 324 is not needed.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -153,6 +164,130 @@ func (c *WorkspaceReconciler) ensureFinalizer(ctx context.Context, workspaceObj
...
+	}
+
+	klog.InfoS("Created ModelMirror CR", "name", crName, "modelID", modelID, "workspace", klog.KObj(wObj))
+	return nil
+}
+
+// waitForModelMirror checks if the ModelMirror CR is Ready.
+// Returns (nil, nil) to proceed, or (*Result, err) to stop — same pattern as reconcileNodes.
+func (c *WorkspaceReconciler) waitForModelMirror(ctx context.Context, wObj *kaitov1beta1.Workspace) (result *reconcile.Result, err error) {
+	if !inference.ModelStreamingEnabled(wObj) {
```
Fei-Guo: Moves the checks (line 265 to line 271) to the caller side.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -434,7 +586,8 @@ func (c *WorkspaceReconciler) applyInference(ctx context.Context, wObj *kaitov1b
 	}
 
 	revisionStr := wObj.Annotations[kaitov1beta1.WorkspaceRevisionAnnotation]
-	workloadObj, err := inference.GeneratePresetInference(ctx, wObj, revisionStr, model, c.Client)
+	workloadObj, err := inference.GeneratePresetInference(ctx, wObj, revisionStr, model, c.Client,
+		c.DefaultModelMirrorStorageClass, c.DefaultStreamingServiceAccount, c.StreamingCloudProvider)
```
Fei-Guo: Can you try not to change the GeneratePresetInference input parameters? I think 
c.DefaultModelMirrorStorageClass, c.DefaultStreamingServiceAccount, c.StreamingCloudProvider
The default values can be global variables, and you can add a wObj.GetStreamerInfo() to apply overwrite based on wObj annoations. Then here you don't need to change the parameter list.

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -589,6 +697,34 @@ func (c *WorkspaceReconciler) syncWorkspaceStatus(ctx context.Context, key types
 		}
 
 		if wObj.Inference != nil {
+			if inference.ModelStreamingEnabled(wObj) && wObj.Inference.Preset != nil {
+
```
Fei-Guo, line 718: The key status update requirement is that if model mirror is not ready, you cannot mark resource ready in workspace. Please make sure this is the case.

api/v1beta1/workspace_validation.go
zhuangqh: and use `ModelStreamingEnabled` ?

pkg/workspace/inference/modelstreaming.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// ModelMirrorCRName derives the ModelMirror CR name from a HuggingFace model ID.
+func ModelMirrorCRName(modelID string) string {
+	return sha256First6(modelID)
+}
+
+// ResolveHFModelID resolves the HuggingFace model ID from a workspace's preset name.
+// Returns "" if the workspace has no inference preset.
+func ResolveHFModelID(ws *v1beta1.Workspace) string {
```
Fei-Guo, line 64: This looks like a common util func that should be used in other places. Make sure this is not a dup of existing code.

pkg/workspace/inference/streaming.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// buildCommonStreamingEnvVars returns env vars common to all cloud providers:
+// KAITO_PROCESSOR (for benchmark probe) and HF_TOKEN (for vLLM model config resolution).
+func buildCommonStreamingEnvVars(modelID, accessSecretName string) []corev1.EnvVar {
+	envVars := []corev1.EnvVar{
+		{Name: "KAITO_PROCESSOR", Value: modelID},
+	}
+
+	if accessSecretName != "" {
```
zhuangqh: both ModelDownloadInfo and streaming need `HF_TOKEN`
Suggest adding an individual opt for this env.

pkg/workspace/inference/streaming_provider.go
```diff
@@ -0,0 +1,68 @@
...
+
+// CloudProvider abstracts provider-specific logic for resolving streaming configuration
+// from a PVC/PV. To add support for a new cloud provider (e.g. S3, GCS):
+//  1. Create a new file streaming_provider_<cloud>.go
+//  2. Implement the CloudProvider interface
+//  3. Add a case to GetCloudProvider() and consts.CSIDriverNameForCloud()
+type CloudProvider interface {
+	// ResolveStreamingConfig reads the PVC and its backing PV to resolve
+	// the full streaming model path, provider-specific env vars, and pod labels.
+	ResolveStreamingConfig(ctx *generator.WorkspaceGeneratorContext, pvcName, pvcNamespace, modelID string) (*StreamingConfig, error)
```
Fei-Guo: ResolveStreamingConfig -> GetStreamingConfig

pkg/workspace/inference/streaming_provider.go
```diff
@@ -0,0 +1,68 @@
...
+	// (e.g. azure.workload.identity/use for Azure WI; empty for AWS/GCP).
+	PodLabels map[string]string
+}
+
+// CloudProvider abstracts provider-specific logic for resolving streaming configuration
+// from a PVC/PV. To add support for a new cloud provider (e.g. S3, GCS):
+//  1. Create a new file streaming_provider_<cloud>.go
+//  2. Implement the CloudProvider interface
+//  3. Add a case to GetCloudProvider() and consts.CSIDriverNameForCloud()
+type CloudProvider interface {
```
Fei-Guo: The intefacename should be ModelSteamer.

pkg/workspace/inference/streaming_provider.go
```diff
@@ -0,0 +1,68 @@
...
+	// the full streaming model path, provider-specific env vars, and pod labels.
+	ResolveStreamingConfig(ctx *generator.WorkspaceGeneratorContext, pvcName, pvcNamespace, modelID string) (*StreamingConfig, error)
+
+	// CSIDriverName returns the expected CSI driver name for validation.
+	CSIDriverName() string
+
+	// ValidateServiceAccount checks that the SA has the required provider-specific
+	// configuration (e.g. Azure WI client-id annotation, AWS IAM role-arn annotation).
+	// Returns nil if valid, error with actionable message if not.
+	ValidateServiceAccount(sa *corev1.ServiceAccount) error
```
Fei-Guo: To be more generic, ValidateServiceAccount should be ValidateAuth.

pkg/workspace/inference/streaming_provider.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// GetCloudProvider returns the CloudProvider implementation for the given cloud.
+// Currently only Azure is supported. To add a new provider, add a case here
+// and in consts.CSIDriverNameForCloud().
+func GetCloudProvider(cloudName string) (CloudProvider, error) {
+	switch cloudName {
+	case consts.AzureCloudName, "":
```
Fei-Guo: Should this be `case consts.AzureCloudName:`?

pkg/workspace/inference/streaming_provider.go
```diff
@@ -0,0 +1,68 @@
...
+//  1. Create a new file streaming_provider_<cloud>.go
+//  2. Implement the CloudProvider interface
+//  3. Add a case to GetCloudProvider() and consts.CSIDriverNameForCloud()
+type CloudProvider interface {
+	// ResolveStreamingConfig reads the PVC and its backing PV to resolve
+	// the full streaming model path, provider-specific env vars, and pod labels.
+	ResolveStreamingConfig(ctx *generator.WorkspaceGeneratorContext, pvcName, pvcNamespace, modelID string) (*StreamingConfig, error)
+
+	// CSIDriverName returns the expected CSI driver name for validation.
+	CSIDriverName() string
```
Fei-Guo: I don't think CSIDriverName() is used externally, maybe we can remove it from interface.

pkg/workspace/controllers/workspace_controller.go
2170chm: Yes, and this is to get the model's DiskStorageRequirement, I will move the resolution inside ensureModelMirror since it is not used elsewhere.

pkg/workspace/inference/modelstreaming.go
2170chm: We have a GetModelByNameWithToken, but it returns a Model object (full resolution including registry lookup and HF generation). So this makes sense since we only need the name, but I will move it to utils/plugin.

pkg/workspace/controllers/workspace_event_handler.go
```diff
@@ -94,3 +95,35 @@ var enqueueWorkspaceForNodeClaim = handler.EnqueueRequestsFromMapFunc(
...
+			wsList := &kaitov1beta1.WorkspaceList{}
+			if err := kubeClient.List(ctx, wsList); err != nil {
+				klog.ErrorS(err, "failed to list workspaces for ModelMirror watch", "modelmirror", crName)
+				return nil
+			}
+
+			var requests []reconcile.Request
+			for i := range wsList.Items {
+				ws := &wsList.Items[i]
+				if ws.Inference == nil || ws.Inference.Preset == nil {
```
Fei-Guo, line 116: We should call ModelStreamingEnabled(ws) here, if the ws has annotations to disable model streaming, we can ignore them for reconciliation.

Fei-Guo (approved):

PR #2112, opened 2026-06-17
feat: promote InferenceSet to v1beta1 and enable EnableInferenceSetController by default

andyzhangx:
### Changes
1. **Promote InferenceSet CRD to v1beta1** - Add InferenceSet types to `api/v1beta1/` with validation, defaulting, and DeepCopy
2. **Set v1beta1 as the storage version** - `+kubebuilder:storageversion` moved from v1alpha1 to v1beta1; CRD updated with v1beta1 as storage and v1alpha1 still served for backward compatibility
3. **Enable `EnableInferenceSetController` feature gate by default** - Change default from `false` to `true` (both in code and Helm chart)
4. **Migrate all internal references** - All controllers (InferenceSet, drift, autoupgrade), webhooks, manifests, metrics, and utilities now use `kaitov1beta1.InferenceSet*` types
5. **Register both v1alpha1 and v1beta1 InferenceSet in webhooks** - Ensures backward compatibility during CRD transition
6. **Update CRD manifest** - `config/crd/bases/kaito.sh_inferencesets.yaml` and Helm chart template now define both v1alpha1 and v1beta1
7. **Migrate E2E tests** - E2E test helpers and functions updated to use v1beta1 InferenceSet types
8. **Use `GetInferenceSetRuntimeName` helper** - Eliminated duplicated runtime selection logic in `inferencePoolTargetPort`, removed cross-version `kaitov1alpha1` dependency
9. **Fix metrics phase label** - `determineInferenceSetPhase` now returns `"succeeded"` to match metric bucket labels

### CRD Version Configuration
The CRD now defines two versions with the following settings:

| Version | `served` | `storage` | Meaning |
|---------|----------|-----------|---------|
| v1alpha1 | true | false | API server still accepts v1alpha1 requests (backward compatible for existing clients/scripts), but objects are no longer stored in this format |
| v1beta1 | true | **true** | API server accepts v1beta1 requests, and all objects are stored in etcd using this format |

- **`served`**: Whether the API server exposes this version as an endpoint. When `true`, clients can use this version in their API requests (e.g., `kubectl get inferencesets.v1alpha1.kaito.sh` still works).
- **`storage`**: Which version format is used to persist objects in etcd. Only one version can be the storage version. The API server automatically converts between served versions and the storage version on read/write.

### Motivation
- InferenceSet has been stable since v0.6.x and is actively used for multi-replica inference workloads
- P/D disaggregation (`MultiRoleInference`) depends on InferenceSet being enabled
- Users currently must manually set `--feature-gates=enableInferenceSetController=true` which is unnecessary friction
- Aligns with Kubernetes API maturity conventions (alpha → beta when API is stable)

### Storage Version
- **v1beta1 is now the storage version** — all new objects are stored as v1beta1
- v1alpha1 remains served for backward compatibility (existing clients can still read/write using v1alpha1 API)
- Since v1alpha1 and v1beta1 share identical Spec/Status schemas, no conversion webhook is needed at this time

### Related Issues
- Related to #2093 (P/D disaggregation support)


andyzhangx:
Re: Helm values.yaml — Agreed, the Helm chart `values.yaml` needs to be updated to remove the explicit `enableInferenceSetController=false` override. This will be done in a follow-up PR as noted in the TODO section.

andyzhangx:
Updated: v1beta1 is now intentionally the storage version. The `+kubebuilder:storageversion` marker is on v1beta1, and the CRD manifest has been updated to define both versions (v1alpha1 served but not storage, v1beta1 served + storage). PR description updated to reflect this.

zhuangqh:
please leave v1alpha1 pkg unchanged. others looks good.

zhuangqh (approved):

PR #2128, opened 2026-06-19
feat: Support streaming passthrough for RAGEngine chat completions

xiaoqi-7:
## Summary

Add OpenAI-compatible streaming passthrough support for the RAGEngine `/v1/chat/completions` endpoint when `stream=true`.

This change forwards upstream LLM SSE responses through FastAPI `StreamingResponse`, allowing clients to receive streamed chat completion chunks from the upstream inference service.

RAG streaming with `index_name` and output guardrails streaming are intentionally left unsupported in this PR.

## Changes

- Add SSE helper utilities for upstream parsing and downstream frame construction.
- Add streaming passthrough support in the inference client using `httpx.AsyncClient.stream`.
- Return `StreamingResponse` from `/v1/chat/completions` for supported `stream=true` passthrough requests.
- Reject unsupported streaming combinations:
  - `stream=true` with `index_name`
  - `stream=true` when output guardrails are enabled
- Add API tests for streaming passthrough and unsupported streaming cases.
- Add real uvicorn-based streaming test to verify the first SSE chunk is flushed to the client before the upstream response finishes.

### DIFF
--- a/presets/ragengine/inference/inference.py
+++ b/presets/ragengine/inference/inference.py
@@ -16,7 +16,7 @@ import asyncio
 import concurrent.futures
 import json
 import logging
-from collections.abc import Sequence
+from collections.abc import AsyncIterator, Sequence
 from typing import Any
 from urllib.parse import urljoin, urlparse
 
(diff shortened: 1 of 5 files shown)


presets/ragengine/inference/inference.py
```diff
@@ -310,6 +310,42 @@ async def chat_completions_passthrough(
             raise HTTPException(
                 status_code=500, detail=f"Error during POST request: {str(e)}"
             )
+
+    async def chat_completions_stream_passthrough(
+        self, chatCompletionsRequest: CompletionCreateParams, **kwargs: Any
+    ) -> AsyncIterator[str]:
+        if "/chat/completions" not in LLM_INFERENCE_URL:
```
Fei-Guo, line 317: Deferred errors: the HTTP 400 guard and raise_for_status() won't fire when chat_completion() is called/awaited — only when the stream is first iterated. 

This check needs to be done in the caller side since this method returns an async generator.

Fei-Guo (approved):

PR #2139, opened 2026-06-22
fix: [breaking] harden max-model-len and node estimation

### LINKED ISSUE #2132: llama3-70b and Mistral-Small-4-119B-2603 deployment failed due to overoptimistic max-model-len estimation
**Describe the bug**
When deploying models Mistral-Small-4-119B-2603/llama3-70b on with Azure SKU `Standard_NC80adis_H100_v5`, vLLM crashed with the following error:
```
ValueError: To serve at least one request with the model's max seq len (1048576),
22.5 GiB KV cache is needed, which is larger than the available KV cache memory (17.39 GiB).
Based on the available memory, the estimated maximum model length is 810208.
```

**Steps To Reproduce**
```
apiVersion: kaito.sh/v1beta1
kind: Workspace
metadata:
  name: workspace-mistral-medium
resource:
  instanceType: "Standard_NC80adis_H100_v5"
  labelSelector:
    matchLabels:
      apps: mistral-medium
inference:
  preset:
    name: "mistralai/Mistral-Small-4-119B-2603"
    presetOptions:
      modelAccessSecret: hf-secret
```
**Expected behavior**

**Logs**

**Environment**
- Install tools:
- Others:

**Additional context**

zhehli688:
**Reason for Change**:

This PR fixed multiple issues in KAITO's `max-model-len` estimation:

## 1. Estimation on runtime overhead is over-optimistic
**Problem:** The estimator reserved a **flat `2.3 GiB`** per GPU for non-KV runtime memory (activations + non-torch). vLLM actually reserves much more for large models — it *measures* this at startup in `determine_available_memory()` (peak activation forward pass) and `profile_cudagraph_memory()` (CUDA-graph capture, ~1.5–1.9 GiB on H100). For a big model like Mistral-Medium the real overhead is ~5.5 GiB/GPU, so the flat 2.3 over-counted available KV by ~3 GiB → the estimator computed a `max-model-len` (1,048,576) larger than vLLM could honor → deterministic `ValueError` at KV-cache init → crash-loop.

**Fix:** Replaced the flat constant with `baseOverheadGiB(2.3) + overheadWeightFactor(0.05) × per-GPU-weight-share`. The per-GPU weight share is a good proxy because activation + CUDA-graph memory grow with hidden-size/layer-count and are sharded across TP ranks the same way weights are.

## 2. KV cache size is underestimated for MLA models
**Problem:** The estimator computed per-token KV as `bytesPerToken / gpuCount`, assuming tensor parallelism **shards** the KV cache across GPUs. That's true for GQA/MHA, but **MLA (Multi-head Latent Attention) replicates its single compressed latent KV on every TP rank** — it isn't sharded. Using Mistral-Small-4-119B (MLA, TP=2) as an example, KAITO under-counted per-token KV by 2× (predicted 11,520, real 23,040), computed ~1.5M tokens, clamped to the 1M model limit, and the inference pod crash-looped because vLLM's real KV only fit ~810K.

**Fix:** Plumbed `AttnType` through to the estimator (`MLAReplicatedKVCache` input) and, for MLA, **don't divide by gpuCount**.

## 3. max-model-len calculation is buggy for pipeline-parallelism scenarios
**Problem:** The multi-node formula was `bytesPerToken / gpuCount * nodes` — it **multiplied** by node count. But pipeline parallelism (PP = nodes) **splits the model's layers across stages**, so each GPU holds only `1/nodes` of the layers' KV. Multiplying instead of dividing over-counted per-token KV by `nodes²` relative to truth. For MiniMax-M2.7 (TP2×PP2) KAITO computed 253,952 B/tok/GPU when the real value was ~63,488 — a 4× over-count — yielding `max-model-len=81,408` and silently throwing away ~60% of the model's usable 204,800 context. This was also internally inconsistent: weights in the same function are already divided by `nodes × gpuCount`, but KV wasn't.

**Fix:** Changed the standard-attention multi-node path to `bytesPerToken / gpuCount / nodes` (shard by TP, split layers by PP) and the MLA path to `bytesPerToken / nodes` (replicated on TP, still split by PP).

## TODOs
1. Per-token KV is greatly over-counted for models using sliding-window attention or hybrid attention 
(gemma4 6.2×, gpt-oss-120b 3.8×, qwen3.6 7.9×) because right now every layer is modeled as full-context. We should enhance KAITO memory estimator to handle such models accurately.
2. max-model-len estimation is too conservative on small GPUs. For example, when deploying model `Qwen3-8b` (16.4 GB) on an A10 GPU, KAITO estimated `max-model-len=11,008` while the actual max context len was 20,272. In future we may take GPU size into account in max-model-len estimation.

**Requirements**:
- [x] verified the max-model-len estimation for the following models on Azure SKU Standard_NC80adis_H100_v5:

| Models | Attn / topology | bytes per-tok KV /GPU (KAITO) | bytes per-tok KV /GPU (vLLM) | avail KV (KAITO) | avail KV (vLLM) | max-model-len (KAITO) | KV-cache size (vLLM)
|---|---|---|---|---|---|---|---|
| Mistral-Small-4-119B-2603 | MLA, TP2 | 23,040 | 23,040 | 16.4 GiB | 17.39 GiB | 761,600 | 810,176 |
| MiniMax-M2.7 | GQA, TP2×PP2 | 63,488 | 63,503 | 19.27 GiB | 19.97 GiB | 204,800 (clamped) | 337,664 |
| Llama-3.3-70B-Instruct | GQA, TP2 | 163,840 | 163,917 | 6.29 GiB | 8.48 GiB | 40,960 | 55,296 |
| gemma-4-31B-it | sliding-window, TP2 | 491,520 | 79,128 | 45.5 GiB | 39.45 GiB | **99,072** | 535,346 |
| gpt-oss-120b | sliding-window, TP2 | 73,728 | 19,602 | — | 40.84 GiB | 131,072 (clamped) | 131,072 |
| Qwen3.6-35B-A3B | hybrid GDN, TP2 | 81,920 | 10,395 | — | 40.04 GiB | 262,144 (clamped) | 2,236,798 |
| DeepSeek-R1 | MLA, TP8×PP2 | 35,136 | 34,560 | — | 19.7 GiB | 163,840 (clamped) | 612,144 |
| Kimi-2.6 | MLA, TP8×PP2 | 35,136 | 34,557 | — | 26.4 GiB | 262,144 (clamped) | 820,304 |
| Qwen3-8b | GQA, TP1 (1 node, A10) | 147,456 | 147,789 | 1.52 Gib | 2.79 GiB | **11,008** | 20,272 |

**Issue Fixed**:

Fixes #2132

**Notes for Reviewers**:

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -129,6 +129,11 @@ type Metadata struct {
 	// +optional
 	BytesPerToken int `yaml:"bytesPerToken,omitempty"`
 
+	// AttnType specifies the attention implementation (e.g., MHA, GQA, MQA, MLA),
+	// computed by the preset generator from the model config.
+	// +optional
+	AttnType string `yaml:"attnType,omitempty"`
+
 	// ModelTokenLimit is the maximum number of tokens (context window) supported by the model.
 	// This field is only for best effort supported vLLM models.
 	// +optional
@@ -199,10 +204,6 @@ type PresetParam struct {
 	// To determine TotalSafeTensorFileSize and BytesPerToken values for a new model,
 	// run the presets/workspace/generator/preset_generator.py script
 	// with the model's Hugging Face repository ID as an argument.
-
-	// AttnType specifies the attention implementation (e.g., MHA, GQA, MLA).
-	// Calculated by the preset generator based on model config.
-	AttnType string `yaml:"attn_type,omitempty"`
 
 	RuntimeParam
 
--- a/pkg/workspace/estimator/max_model_len.go
+++ b/pkg/workspace/estimator/max_model_len.go
@@ -32,19 +32,19 @@ const (
 	// loaded by vLLM relative to the on-disk safetensor size.
 	weightExpansionFactor = 1.02
 
-	// baseOverheadGiB is the model-independent part of vLLM's fixed per-GPU
+	// BaseOverheadGiB is the model-independent part of vLLM's fixed per-GPU
 	// overhead: non-torch allocations such as the CUDA context and NCCL buffers
 	// (~0.6 GiB) plus a baseline for small-model activations and CUDA graphs
-	// (~1.7 GiB). Larger models add to this via overheadWeightFactor below.
-	baseOverheadGiB = 2.3
+	// (~1.7 GiB). Larger models add to this via OverheadWeightFactor below.
+	BaseOverheadGiB = 2.3
 
-	// overheadWeightFactor scales the runtime overhead with the per-GPU model
-	// weight share (modelWeightOverhead). Peak activation memory and CUDA graph
-	// capture both grow with hidden size / layer count and are sharded across TP
-	// ranks the same way weights are, so the per-GPU weight share is a good proxy
-	// for them. vLLM measures these empirically in determine_available_memory() and
+	// OverheadWeightFactor scales the runtime overhead with the per-GPU model
+	// weight share. Peak activation memory and CUDA graph capture both grow with
+	// hidden size / layer count and are sharded across TP ranks the same way
+	// weights are, so the per-GPU weight share is a good proxy for them. vLLM
+	// measures these empirically in determine_available_memory() and
 	// profile_cudagraph_memory(). We approximate at best effort here.
-	overheadWeightFactor = 0.05
+	OverheadWeightFactor = 0.05
 )
 
 // MaxModelLenInput contains the inputs required to compute the optimal
@@ -160,7 +160,7 @@ func calculateMemoryParameters(in MaxModelLenInput, weightGiB float64) (float64,
 	// Fixed runtime overhead per GPU: a model-independent base plus a term that
 	// scales with the per-GPU model weight share, approximating the activation and
 	// CUDA-graph memory that grow with model size (see constant docs).
-	staticOverhead := baseOverheadGiB + overheadWeightFactor*modelWeightOverhead
+	staticOverhead := BaseOverheadGiB + OverheadWeightFactor*modelWeightOverhead
 
 	availableMemoryGiB := usableMemoryPerGPU - modelWeightOverhead - staticOverhead
 	availableMemoryBytes := availableMemoryGiB * (1 << 30)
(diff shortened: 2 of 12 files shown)


zhuangqh:
Nice work. Thanks for addressing this.
1 2.3GiB is an experience value to reserve space for cuda driver and non-torch usage. I think it's good for deepseek series. Not sure if it's mistral model introduce this extra 3GiB overhead. Can you double check this value for deepseek series.
2 --max-model-len silently dropped when ComputeMaxModelLen returns 0
The logic here wont return an error since we can enlarge node count to place this model. can you double check that? is it because give multimode but skip the maxmodellen param?
4. max-model-len calculation is buggy for pipeline-parallelism scenarios
yeah. we have the same understanding. but the code is buggy. we are intended to do 
`adjustedBytesPerToken = bytesPerToken / (gpuCount * nodes)`
not `adjustedBytesPerToken = bytesPerToken / gpuCount * nodes` ....

zhuangqh:
one big challenge here is to find a context size for a moderate model running on a small gpu.
like run Qwen/Qwen3.5-9B with A10. can you check this is good with the new formular.

And please update the doc https://kaito-project.github.io/kaito/docs/memory-estimator

zhehli688:
@zhuangqh 

> 2.3GiB is an experience value to reserve space for cuda driver and non-torch usage. I think it's good for deepseek series. Not sure if it's mistral model introduce this extra 3GiB overhead. Can you double check this value for deepseek series.

Tested Deepseek R1 with PP=2. The overhead was 2.4 GiB in PP0 stage and 6.6 GiB in PP1 stage. The estimator in this PR gave 4.34 GiB.

> 2 --max-model-len silently dropped when ComputeMaxModelLen returns 0
The logic here wont return an error since we can enlarge node count to place this model. can you double check that? is it because give multimode but skip the maxmodellen param?

There're multiple code paths returning 0 in `ComputeMaxModelLen`. I just added a sanity check to raise errors early and avoid pod crashing afterwards.

> one big challenge here is to find a context size for a moderate model running on a small gpu. like run Qwen/Qwen3.5-9B with A10. can you check this is good with the new formular.
> 

Qwen/Qwen3.5-9B still works on A10 but the underestimation of max-model-len (11,008 vs 20,272) is larger compared with H100. On A10 even the 2.3GB base overhead is an overestimate. We may take GPU size into account as well in our estimator in future.

zhuangqh (approved):

Fei-Guo (approved):

PR #2144, opened 2026-06-23
feat: migrate to llm-d-router-gateway chart v0.9.0 and enable KV cache-aware routing

### LINKED ISSUE #924: inferencepool chart: bool flags rendered as two args instead of --key=value
## Description

The `inferencepool` helm chart's `flags` field renders each entry as two separate args (`--key` and `"value"`). For boolean flags like `--secure-serving`, Go's `pflag` requires `--secure-serving=false` as a single arg — the two-arg form `--secure-serving` `"false"` is not recognized, so the flag stays at its default (`true`).

There is no way to express `--key=value` (single arg) through the chart's `flags` field.

## Environment

- Chart: `oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool` v1.4.0
- EPP image: `ghcr.io/llm-d/llm-d-inference-scheduler:v0.8.0-rc.1`

## Steps to reproduce

1. Install with `secure-serving: "false"` in values:
```yaml
inferenceExtension:
  flags:
    secure-serving: "false"
```

2. Check rendered args in the deployment:
```bash
kubectl get deployment <epp> -o jsonpath='{.spec.template.spec.containers[?(@.name=="epp")].args}'
```

3. Observe two separate args instead of one:
```
["--secure-serving", "false"]   # actual (broken)
["--secure-serving=false"]      # expected
```

4. Confirm in EPP logs that the flag is ignored:
```bash
kubectl logs <epp-pod> -c epp | grep secure-serving
# shows "secure-serving":true despite setting "false" in values
```

## Workaround

Manually patch the deployment after helm install:
```bash
kubectl patch deployment <epp> --type='json' \
  -p='[{"op":"add","path":"/spec/template/spec/containers/1/args/-","value":"--secure-serving=false"}]'
```

Note: This workaround is fragile — any `helm upgrade` will overwrite the patch.

andyzhangx:
## What type of PR is this?
/kind feature

## What this PR does / why we need it

Migrates the InferencePool Helm chart dependency from the Gateway API Inference Extension (GWIE) `inferencepool` chart to the `llm-d-router-gateway` chart at **v0.9.0**, and enables KV cache-aware routing in the default EPP EndpointPickerConfig.

### Why migrate?

1. **Bool flags rendering bug** — The GWIE chart (v1.3.1) renders boolean flags as two separate args (`--key` `"value"`), which breaks Go pflag parsing (see [llm-d/llm-d-router#924](https://github.com/llm-d/llm-d-router/issues/924), [kubernetes-sigs/gateway-api-inference-extension#2871](https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2871)). The llm-d-router chart already has this fix.
2. **Better alignment** — KAITO already uses the llm-d EPP image; using the llm-d chart directly removes the need for image override hacks and ensures chart+image version consistency.
3. **Advanced features** — The llm-d-router chart supports proxy mode, tokenizer sidecar, latency predictor, and other features that KAITO can leverage in future PRs.

### KV cache-aware routing

With [#1890](https://github.com/kaito-project/kaito/pull/1890) enabling KV cache events on vLLM pods by default, this PR adds the `kv-cache-utilization-scorer` plugin to the default P/D EndpointPickerConfig. The scorer reads each endpoint's KV cache utilization from vLLM metrics (`/metrics`) and computes `score = 1 - kvCacheUsagePercent`, spreading traffic away from pods with high KV cache pressure.

- Plugin wired into both **prefill** and **decode** scheduling profiles with **weight 5** (half of `load-aware-scorer`'s weight 10)
- Queue depth (`load-aware-scorer`) remains the primary routing signal; KV cache pressure acts as a secondary signal
- No additional sidecar or ZMQ subscription is required — the scorer consumes the same metrics endpoint EPP already scrapes

### Why `kv-cache-utilization-scorer` and not `prefix-cache-scorer` with `precise-prefix-cache-producer`?

The llm-d Router provides two levels of prefix-cache-aware routing:

| | Approximate (current) | Precise (future) |
|---|---|---|
| **Producer** | `approx-prefix-cache-producer` | `precise-prefix-cache-producer` + `token-producer` |
| **How it works** | EPP approximates tokens using character-to-token ratios, maintains an in-memory LRU index of prefix hashes | Subscribes to vLLM KV cache events via ZMQ, maintains a globally consistent index of exact token blocks per pod |
| **Accuracy** | Heuristic (character-based) | 100% (token-based) |
| **Dependencies** | None beyond EPP | Tokenizer sidecar (vLLM render endpoint), ZMQ connectivity to each vLLM pod, KV-Cache Indexer |
| **Resource overhead** | Negligible | ~500m CPU / 1Gi memory per MRI (for tokenizer sidecar) |

**This PR stays with the approximate approach** because:

1. **No tokenizer sidecar** — The `precise-prefix-cache-producer` requires a `token-producer` plugin that sends prompts to vLLM's `/v1/completions/render` endpoint (typically a `vllm launch render` sidecar in the EPP pod). KAITO does not currently deploy this sidecar, and adding one would cost ~500m CPU / 1Gi memory per MRI with no benefit for the approximate pipeline.

2. **KV-Cache Indexer not configured** — While [#1890](https://github.com/kaito-project/kaito/pull/1890) enables the ZMQ publisher on vLLM pods (port 5557), the EPP side's KV-Cache Indexer is not yet configured to subscribe to these events. The `precise-prefix-cache-producer` depends on this indexer to maintain the global block-to-pod mapping.

3. **`kv-cache-utilization-scorer` is complementary, not overlapping** — It reads the `kvCacheUsagePercent` metric from vLLM's `/metrics` endpoint (which EPP already scrapes for `load-aware-scorer`), requiring zero additional infrastructure. It provides a different signal: **how full is each pod's KV cache** (distribution/load-balancing), vs. **how much of this specific request's prefix is cached on each pod** (affinity). Both are valuable.

4. **The approximate producer is already in the config** — `approx-prefix-cache-producer` + `prefix-based-pd-decider` handles prefix-aware P/D routing without any external dependencies. Adding a `prefix-cache-scorer` on top would partially overlap with the decider's routing decisions in the P/D context.

### Next steps: Upgrade to precise prefix-cache routing (follow-up PRs)

Upgrading from the current approximate approach to precise prefix-cache routing will bring significant improvements, especially under high load and in P/D disaggregation scenarios:

#### What precise routing improves

| Aspect | Approximate (current) | Precise (future) |
|---|---|---|
| **Cache state awareness** | EPP **guesses** — "I sent prefix X to pod-2 last time, so pod-2 should still have it" | EPP **knows** — vLLM sends real-time `BlockStored` / `BlockRemoved` / `AllBlocksCleared` events via ZMQ, EPP maintains a globally consistent index |
| **Eviction handling** | ❌ EPP's LRU assumption breaks when pods evict blocks under memory pressure → stale routing → unnecessary prefill recomputation | ✅ EPP receives `BlockRemoved` events immediately → never routes to a pod that already evicted the prefix |
| **Tokenization accuracy** | Character-to-token ratio approximation, hash block boundaries may not align with real tokens → cache miss even when cache exists | Exact Token IDs from vLLM render endpoint → if cache is there, it will hit |
| **P/D transfer optimization** | Coarse-grained prefill/decode split only | Precise knowledge of which blocks are on which pod → optimal KV transfer decisions between prefill and decode pods |
| **Cache hit rate** | Good for homogeneous workloads with low eviction | Significantly higher under high load, heterogeneous prompts, or frequent cache eviction |

#### Concrete example

A user sends multiple prompts sharing a long system prompt (e.g., 4K tokens of instructions):

- **Approximate**: EPP hashes the system prompt characters → routes to pod-3 (where it sent a similar prefix before). But pod-3 evicted those blocks 30 seconds ago due to memory pressure. Result: **full prefill recomputation** (~2s latency penalty).
- **Precise**: EPP's KV-Cache Indexer received `BlockRemoved` from pod-3, knows pod-1 still has 90% of the blocks cached. Routes to pod-1. Result: **cache hit, near-zero prefill latency**.

#### Implementation roadmap

| Phase | PR | What | Prerequisite |
|---|---|---|---|
| **Phase 1** | Tokenizer sidecar | Add optional `vllm launch render` sidecar to EPP deployment, gated by feature flag or MRI spec field. Enables `token-producer` plugin for exact tokenization. | llm-d-router chart supports sidecar containers |
| **Phase 2** | KV-Cache Indexer | Configure EPP to subscribe to vLLM KV cache events on ZMQ port 5557 (already exposed by [#1890](https://github.com/kaito-project/kaito/pull/1890)). Maintains real-time block→pod mapping. | Phase 1 (token-producer needed for block hash computation) |
| **Phase 3** | Precise prefix-cache routing | Add `precise-prefix-cache-producer` + update `prefix-cache-scorer` with `prefixMatchInfoProducerName: precise-prefix-cache-producer`. Achieves 100% accurate prefix-cache-aware routing. | Phase 1 + Phase 2 |
| **Phase 4** | Speculative indexing | Enable `precise-prefix-cache-producer`'s speculative indexing to proactively update the index after routing decisions, closing the gap between routing and KV event arrival. | Phase 3 |

#### Trade-offs

- **Cost**: Tokenizer sidecar adds ~500m CPU / 1Gi memory per MRI; ZMQ connections add minor network overhead
- **Complexity**: More moving parts (sidecar lifecycle, ZMQ connectivity, indexer consistency)
- **When to upgrade**: Production deployments with high QPS, shared long prefixes (RAG, multi-turn chat), or latency-sensitive workloads where prefill recomputation is the bottleneck

### Changes

| Component | Before | After |
|-----------|--------|-------|
| Chart URL | `oci://registry.k8s.io/gateway-api-inference-extension/charts/inferencepool` | `oci://ghcr.io/llm-d/charts/llm-d-router-gateway` |
| Chart version | `v1.3.1` | `v0.9.0` |
| EPP image | `mcr.microsoft.com/oss/v2/llm-d/llm-d-inference-scheduler:v0.8.0` | `mcr.microsoft.com/oss/v2/llm-d/llm-d-router-endpoint-picker:v0.9.1` |
| Values schema | `inferenceExtension.image.{hub,name}` | `router.epp.image.{registry,repository}` |
| Values schema | `inferencePool.{targetPorts,modelServers}` | `router.modelServers.{targetPorts,matchLabels}` |
| EPP config | `load-aware-scorer` only | `load-aware-scorer` + `kv-cache-utilization-scorer` |

### Files changed

| File | Change |
|------|--------|
| `pkg/utils/consts/consts.go` | Update chart URL, version (v0.9.0), and EPP image constants |
| `pkg/workspace/manifests/manifests.go` | Update Helm values to match llm-d-router schema |
| `pkg/controllers/multiroleinference/controller.go` | Update schema + add `kv-cache-utilization-scorer` to default EndpointPickerConfig |
| `pkg/workspace/manifests/manifests_test.go` | Update expected values in tests |
| `test/e2e/preset_vllm_test.go` | Add e2e validation for `kv-cache-utilization-scorer` in EPP ConfigMap and pod logs |

## Which issue(s) this PR fixes
Resolves https://github.com/llm-d/llm-d-router/issues/924 (for KAITO users)

## Related PRs
- https://github.com/kaito-project/kaito/pull/1890 (enable KV cache events on vLLM pods by default)
- https://github.com/kubernetes-sigs/gateway-api-inference-extension/pull/2871 (upstream bool flag fix)
- https://github.com/llm-d/llm-d-inference-payload-processor/pull/17 (llm-d port of the fix)

### DIFF
--- a/pkg/workspace/manifests/manifests_test.go
+++ b/pkg/workspace/manifests/manifests_test.go
@@ -67,21 +67,21 @@ func TestGenerateInferencePoolHelmRelease(t *testing.T) {
 			name:      "statefulset inference pool helm values",
 			workspace: base.DeepCopy(),
 			expected: map[string]any{
-				"inferenceExtension": map[string]any{
-					"image": map[string]any{
-						"hub":        consts.EPPImageHub,
-						"name":       consts.EPPImageName,
-						"tag":        consts.EPPImageTag,
-						"pullPolicy": string(corev1.PullIfNotPresent),
-					},
-				},
-				"inferencePool": map[string]any{
-					"targetPorts": []any{
-						map[string]any{
-							"number": float64(consts.PortInferenceServer),
+				"router": map[string]any{
+					"epp": map[string]any{
+						"image": map[string]any{
+							"registry":   consts.EPPImageRegistry,
+							"repository": consts.EPPImageRepository,
+							"tag":        consts.EPPImageTag,
+							"pullPolicy": string(corev1.PullIfNotPresent),
 						},
 					},
 					"modelServers": map[string]any{
+						"targetPorts": []any{
+							map[string]any{
+								"number": float64(consts.PortInferenceServer),
+							},
+						},
 						"matchLabels": map[string]any{
 							consts.WorkspaceCreatedByInferenceSetLabel: base.Name,
 							appsv1.PodIndexLabel:                       "0",
@@ -103,21 +103,21 @@ func TestGenerateInferencePoolHelmRelease(t *testing.T) {
 				return ws
 			}(),
 			expected: map[string]any{
-				"inferenceExtension": map[string]any{
-					"image": map[string]any{
-						"hub":        consts.EPPImageHub,
-						"name":       consts.EPPImageName,
-						"tag":        consts.EPPImageTag,
-						"pullPolicy": string(corev1.PullIfNotPresent),
-					},
-				},
-				"inferencePool": map[string]any{
-					"targetPorts": []any{
-						map[string]any{
-							"number": float64(consts.PortInferenceServer),
+				"router": map[string]any{
+					"epp": map[string]any{
+						"image": map[string]any{
+							"registry":   consts.EPPImageRegistry,
+							"repository": consts.EPPImageRepository,
+							"tag":        consts.EPPImageTag,
+							"pullPolicy": string(corev1.PullIfNotPresent),
 						},
 					},
 					"modelServers": map[string]any{
+						"targetPorts": []any{
+							map[string]any{
+								"number": float64(consts.PortInferenceServer),
+							},
+						},
 						"matchLabels": map[string]any{
 							consts.WorkspaceCreatedByInferenceSetLabel: base.Name,
 							appsv1.PodIndexLabel:                       "0",
@@ -139,21 +139,21 @@ func TestGenerateInferencePoolHelmRelease(t *testing.T) {
 				return ws
 			}(),
 			expected: map[string]any{
-				"inferenceExtension": map[string]any{
-					"image": map[string]any{
-						"hub":        consts.EPPImageHub,
-						"name":       consts.EPPImageName,
-						"tag":        consts.EPPImageTag,
-						"pullPolicy": string(corev1.PullIfNotPresent),
-					},
-				},
-				"inferencePool": map[string]any{
-					"targetPorts": []any{
-						map[string]any{
-							"number": float64(consts.PortInferenceServer),
+				"router": map[string]any{
+					"epp": map[string]any{
+						"image": map[string]any{
+							"registry":   consts.EPPImageRegistry,
+							"repository": consts.EPPImageRepository,
+							"tag":        consts.EPPImageTag,
+							"pullPolicy": string(corev1.PullIfNotPresent),
 						},
 					},
 					"modelServers": map[string]any{
+						"targetPorts": []any{
+							map[string]any{
+								"number": float64(consts.PortInferenceServer),
+							},
+						},
 						"matchLabels": map[string]any{
 							consts.WorkspaceCreatedByInferenceSetLabel: base.Name,
 							appsv1.PodIndexLabel:                       "0",
--- a/pkg/workspace/manifests/manifests_test.go
+++ b/pkg/workspace/manifests/manifests_test.go
@@ -75,6 +75,15 @@ func TestGenerateInferencePoolHelmRelease(t *testing.T) {
 							"tag":        consts.EPPImageTag,
 							"pullPolicy": string(corev1.PullIfNotPresent),
 						},
+						"resources": map[string]any{
+							"requests": map[string]any{
+								"cpu":    "1",
+								"memory": "2Gi",
+							},
+							"limits": map[string]any{
+								"memory": "16Gi",
+							},
+						},
 					},
 					"modelServers": map[string]any{
 						"targetPorts": []any{
@@ -111,6 +120,15 @@ func TestGenerateInferencePoolHelmRelease(t *testing.T) {
 							"tag":        consts.EPPImageTag,
 							"pullPolicy": string(corev1.PullIfNotPresent),
 						},
+						"resources": map[string]any{
+							"requests": map[string]any{
+								"cpu":    "1",
+								"memory": "2Gi",
+							},
+							"limits": map[string]any{
+								"memory": "16Gi",
+							},
+						},
 					},
 					"modelServers": map[string]any{
 						"targetPorts": []any{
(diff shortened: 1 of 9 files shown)


pkg/workspace/manifests/manifests_test.go
```diff
@@ -67,21 +67,33 @@ func TestGenerateInferencePoolHelmRelease(t *testing.T) {
...
+							"tag":        consts.EPPImageTag,
+							"pullPolicy": string(corev1.PullIfNotPresent),
+						},
+						"resources": map[string]any{
+							"requests": map[string]any{
+								"cpu":    "1",
+								"memory": "2Gi",
+							},
+							"limits": map[string]any{
+								"memory": "16Gi",
```
zhuangqh, line 84: how to determine the resource req here.

andyzhangx: default is as following in upstream epp chart: https://github.com/llm-d/llm-d-router/blob/v0.9.0/config/charts/README.md#epp-core-configuration-parameters

```
requests.cpu: "8"
requests.memory: 8Gi
limits.memory: 16Gi
```

and requests.cpu, requests.memory is too big IMO, so I decrease it in our config, and limits are the same

zhuangqh (approved):

Fei-Guo (approved):

PR #2150, opened 2026-06-24
feat: derive NodeClass RBAC from karpenterProvider config

rambohe-ch:
**Reason for Change**:

Replace hardcoded azure/aws cloudProviderName branches with the selected karpenterProviders entry's group and resourceName, mirroring the nodeclass-configmap template. Add value validation that fails rendering with a clear message when the provider or its group/resourceName is missing.

Btw: I will add a MockNodeClass in producation-stack for co-existence with azure karpenter.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


zhuangqh (approved):
approved. upgrade test failed with known issue. Will address later.

PR #2158, opened 2026-06-26
feat: Fix buffer holdback boundary scanning

xiaoqi-7:
## Summary

Fix streaming guardrails to detect blocked text that spans emitted content and the holdback buffer.

## Changes

- Scan the full pending buffer during `feed()`, including the holdback tail, before emitting any prefix.
- Keep emission limited to the safe prefix before the holdback boundary.
- Add a regression test for blocked content crossing the holdback boundary.
- Update existing holdback test expectations to verify the scanner sees the retained tail.

## Validation

```bash
python -m pytest presets/ragengine/tests/streaming

### CODE REVIEW

File: presets/ragengine/streaming/buffer_window.py [hunk 1]
```diff
@@ -72,42 +65,50 @@ def feed(self, text: str) -> WindowEmitResult:
             return WindowEmitResult(chunks=(), blocked=True)
 
         self._pending_buffer += text
-        scan_text = self._scan_window()
-        if len(scan_text) < self._min_scan_chars:
+        emit_prefix_len = self._emit_prefix_len()
```
Fei-Guo: Change emit-prefix-len -> emit-len,  (prefix is the term used to indicate there is an addition to the origin string, which is not the case here). 
_emit_prefix_len() -> _calc_emit_len()

holdback_chars -> holdback_len throughout the file.

Fei-Guo (approved):

PR #2161, opened 2026-06-28
fix: pin inference pods to provisioned nodes

zhuangqh:
**Reason for Change**:

Add BuildNodeSelector to the NodeProvisioner interface so each provisioner contributes node selector requirements that isolate inference pods to the nodes it manages for the workspace:

- AzureGPUProvisioner: kaito.sh/workspace[namespace]
- KarpenterProvisioner: karpenter.kaito.sh/workspace-name[-namespace]
- BYOProvisioner: nil (user-supplied labelSelector only)

Thread the provisioner through WorkspaceGeneratorContext and apply the selector via a new SetProvisionerNodeSelector pod-spec modifier.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


Fei-Guo (commented):
https://github.com/kaito-project/kaito/blob/28535122050dea30a3219394ef67f5d395a65233/pkg/utils/nodes/nodes.go#L127

I think this needs to be updated as well. Otherwise, the counting for ready node is still wrong when two workspaces share the same resournce.labelselector.

zhuangqh:
> https://github.com/kaito-project/kaito/blob/28535122050dea30a3219394ef67f5d395a65233/pkg/utils/nodes/nodes.go#L127
> 
> I think this needs to be updated as well. Otherwise, the counting for ready node is still wrong when two workspaces share the same resournce.labelselector.

correct. fixec.

Fei-Guo (approved):

PR #2169, opened 2026-06-30
feat: distributed cache integration

croomes:
**Reason for Change**:

Introduces a pluggable distributed caching framework for KAITO that enables near-instant model loading from a local/distributed cache, substantially reducing cold-start latency for large models (70B+). Caching is opt-in via a new `CacheSpec` API and gated behind a feature flag.

Aligns with the design doc in `docs/proposals/20260518-distributed-cache-integration.md` (proposal #2088).

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

What's included:

**API**
- New `CacheSpec` types on `Workspace` (and `InferenceSet`) with per-concern
  provider selection for `modelWeights` and `kvCache`
- New `ModelWeightsCacheReady` / `KVCacheReady` conditions
- Validation webhooks for cache configuration

**Cache framework (`pkg/cache`)**
- Provider interface + registry (pluggable backends)
- `PodMutations` pattern to inject cache wiring into the inference pod spec
- Cache Controller with node-topology watches and state-transition events
- `ReconcileCache` with timeout handling, `Required`-mode blocking, and
  `Opportunistic` mode that proceeds regardless of cache readiness
- `cacheName` pass-through, auto-detecting `DefaultCacheName` or a single CR,
  and erroring on ambiguous multiple CRs

**Distributed Accelerator for Cloud Storage (DACS) provider**
- StorageIntercept-based model weight caching
- Mounts client libraries via OCI `ImageVolume`
- Injects env vars: `RUNAI_STREAMER_EXPERIMENTAL_AZURE_CACHE_LIB`,
  `LD_LIBRARY_PATH`, `CACHE_DISCOVERY_URL`, `CACHE_SERVER_PORT`
- Noop provider for disabled/default path

**Integration & wiring**
- InferenceSet cache propagation to child Workspaces
- Config ConfigMap merge (Helm defaults + user overrides)
- Helm wiring: feature gate, provider config, RBAC, CRD updates
- Cache-prewarm container image (`docker/cache-prewarm`)

**Tests & docs**
- Unit + integration + e2e tests (coverage: cache 73%, dacs 92%, noop 100%)
- Distributed cache user guide (`website/docs/distributed-cache.md`)
- CSI migration plan (`docs/proposals/cache-csi-migration-plan.md`)

## Modes
- `Required` — blocks deployment until cache is ready
- `Opportunistic` — deploys regardless of cache readiness

## Testing
- `go test ./pkg/cache/...`
- Unit, integration, and e2e suites included

Co-authored by: @hasethuraman


docs/proposals/cache-csi-migration-plan.md
croomes, line 1: @hasethuraman I think we can remove this file?

docs/proposals/20260518-distributed-cache-integration.md
croomes, line 1: @hasethuraman should be renamed to `20260609-distributed-cache-integration.md`

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -327,6 +328,21 @@ func (c *WorkspaceReconciler) addOrUpdateWorkspace(ctx context.Context, wObj *ka
 		}
 	}
 
+	// TODO(distributed-cache): Integrate ModelMirror with cache warming. When cache is configured,
+	// set a provider-specific StorageClass on the ModelMirror PVC so downloads write-through to
+	// cache. In Required mode, also gate on provider.IsReady() after ModelMirror reaches Ready.
```
croomes: Is this still relevant, and can it support both CSI & non-CSI (interceptor)?

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,237 @@
...
+4. **Azure Workload Identity** configured on the KAITO nodes (for DefaultAzureCredential access to blob storage)
+
+## Installation
+
+### 1. Install Tachyon
+
+Install the Tachyon distributed cache operator, CSI driver, and mutating webhook. The webhook automatically injects cache libraries into pods labeled with `tachyon.azure.com/inject: "true"`.
+
+```bash
+helm install tachyon-cache oci://<registry>/charts/tachyon-cache \
```
andyzhangx: it depends on Tachyon ACR image, but this ACR image is private instead of public access, right?

andyzhangx: > Set up registry access The Tachyon container registry ( tachyonexternal.azurecr.io ) requires authenticated pull access. The Container Storage team will provision credentials and store them securely in your Azure Key Vault.

hasethuraman: Renamed it to dacs which will be for public.

docs/proposals/cache-csi-migration-plan.md
hasethuraman: Done. I cant resolve this comment.

docs/proposals/20260518-distributed-cache-integration.md
hasethuraman: Done. I cant resolve this comment.

pkg/workspace/controllers/workspace_controller.go
hasethuraman: Done. Please check. I cant resolve this comment.

docker/cache-prewarm/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:3.12-slim
```
andyzhangx: what is this Dockerfile for? there is no image build with this Dockerfile

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -109,7 +109,7 @@ def _reset_vllm_defaults(self):
 
         # See https://docs.vllm.ai/en/stable/serving/engine_args.html for more args
         engine_default_args = {
-            "model": "/workspace/vllm/weights",
+            "model": os.environ.get("KAITO_MODEL_PATH", "/workspace/vllm/weights"),
```
andyzhangx: why making this change?

docker/cache-prewarm/Dockerfile
hasethuraman: Removed it at the moment. Will discuss this with @croomes & patch soon

presets/workspace/inference/vllm/inference_api.py
hasethuraman: Removed.

pkg/cache/mutations.go
```diff
@@ -0,0 +1,437 @@
...
+func ApplyTemplateCacheMutations(ctx context.Context, ws *kaitov1beta1.Workspace, ss *appsv1.StatefulSet) {
+	if ws.Cache == nil {
+		return
+	}
+
+	mutations, err := collectMutations(ctx, nil, ws, "", "")
+	if err != nil {
+		klog.V(2).InfoS("Failed to collect cache mutations for template inference", "error", err)
+		return
+	}
```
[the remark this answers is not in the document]
hasethuraman: Addressed in fe6e5b4c — removed KAITO_MODEL_PATH injection entirely. vLLM uses  --model=az://...  directly with RunAI streamer; the DACS cache is a transparent read-through layer via env vars. No local model path is needed

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,237 @@
...
+inference:
+  preset:
+    name: "microsoft/phi-4"
+cache:
+  modelWeights:
+    provider: dacs
+    mode: Opportunistic
+  kvCache:
+    provider: dacs
+    mode: Opportunistic
```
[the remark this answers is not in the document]
hasethuraman: Addressed in fe6e5b4c — rewrote the docs section. All examples now use  cache.modelCache  matching the CRD/API field name

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,237 @@
...
+You can configure each concern independently. For example, use `Required` for model weights (to guarantee fast startup) but `Opportunistic` for KV cache (which is an optimization, not a requirement):
+
+```yaml
+cache:
+  modelWeights:
+    provider: dacs
+    mode: Required
+  kvCache:
+    provider: dacs
+    mode: Opportunistic
```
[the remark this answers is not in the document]
hasethuraman: Addressed in fe6e5b4c — same as above, all examples corrected.

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,237 @@
...
+2. AI Runway resolves provider-specific config and writes it to `status.cache.kvCache`
+3. KAITO reads the resolved config and applies pod mutations
+
+This allows AI Runway to manage fleet-level cache policies while KAITO handles per-pod injection.
+
+## Troubleshooting
+
+### Cache not ready
+
+If your Workspace is stuck with condition `ModelWeightsCacheReady=False`:
```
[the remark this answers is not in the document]
hasethuraman: Addressed in fe6e5b4c — condition name corrected to  ModelCacheReady  throughout the docs.

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,237 @@
...
+   - Full connector class path: `dacs_client.connectors.vllm_connector.DacsKVConnector`
+   - Discovery endpoint, protocol, and TTL settings
+
+### Auto-Discovery
+
+If `discoveryEndpoint` is left empty in the Helm configuration, KAITO automatically reads the endpoint from the Cache CR's `status.discoveryEndpoint` field. This enables zero-configuration when DACS is installed in the same cluster.
+
+### Prewarm
+
+If a `prewarmImage` is configured, KAITO can create Kubernetes Jobs that download model weights from Hugging Face and upload them to blob storage before the inference pod starts. Prewarm Jobs also receive the `dacs.azure.com/inject` label so the webhook injects cache client libraries.
```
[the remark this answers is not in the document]
hasethuraman: Addressed in fe6e5b4c — removed prewarm section entirely (no prewarm implementation). Auto-discovery is implemented:  resolveDiscoveryEndpoint()  reads  status.discoveryURL  from the Cache CR. Docs updated to reflect actual behavior.

pkg/cache/mutations.go
```diff
@@ -0,0 +1,437 @@
...
+	if len(src.Labels) > 0 {
+		if dst.Labels == nil {
+			dst.Labels = make(map[string]string, len(src.Labels))
+		}
+		for k, v := range src.Labels {
+			dst.Labels[k] = v
+		}
+	}
+
+	// Deduplicate env vars by name (last wins).
```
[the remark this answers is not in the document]
hasethuraman: This is intentional — user-specified env vars in the pod template should take priority over cache-injected ones.

pkg/cache/node_handler.go
```diff
@@ -0,0 +1,67 @@
...
+		return false
+	}
+	// Trigger reconcile when a node becomes Ready or is removed.
+	for _, cond := range node.Status.Conditions {
+		if cond.Type == corev1.NodeReady {
+			return true
+		}
+	}
+	return true
+}
```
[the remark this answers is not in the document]
hasethuraman: Fixed — simplified to type-assert + return true. The condition loop was dead code since every node

charts/kaito/workspace/values.yaml
```diff
@@ -186,4 +187,24 @@ local-csi-driver:
...
+
+# Distributed cache provider configuration.
+# Only active when featureGates.distributedCache is true.
+# The DACS CSI driver + mutating webhook must be pre-installed in the cluster.
+# KAITO adds pod labels to trigger webhook-based cache library injection.
+cache:
+  providers:
+    dacs:
+      enabled: false
+      discoveryEndpoint: "http://cacheserver-discovery.dacs-cache-system.svc.cluster.local:9065"
```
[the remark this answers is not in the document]
hasethuraman: Fixed — set to empty string to enable auto-discovery from Cache CR

cmd/workspace/main.go
```diff
@@ -202,6 +206,20 @@ func main() {
...
+		dynamicClient, dynErr := dynamic.NewForConfig(cfg)
+		if dynErr != nil {
+			klog.ErrorS(dynErr, "unable to create dynamic client for cache providers")
+			exitWithErrorFunc()
+		}
+		dacsCfg := dacs.ConfigFromEnv()
+		cache.Register(dacs.New(dynamicClient, dacsCfg))
+		klog.InfoS("Registered DACS cache provider", "discoveryEndpoint", dacsCfg.DiscoveryEndpoint,
+			"kvCacheEnabled", dacsCfg.KVCacheEnabled)
+	}
```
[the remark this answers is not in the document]
hasethuraman: Fixed — DACS now requires both the feature gate AND DACS_ENABLED=true env var. Helm template sets this when cache.providers.dacs.enabled=true

pkg/cache/dacs/provider.go
```diff
@@ -0,0 +1,436 @@
...
+func DefaultConfig() Config {
+	return Config{
+		DiscoveryEndpoint:   DefaultDiscoveryEndpoint,
+		ClientImage:         DefaultClientImage,
+		KVCacheEnabled:      true,
+		KVConnectorProtocol: "tcp",
+		BlobContainer:       "kaito-models",
+		BlobPrefix:          DefaultBlobPrefix,
+	}
+}
```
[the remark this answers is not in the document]
hasethuraman: No longer sets non-empty value

test/e2e/preset_vllm_test.go
```diff
@@ -467,6 +468,34 @@ var _ = Describe("Workspace Preset on vllm runtime", func() {
 		validateModelsEndpoint(workspaceObj)
 		validateChatCompletionsEndpoint(workspaceObj)
 	})
+
+	It("should serve a public preset GPU workspace with no cache mutations when no cache is installed", utils.GinkgoLabelFastCheck, func() {
```
andyzhangx: have you enabled `FeatureFlagDistributedCache` somewhere in this e2e test?

hasethuraman: It is from helm values. Disabled at the moment here https://github.com/kaito-project/kaito/pull/2169/changes#diff-35b71f2c0558283e59a7a3b152d4a4b4ff6a6f860cfd0afcd72ea8d2b04b9c2aR48

pkg/cache/mutations.go
```diff
@@ -0,0 +1,440 @@
...
```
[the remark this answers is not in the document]
hasethuraman: Fixed in 8d84e084 — comment corrected to "first wins" (user-specified env vars take priority; cache-injected duplicates are skipped).

pkg/cache/mutations.go
```diff
@@ -0,0 +1,440 @@
...
+	}
+
+	// Inject env vars and volume mounts into the first (model) container.
+	spec.Containers[0].Env = append(spec.Containers[0].Env, mutations.EnvVars...)
+	spec.Containers[0].VolumeMounts = append(spec.Containers[0].VolumeMounts, mutations.VolumeMounts...)
+
+	// Inject volumes and init containers at the pod level.
+	spec.Volumes = append(spec.Volumes, mutations.Volumes...)
+	spec.InitContainers = append(spec.InitContainers, mutations.InitContainers...)
+}
```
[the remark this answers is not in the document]
hasethuraman: Fixed in 8d84e084 —  applyMutations  now deduplicates by name for env vars, volumes, volume mounts, and init containers. Existing entries in the pod template are preserved (first wins)

pkg/cache/mutations.go
```diff
@@ -0,0 +1,440 @@
...
+			return ""
+		}
+	} else {
+		// Update
+		existing.Data = merged
+		if err := kubeClient.Update(ctx, existing); err != nil {
+			klog.V(2).InfoS("Failed to update runtime cache ConfigMap", "name", runtimeName, "error", err)
+			return ""
+		}
+	}
```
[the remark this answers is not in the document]
hasethuraman: Fixed in 8d84e084 — added  mapsEqual  helper; ConfigMap update is skipped when merged data is unchanged, avoiding unnecessary API writes

pkg/workspace/inference/template_inference.go
```diff
@@ -19,12 +19,21 @@ import (
...
 	"github.com/kaito-project/kaito/pkg/workspace/manifests"
 )
 
 func CreateTemplateInference(ctx context.Context, workspaceObj *kaitov1beta1.Workspace, kubeClient client.Client) (client.Object, error) {
 	depObj := manifests.GenerateManifestWithPodTemplate(workspaceObj, defaultTolerations(workspaceObj))
+
+	// Apply cache mutations for template-based workspaces.
+	if featuregates.FeatureGates[consts.FeatureFlagDistributedCache] && workspaceObj.Cache != nil {
+		cache.ApplyTemplateCacheMutations(ctx, workspaceObj, kubeClient, depObj)
+	}
```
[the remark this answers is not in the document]
hasethuraman: Fixed in 8d84e084 —  ApplyTemplateCacheMutations  now returns an error. In Required mode, the error propagates and blocks workspace creation. In Opportunistic mode, it's logged and the workspace proceeds without cache.

pkg/cache/dacs/provider.go
```diff
@@ -0,0 +1,399 @@
...
+	// Discovery endpoint for DACS cache servers (hostname only, no scheme or port).
+	DefaultDiscoveryEndpoint = DefaultCacheName + "-discovery." + CacheNamespace + ".svc.cluster.local"
+	DefaultDiscoveryPort     = 9065
+
+	// ImageVolume configuration for the cache client sidecar libraries.
+	DefaultClientImage = "hariazstortest.azurecr.io/dacs-client:20260701.7"
+	ClientVolumeName   = "cache-client"
+	ClientMountPath    = "/opt/cache-client"
+	ClientLibPath      = ClientMountPath + "/usr/local/lib/python3.10/dist-packages/dacs_client/libStorageDirect.so"
+
```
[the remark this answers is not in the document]
hasethuraman: Fixed in 8d84e084 —  DefaultClientImage  is now empty string. Must be configured via  DACS_CLIENT_IMAGE  env var (exposed in Helm as  cache.providers.dacs.clientImage ).  PodMutations  returns an error if unset.

charts/kaito/workspace/templates/deployment.yaml
```diff
@@ -79,6 +79,18 @@ spec:
               value: {{ .Values.cloudProviderName | quote }}
             - name: CLUSTER_NAME
               value: {{ .Values.clusterName | quote }}
+            {{- if and .Values.featureGates.distributedCache .Values.cache.providers.dacs.enabled }}
+            - name: DACS_ENABLED
+              value: "true"
+            - name: DACS_DISCOVERY_ENDPOINT
+              value: {{ .Values.cache.providers.dacs.discoveryEndpoint | quote }}
+            - name: DACS_KV_CACHE_ENABLED
```
andyzhangx, line 88: what are these env vars for?

cmd/workspace/main.go
```diff
@@ -202,6 +206,20 @@ func main() {
 	cfg.UserAgent = workspaceController
 	setRestConfig(cfg, kubeClientQPS, kubeClientBurst)
 
+	// Register cache providers based on feature gates and configuration.
+	cache.Register(cachenoop.NewProvider())
+	if featuregates.FeatureGates[consts.FeatureFlagDistributedCache] && os.Getenv("DACS_ENABLED") == "true" {
```
andyzhangx, line 214: why use both `FeatureFlagDistributedCache` and env var `DACS_ENABLED` for feature enabling? is it redudant?

charts/kaito/workspace/templates/deployment.yaml
hasethuraman: Will check and remove if I can derive using the cache name

            - name: DACS_DISCOVERY_ENDPOINT
              value: {{ .Values.cache.providers.dacs.discoveryEndpoint | quote }}

DACS_KV_CACHE_ENABLED is to disable the kv cache of dacs while the model cache of dacs will be enabled.

cmd/workspace/main.go
hasethuraman: The cache provider layer is abstracted, and we have added dacs as one type of cache. Hence FeatureFlag is to complete disable the cache while DACS_ENABLED is to selectively disable the dacs. This will be helpful when we have multiple cache systems co-exist in the cluster and enable/disable at kaito-system level

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,233 @@
...
+resource:
+  instanceType: "Standard_NC24ads_A100_v4"
+  labelSelector:
+    matchLabels:
+      apps: phi-4
+inference:
+  preset:
+    name: "microsoft/phi-4"
+cache:
+  modelWeights:
```
andyzhangx: it should be `modelCahce` in the example:
```
type CacheSpec struct {
    ModelCache *ModelCacheSpec `json:"modelCache,omitempty"`
    KVCache    *KVCacheSpec    `json:"kvCache,omitempty"`
}
```

charts/kaito/workspace/templates/deployment.yaml
hasethuraman: I would like to retain DACS_DISCOVERY_ENDPOINT and we can extract the cache name, namespace etc for default configurations. WIP

andyzhangx:
Tested this PR end-to-end on an AKS cluster (`workspace:0.11.0-cache` image + DACS cache-servers Ready, `Workspace phi-4-cached`) and hit a silent failure worth flagging before merge.

## TL;DR

Workspace reaches `WorkspaceSucceeded=True` / `ModelCacheReady=True` / `BenchmarkCompleted=True` and reports 306k tokens/min — but **DACS was never touched**. vLLM downloaded the full 27.31 GiB from HuggingFace Hub in ~68s and served from `/workspace/weights`. Cache-servers received zero traffic.

Full write-up with commands, logs, and the component-by-component green/red table:
👉 https://github.com/andyzhangx/demo/blob/master/llm/distributed-cache/enable-dacs-in-kaito-workspace.md#verification

## Root cause

`pkg/cache/dacs/provider.go` injects the right label, ImageVolume, and env vars (`RUNAI_STREAMER_EXPERIMENTAL_AZURE_CACHE_ENABLED=true`, `CACHE_DISCOVERY_URL=…`, `LD_LIBRARY_PATH=…`) — but `pkg/workspace/inference/preset_inferences.go` still renders the vLLM invocation as:

```
python3 /workspace/vllm/inference_api.py --load_format=auto \
    --model=microsoft/phi-4 --download-dir=/workspace/weights ...
```

With `--load_format=auto` and `--model=<HF repo id>`, vLLM takes the default HF loader path and never dlopens `libStorageDirect.so`. The env vars have no reader. Confirmed with:

```
$ grep -inE 'runai|streamer|StorageDirect|dacs|azure_cache' /workspace/vllm/inference_api.py
(only kv_cache_* — nothing about the model streamer)
```

## Gaps I think need to land before / with this PR

1. **vLLM wiring** — when `workspace.Cache.ModelCache != nil` the preset renderer needs to switch `--load-format` to `runai_streamer_azure` (or `runai_streamer`) **and** rewrite `--model` to the `az://<account>/<container>/<path>` URI the DACS client understands, similar to the existing ModelMirror `streamingEnabled` branch. Otherwise this whole PR is a no-op at runtime.

2. **`LD_LIBRARY_PATH` breaks glibc-older base images.** The injection prepends `/opt/cache-client/usr/lib/x86_64-linux-gnu`, which shadows the base image's libc. If the DACS client image is Ubuntu 24.04 (glibc 2.38) and the workspace base is Ubuntu 22.04 (glibc 2.35), every binary in the pod — including `/bin/sh` — crashes with `GLIBC_2.38 not found`. Suggest either scoping `LD_LIBRARY_PATH` to just the `dacs_client` dir, RPATH-ing `libStorageDirect.so` at build time, or preflight-checking the client's glibc floor against the base image and failing `ModelCacheReady` with a clear reason.

3. **`ModelCacheReady=True` is a false positive.** It currently mirrors the Cache CR's own `Ready=True` without confirming the workspace pod is consuming the cache. Consider tying it to an observable pod-side signal (detected `--load-format runai_streamer_azure`, or a streamer-side counter).

4. **`Cleanup(cleanupOnDelete=true)` is a TODO** — provider.go logs "not yet implemented". Either implement or document as ignored.

5. **e2e conformance can't catch this.** `test/e2e/cache_integration_test.go` asserts env/label/volume injection + condition flip — exactly the shape of "green" we saw with zero DACS traffic. Adding a `delta(cache-server bytes_streamed) > 0` assertion after workspace-ready would have caught the regression.

## Suggestion

Splitting into (a) scaffolding — API/CRD/feature gate/injection contract/conformance harness (all looks good) and (b) wiring — `--load-format` switch, `az://` model URI rewrite, glibc-safe `LD_LIBRARY_PATH`, `Cleanup`, DACS-traffic e2e — would keep the "ready-but-silently-defeated" state from shipping.

Happy to help with the vLLM invocation switch if useful.

docker/cache-prewarm/Dockerfile
croomes: No need for pre-warm/dockerfile.

charts/kaito/workspace/templates/deployment.yaml
hasethuraman: I have retained DACS_DISCOVERY_ENDPOINT 

Now it becomes easier to just give this discovery endpoint like <dacs-installation-name>.<namespace>.svc.cluster.local. We extract the name and namespace from this & easier from now on that hard requirement of cache-sample is not requried.

Fixed

website/docs/distributed-cache.md
hasethuraman: Thanks @andyzhangx addressed.

hasethuraman:
> Tested this PR end-to-end on an AKS cluster (`workspace:0.11.0-cache` image + DACS cache-servers Ready, `Workspace phi-4-cached`) and hit a silent failure worth flagging before merge.
> 
> ## TL;DR
> Workspace reaches `WorkspaceSucceeded=True` / `ModelCacheReady=True` / `BenchmarkCompleted=True` and reports 306k tokens/min — but **DACS was never touched**. vLLM downloaded the full 27.31 GiB from HuggingFace Hub in ~68s and served from `/workspace/weights`. Cache-servers received zero traffic.
> 
> Full write-up with commands, logs, and the component-by-component green/red table: 👉 https://github.com/andyzhangx/demo/blob/master/llm/distributed-cache/enable-dacs-in-kaito-workspace.md#verification
> 
> ## Root cause
> `pkg/cache/dacs/provider.go` injects the right label, ImageVolume, and env vars (`RUNAI_STREAMER_EXPERIMENTAL_AZURE_CACHE_ENABLED=true`, `CACHE_DISCOVERY_URL=…`, `LD_LIBRARY_PATH=…`) — but `pkg/workspace/inference/preset_inferences.go` still renders the vLLM invocation as:
> 
> ```
> python3 /workspace/vllm/inference_api.py --load_format=auto \
>     --model=microsoft/phi-4 --download-dir=/workspace/weights ...
> ```
> 
> With `--load_format=auto` and `--model=<HF repo id>`, vLLM takes the default HF loader path and never dlopens `libStorageDirect.so`. The env vars have no reader. Confirmed with:
> 
> ```
> $ grep -inE 'runai|streamer|StorageDirect|dacs|azure_cache' /workspace/vllm/inference_api.py
> (only kv_cache_* — nothing about the model streamer)
> ```
> 
> ## Gaps I think need to land before / with this PR
> 1. **vLLM wiring** — when `workspace.Cache.ModelCache != nil` the preset renderer needs to switch `--load-format` to `runai_streamer_azure` (or `runai_streamer`) **and** rewrite `--model` to the `az://<account>/<container>/<path>` URI the DACS client understands, similar to the existing ModelMirror `streamingEnabled` branch. Otherwise this whole PR is a no-op at runtime.
> 2. **`LD_LIBRARY_PATH` breaks glibc-older base images.** The injection prepends `/opt/cache-client/usr/lib/x86_64-linux-gnu`, which shadows the base image's libc. If the DACS client image is Ubuntu 24.04 (glibc 2.38) and the workspace base is Ubuntu 22.04 (glibc 2.35), every binary in the pod — including `/bin/sh` — crashes with `GLIBC_2.38 not found`. Suggest either scoping `LD_LIBRARY_PATH` to just the `dacs_client` dir, RPATH-ing `libStorageDirect.so` at build time, or preflight-checking the client's glibc floor against the base image and failing `ModelCacheReady` with a clear reason.
> 3. **`ModelCacheReady=True` is a false positive.** It currently mirrors the Cache CR's own `Ready=True` without confirming the workspace pod is consuming the cache. Consider tying it to an observable pod-side signal (detected `--load-format runai_streamer_azure`, or a streamer-side counter).
> 4. **`Cleanup(cleanupOnDelete=true)` is a TODO** — provider.go logs "not yet implemented". Either implement or document as ignored.
> 5. **e2e conformance can't catch this.** `test/e2e/cache_integration_test.go` asserts env/label/volume injection + condition flip — exactly the shape of "green" we saw with zero DACS traffic. Adding a `delta(cache-server bytes_streamed) > 0` assertion after workspace-ready would have caught the regression.
> 
> ## Suggestion
> Splitting into (a) scaffolding — API/CRD/feature gate/injection contract/conformance harness (all looks good) and (b) wiring — `--load-format` switch, `az://` model URI rewrite, glibc-safe `LD_LIBRARY_PATH`, `Cleanup`, DACS-traffic e2e — would keep the "ready-but-silently-defeated" state from shipping.
> 
> Happy to help with the vLLM invocation switch if useful.

@andyzhangx thanks for catching the crucial ones.

Fixed them & documented the non-critical ones. Please find the summary here

────────────────────

Reply to Claim 1 — vLLM wiring / silent no-op (blocker)

Great catch, and thanks for the end-to-end write-up — this was the real bug.

Fixed in  a2b40440 . Rather than force-rewriting  --load-format / --model  inside the cache path (which would collide with the existing ModelStreaming/ModelMirror rendering that already owns the  az://  +  runai_streamer  switch), we gate cache injection on the load format that actually consumes it:

// preset_inferences.go
cacheApplicable := workspaceObj.Cache != nil && streamingLoadFormat == "runai_streamer"

• When model streaming is enabled, the preset already renders  --load-format=runai_streamer --model=az://… , so  libStorageDirect.so  is dlopened and the injected env vars have a reader — cache is consumed.
• When it's the default HF download path, we no longer inject env/volume/label that nothing reads; it's skipped instead of a green-but-defeated no-op. The BYO-template path ( ApplyTemplateCacheMutations ) does the same by scanning container args for  runai_streamer .

Verified on a live AKS cluster: cache e2e 36 passed / 0 failed, including a warm→reload spec that confirms remote cache hits on the second load (custom-runai base + injected dacs-client via ImageVolume).

────────────────────

Reply to Claim 2 —  LD_LIBRARY_PATH  breaks glibc-older bases

Addressed in  a2b40440 . The injection no longer touches  LD_LIBRARY_PATH  at all — that prepend (which shadowed the base image's libc) is removed. We now point the streamer straight at the vendored library:

RUNAI_STREAMER_EXPERIMENTAL_AZURE_CACHE_LIB=/opt/cache-client/usr/local/lib/python3.10/dist-packages/dacs_client/libStorageDirect.so

The wheel is  auditwheel -repaired and vendors all native deps (incl. libjsoncpp) under  dacs_client_dev.libs/ , resolved via RPATH — so no  LD_LIBRARY_PATH  surgery is needed and  /bin/sh  etc. keep the base libc. On top of that, we documented the remaining hard constraint: the DACS client is built  manylinux_2_35 , so the base image must be glibc ≥ 2.35 (Ubuntu 22.04+). This is now called out in the  clientImage  godoc ( provider.go ), Helm  values.yaml , and a  :::warning  in  website/docs/distributed-cache.md .

────────────────────

Reply to Claim 3 —  ModelCacheReady=True  is a false positive

Agreed on the semantics. Today  ModelCacheReady  mirrors the cluster-level Cache CR  Ready  and does not observe pod-side consumption, so it's a readiness-of-dependency signal, not a per-model warm/hit signal. We didn't want to silently overload the condition, so for this PR we've documented the exact meaning (new "Cache Readiness Conditions"  :::note  in  website/docs/distributed-cache.md ) and left a follow-up to tie it to an observable streamer-side signal ( --load-format runai_streamer_azure  detected / streamer counter). With Claim 1 fixed, the condition is now only set on paths where the cache is genuinely wired, which removes the misleading "ready but defeated" combination you hit.

────────────────────

Reply to Claim 4 —  Cleanup(cleanupOnDelete=true)  is a TODO

Documented as ignored for now ( a2b40440 ).  cleanupOnDelete  is a released  v1beta1  field we can't remove, and  Cleanup()  is still a placeholder with no finalizer, so we've marked it explicitly rather than imply behavior: the godoc on  CleanupOnDelete  ( api/v1beta1/workspace_types.go ) now reads "Not yet implemented / reserved," CRDs were regenerated ( make manifests ), and a "Cleanup on Delete"  :::note  was added to the user docs. Implementing the actual finalizer-driven cache eviction is tracked as a follow-up.

────────────────────

Reply to Claim 5 — e2e conformance can't catch this

Right — the injection-shape assertions would stay green under the no-op. Two things landed:

1. A unit regression test ( TestGeneratePresetInference_CacheGatedOnStreaming ) with a guard provider that injects a sentinel label/env; it fails if the preset path ever injects cache mutations when the load format isn't  runai_streamer . Confirmed it fails against the old gate and passes with the fix.
2. The DACS e2e now exercises a warm→reload scenario that validates actual remote cache hits on the second load (not just injection), gated behind  TEST_CACHE_ENABLED  + the  DACS_CACHE_*  env. A cache-server  bytes_streamed  delta assertion is a good hardening follow-up on top of that.

────────────────────

PR #2177, opened 2026-07-02
fix: set expectations to prevent duplicate workspaces in InferenceSet

rambohe-ch:
**Reason for Change**:

The InferenceSet controller constructed the ControllerExpectations cache and checked SatisfiedExpectations, but never actually set expectations before creating or deleting workspaces. As a result SatisfiedExpectations always returned true, providing no protection against stale cache reads.

For a replicas=1 InferenceSet this allowed a create race: two near- simultaneous reconciles both read 0 workspaces from the stale cache and each created one, producing a redundant workspace that was later deleted.

Set creation expectations via ExpectCreations before the create loop and deletion expectations via ExpectDeletions before issuing deletes, lowering them on failure so a subsequent reconcile short-circuits until the create/delete events are observed by the workspace event handler.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/inferenceset/inferenceset_controller.go
+++ b/pkg/inferenceset/inferenceset_controller.go
@@ -17,7 +17,6 @@ import (
 	"context"
 	"fmt"
 	"maps"
-	"slices"
 	"sort"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/inferenceset/inferenceset_controller.go
```diff
@@ -216,50 +215,55 @@ func (c *InferenceSetReconciler) addOrUpdateInferenceSet(ctx context.Context, iO
...
-					klog.InfoS("Deleting extra workspace...", "workspace", klog.KObj(&ws))
-					if err := c.Client.Delete(ctx, &ws, &client.DeleteOptions{}); err != nil {
-						klog.ErrorS(err, "failed to delete extra workspace", "workspace", klog.KObj(&ws))
+		if len(toDelete) > 0 {
+			// Set deletion expectations before issuing any delete so that stale
+			// cache reads in subsequent reconciles do not over-delete workspaces.
+			// The expectation is lowered when the delete event is observed by the
+			// workspace event handler, or here if the delete call does not result
+			// in an eventual delete event (already-gone or failed).
+			if err := c.expectations.ExpectDeletions(c.klogger, isKey, len(toDelete)); err != nil {
```
zhuangqh, line 253: do you really need this?

rambohe-ch: create and delete are a pair of expectations. maybe we should keep it.

zhuangqh (approved):

PR #2179, opened 2026-07-06
fix: enable Karpenter static provisioning for AWS and refresh provisioner docs

rambohe-ch:
**Reason for Change**:

- Makefile aws-karpenter-helm: enable settings.featureGates.staticCapacity=true so KAITO NodePool spec.replicas actually provisions GPU nodes; bump AWS_KARPENTER_VERSION to 1.13.0 and move cluster config under examples/aws
- examples/aws/clusterconfig.yaml.template: split Karpenter controller policy into the granular v1 policies and add zonalShiftConfig; use AmazonLinux2023 nodegroup
- hack/deploy/generate-identities.sh: use --assignee-object-id/--assignee-principal-type to avoid Graph lookup blocked by conditional access
- docs: migrate azure.md to Azure Karpenter (deprecate gpu-provisioner); document the StaticCapacity gate and link the official Karpenter getting-started guide in aws.md

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


rambohe-ch:
BTW: I've met an error for nvme csi provider on aws gpu nodes, and posted an issue to track this error: https://github.com/Azure/local-csi-driver/issues/529

rambohe-ch:
A temporary workaround is delete the storageclass: kaito-local-nvme-disk in kaito-workspace namespace.

website/docs/azure.md
```diff
@@ -34,100 +45,81 @@ Alternative: If you already have GPU nodes or manage them separately, use the [b
...
 ```
 
-### Step 3: Install GPU Provisioner
-
-Install the Azure GPU Provisioner using Helm:
+### Step 3: Install Azure Karpenter
 
 ```bash
-export GPU_PROVISIONER_VERSION=0.4.2
-
```
zhuangqh, line 81: I think we can fully delegate these steps to the official Karpenter documentation rather than using a make command. This way, we don't have to worry about any potential inconsistencies between them.

rambohe-ch: I think we've provided these make target for users, so I'd like to use them in our docs. If we delegate these steps to the related official Karpenter docs, should we remove related codes in Makefile?

zhuangqh: the main idea here is that user dont have to clone kaito repo.

rambohe-ch: Ok, I've modified to use the official Karpenter for env setup, and added makefile targets as an alternative solution.

zhuangqh (approved):

PR #2184, opened 2026-07-10
feat: add static model mirror support

### LINKED ISSUE #2165: Stream model from url in annotation
Users want to provide their own storage with model weights to stream.
Have the workspace accept an annotation of a storage url where vLLM directly uses runaistreamer to stream the weights from this url to GPU.

**Additional context**

2170chm:
**Reason for Change**:

Summary
Adds SAS-authenticated blob streaming. It's an optimization on the existing Model Mirror + Streaming feature that streams vLLM model weights directly from a pre-existing external Azure blob, skipping the download-to-PVC step. When a Workspace carries the required inference.kaito.sh/stream-* annotations, KAITO creates a static ModelMirror CR (no PVC, no download Job) and runs a per-pod fetch-sas init container that mints a SAS token at pod start via Workload Identity → AAD → datarefs, then streams from az:// with --load-format=runai_streamer.

Key changes

- New provider path: SASBlobProvider selected via SelectModelStreamer when the core SAS annotations are present; falls back to the existing HF/mirror provider otherwise (modelstreaming/azure/sasblob.go, modelstreaming/modelstreaming.go, selector in modelstreaming/registry).

- Annotation contract: modelstreaming/annotations.go centralizes the AnnotationStream* consts + Has/Require/CountSASBlobStreamingAnnotations. The core-required set is stream-uri, stream-account, stream-datarefs-url, stream-blob-uri, stream-identity-client-id (all together or none); stream-asset-id and stream-token-audience are optional/path-specific. Partial (some-but-not-all core) sets fail fast with an "incomplete SAS blob streaming configuration" error.

- Static CR: buildStaticModelMirror builds a registry=azureml, mode=Static CR with nil StorageClassName and empty JobNamespace; the mirror controller marks it Ready immediately with empty modelPath (no PVC/Job/finalizer).

- Auth validation: SAS path validates the same prerequisites as the mirror baseline. StorageClass + CSI provisioner check runs unconditionally, and ValidateAuth verifies the streaming ServiceAccount + Workload Identity annotation.

- Init container + hand-off: a slim python:3.12-slim fetch-sas init container pip-installs azure-identity at runtime and receives fetch_sas.py via an env var (embedded with //go:embed). It builds the credential from an explicit identity client ID (WorkloadIdentityCredential) and writes the SAS as an env file to a memory-backed emptyDir at /mnt/streaming/env. The main container runs a transparent entrypoint wrapper (stream-entrypoint.sh, baked into the base image) that sources that env file then execs the original vLLM command..

**Requirements**

**Issue Fixed**:

Closes #2165 
**Notes for Reviewers**:

Following manual e2e tests performed and passed:

Test 1 — SAS happy path: phi-4 on A100, skip-download CR (Ready, no PVC/Job), fetch-sas init container mints token, streams from az://, inference served.
Test 2 — Partial annotations → fail: 4-of-5 annotations blocks with "incomplete SAS blob streaming configuration"; no silent fallback, no CR.
Test 3 — SAS mint failure → pod blocked: bad datarefs URL fails the init container; pod stuck in Init, no fallback download.
Test 4 — Opt-out wins: model-streaming: disabled overrides all 5 SAS annotations. Normal download path, no fetch-sas, no az://.
Test 5 — Multi-node distributed SAS: phi-4 spilled to 2× A10; per-pod token mint on leader+worker, pp2, 243/243 shards streamed on the ray worker, inference served.
R1 — Mirror regression: plain streaming workspace still uses PVC/download path (registry=huggingface, real modelPath), no fetch-sas.
R2 — DAR regression: model-streaming: disabled still downloads to weights PVC, no az://, no fetch-sas.
Health sweep: controller 1/1, no panics/crashloops, no stray CRs.


api/v1alpha1/modelmirror_types.go
```diff
@@ -39,14 +39,27 @@ type ModelMirrorSpec struct {
 	// +kubebuilder:validation:Required
 	Storage ModelMirrorStorage `json:"storage"`
 	// JobNamespace is the namespace where the PVC and download Job will be created.
-	// +kubebuilder:validation:Required
-	JobNamespace string `json:"jobNamespace"`
+	// Empty for stream-only sources that create no PVC or Job.
```
Fei-Guo: Change `steam-only` to "static model mirror" in the whole file.

api/v1alpha1/modelmirror_types.go
```diff
@@ -57,12 +70,14 @@ type ModelMirrorSource struct {
 }
 
 type ModelMirrorStorage struct {
-	// Size is the PVC size (e.g. "20Gi").
+	// Size is the requested model storage: the PVC size for mirror (huggingface)
+	// sources, or informational for stream-only (azureml) sources that create no PVC.
```
Fei-Guo: remove azureml in the comment.

api/v1alpha1/modelmirror_types.go
Fei-Guo, line 36: Let us add a new field in ModelMirrorSpec to represent a "static" model mirror.
// +kubebuilder:validation:Enum=true;false
// + optional, default false.
 Static string `json:"static"`

A static model mirror maps to an existing (BYO) storage where the model weights are already available. The storage url is represented in status.modelpath. Controller will not create download job for static model mirror.

pkg/modelmirror/controllers/modelmirror_controller.go
```diff
@@ -72,6 +72,19 @@ func (r *ModelMirrorReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return r.handleDeletion(ctx, cr)
 	}
 
+	// Skip-download source (no StorageClass): there is no PVC to provision and no weights
+	// to download, so mark Ready immediately and never create a finalizer, PVC, or Job.
+	if cr.Spec.Storage.StorageClassName == nil {
```
Fei-Guo: change to check if it is static mirror.

pkg/workspace/inference/modelstreaming/modelstreamer.go
```diff
@@ -37,6 +37,12 @@ type StreamingConfig struct {
 	// PodLabels are provider-specific labels to add to the pod template
 	// (e.g. azure.workload.identity/use for Azure WI; empty for AWS/GCP).
 	PodLabels map[string]string
+	// InitContainers are provider-contributed init containers (e.g. a SAS-token
```
Fei-Guo, line 38: Let us consider reorganizing the files in this directory.
Move modelstreamer.go to
inference/modelstreaming/modelstreamer.go
for all azure specific code to  
inference/modelstreaming/azure/modelstea*.go

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -252,6 +264,32 @@ func (c *WorkspaceReconciler) ensureModelMirror(ctx context.Context, wObj *kaito
 	return nil
 }
 
+// buildStreamOnlyModelMirror constructs a skip-download ModelMirror CR: the model weights
+// live in a pre-existing external location, so the mirror creates no PVC and runs no
+// download Job (signalled by a nil StorageClassName and empty JobNamespace).
+func buildStreamOnlyModelMirror(
```
Fei-Guo: rename to buildStaticModelMirror.

pkg/workspace/inference/modelstreaming.go
```diff
@@ -106,31 +106,67 @@ func buildCommonStreamingEnvVars(modelID string) []corev1.EnvVar {
...
+			spec.Containers[mainIdx].VolumeMounts = append(spec.Containers[mainIdx].VolumeMounts, corev1.VolumeMount{
+				Name:      sasSharedVolumeName,
+				MountPath: sasSharedMountPath,
+			})
+
+			// Prepend a shell export so the inference process sees the SAS token written by the
+			// init container. The command is expected to be [/bin/sh, -c, <cmd>] (produced by
+			// utils.ShellCmd)
+			cmd := spec.Containers[mainIdx].Command
+			if len(cmd) == 3 && cmd[0] == "/bin/sh" && cmd[1] == "-c" {
```
Fei-Guo: This looks very hacky. May need to consider change the main container cmd.

api/v1alpha1/modelmirror_types.go
zhuangqh: my lesson from last api review: be careful to introduce a boolean field. 
Instead of adding a static boolean, consider introducing a `mode` field (e.g. Managed / Static or Mirror / Reference). It better describes the resource behavior and leaves room for future extension.

pkg/workspace/inference/modelstreaming.go
```diff
@@ -106,31 +106,67 @@ func buildCommonStreamingEnvVars(modelID string) []corev1.EnvVar {
...
+				tokenPath := sasSharedMountPath + "/" + sasTokenFileName
```
zhuangqh: Consider adding a transparent wrapper. 
notes: honor kaito workspace dir structure. config are placed under /mnt/config/

```
#!/bin/sh
set -eu

# Load environment variables if present.
if [ -f /mnt/config/env ]; then
    set -a
    . /mnt/config/env
    set +a
fi

# Replace the shell with the target command.
exec "$@"
```

api/v1alpha1/modelmirror_validation.go
```diff
@@ -45,17 +47,33 @@ func (m *ModelMirror) Validate(ctx context.Context) (errs *apis.FieldError) {
 	}
 
 	// Value validation (presence enforced by CRD schema)
-	if m.Spec.Source.Registry != "huggingface" {
+	if !slices.Contains(SupportedRegistries, m.Spec.Source.Registry) {
```
zhuangqh: This check is redundant technically since you added `+kubebuilder:validation:Enum=huggingface;azureml`.
But it's ok to leave it.

api/v1alpha1/modelmirror_validation.go
```diff
@@ -45,17 +47,33 @@ func (m *ModelMirror) Validate(ctx context.Context) (errs *apis.FieldError) {
...
 		errs = errs.Also(apis.ErrInvalidValue(m.Spec.Storage.Size, "spec.storage.size"))
 	}
-	sc := &storagev1.StorageClass{}
-	if err := k8sclient.Client.Get(ctx, types.NamespacedName{Name: m.Spec.Storage.StorageClassName}, sc); err != nil {
-		errs = errs.Also(apis.ErrInvalidValue(m.Spec.Storage.StorageClassName, "spec.storage.storageClassName"))
+	// StorageClass rules differ by mode:
+	//   - Managed mirrors download to a PVC and require an existing StorageClass.
+	//   - Static mirrors map to BYO storage, create no PVC, so StorageClass must be absent.
+	// Mode defaults to Managed (via the CRD schema); treat an empty value as Managed.
+	static := m.Spec.Mode == ModelMirrorModeStatic
```
zhuangqh: it's ok but I prefer a switch case here.

switch m.Spec.Mode {
case static:
  validateStaticMirrorMode(xxxx)
case managed:
validatemanagedmirrormode(xxx)

pkg/workspace/inference/modelstreaming/azure/auth.go
```diff
@@ -0,0 +1,44 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/kaito-project/kaito/api/v1beta1"
+)
+
+// ValidateStreamingServiceAccount verifies the streaming ServiceAccount exists in the
+// workspace namespace and carries the Azure Workload Identity client-id annotation.
+// Shared by different streaming providers.
+func ValidateStreamingServiceAccount(ctx context.Context, ws *v1beta1.Workspace, kubeClient client.Client, defaultSA string) error {
```
zhuangqh, line 31: I think this file should be placed under azure/ folder.

pkg/workspace/inference/modelstreaming/azure/sasblob.go
```diff
@@ -0,0 +1,105 @@
...
+func (s *SASBlobProvider) GetStreamingConfig(ctx *generator.WorkspaceGeneratorContext, modelID string) (*modelstreaming.StreamingConfig, error) {
+	ann := ctx.Workspace.Annotations
+
+	envFilePath := modelstreaming.SASSharedMountPath + "/" + modelstreaming.SASEnvFileName
+	initContainer := corev1.Container{
+		Name:    modelstreaming.SASFetchInitContainerName,
+		Image:   sasInitImage,
+		Command: []string{"/bin/sh", "-c", initShellCommand},
+		Env: []corev1.EnvVar{
+			{Name: "FETCH_SAS_SCRIPT", Value: fetchSASScript},
```
zhuangqh, line 70: A potential bug here. linux env value size have a limit. usually 2MB. it's acceptable for this case. add a comment to describe this trick.

pkg/workspace/controllers/workspace_controller.go
zhuangqh, line 753: I feel you break the existing pattern to report workspace status. 
will it be reasonable to group the sas token issue into WorkspaceConditionTypeModelMirrorReady condition since it's related to modelmirror functionality.
cc @rambohe-ch

zhuangqh (commented):
not sure if it will be clearer to rename pr title to something like
feat: add static model mirror support

pkg/workspace/controllers/workspace_controller.go
2170chm: per our discussion, we'll keep it as is.

api/v1alpha1/modelmirror_types.go
```diff
@@ -57,12 +91,14 @@ type ModelMirrorSource struct {
 }
 
 type ModelMirrorStorage struct {
-	// Size is the PVC size (e.g. "20Gi").
+	// Size is the requested model storage: the PVC size for managed mirrors, or
```
Fei-Guo: "informational for static mirrors that create no PVC." this seems to be unnecessary because your webhook already blocks adding modelmirrorstorage for static mode.

pkg/workspace/inference/modelstreaming/annotations.go
```diff
@@ -0,0 +1,95 @@
...
+func HasSASBlobStreamingAnnotations(annotations map[string]string) bool {
+	return CountSASBlobStreamingAnnotations(annotations) == len(coreSASBlobStreamingAnnotationKeys)
+}
+
+// RequireStaticModelMirror enforces the static-mirror sas annotations.
+func RequireStaticModelMirror(annotations map[string]string) error {
+	if !StaticModelMirrorEnabled(annotations) {
+		return nil
+	}
+	if HasSASBlobStreamingAnnotations(annotations) {
```
Fei-Guo: Maybe we can in-line this simple HasSASBlobStreamingAnnotations method here since it is only used once in this file.

pkg/workspace/inference/modelstreaming/annotations.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// HasSASBlobStreamingAnnotations reports whether all core stream-* annotations are
+// present and non-empty.
+func HasSASBlobStreamingAnnotations(annotations map[string]string) bool {
+	return CountSASBlobStreamingAnnotations(annotations) == len(coreSASBlobStreamingAnnotationKeys)
+}
+
+// RequireStaticModelMirror enforces the static-mirror sas annotations.
+func RequireStaticModelMirror(annotations map[string]string) error {
```
Fei-Guo: RequireStaticModelMirror -> ValidateStaticModelMirrorAnnotations.

Fei-Guo (approved):

PR #2186, opened 2026-07-10
feat: add NVIDIA MIG support for BYO nodes

### LINKED ISSUE #1744: Feature Request: Support for NVIDIA Multi-Instance GPU (MIG)

zhehli688:
**Reason for Change**:

Adds NVIDIA Multi-Instance GPU (MIG) support so inference workloads can run on a single GPU *slice* instead of a whole card, letting several small models share one large GPU (e.g. multiple `phi-4-mini` replicas on one H100/A100). MIG is enabled behind the `enableMIG` feature gate for **bring-your-own (BYO) nodes** and works with both NVIDIA device-plugin strategies (**mixed** and **single**).

What's included:

- **API** — `MIGSpec{ Profile string }` on `Workspace.ResourceSpec` and `InferenceSet` template (`v1alpha1` + `v1beta1`); `InstanceType` made optional to allow BYO. Regenerated CRDs/deepcopy.
- **Feature gate** — `enableMIG` (default off) in consts, featuregates, and chart values.
- **Validation** (webhook) — structural checks only: feature gate, BYO-only, known MIG profile, tensor-parallel rejection, tuning rejection, immutability. Extracted into a `validateMIG` helper.
- **Scheduling** — mixed strategy requests the per-profile extended resource `nvidia.com/mig-<profile>` + MIG toleration; single strategy keeps `nvidia.com/gpu`.
- **MIG node detection** — `GetGPUConfigFromNodeLabels` sets `IsMIG` from `nvidia.com/mig.config != "all-disabled"` AND `mig.config.state == "success"` (works for both strategies).
- **Estimator** — authoritative single-slice memory-fit check with a detailed overhead model (weight expansion + base overhead + KV cache + weight-scaled term vs `slice × 0.84`), extracted into `estimateMIGNodeCount`.
- **Reliability** — disables LMCache CPU KV-cache offload for MIG pods (`kaito-kv-cache-cpu-memory-utilization=0`) to prevent host-RAM OOM on shared MIG nodes.
- **Preset (Python)** — `get_max_gpu_memory_utilization()` now uses `torch.cuda.mem_get_info()` instead of `pynvml`, which fails / misreports on MIG slices.
- **InferenceSet** — propagates `MIG` to child Workspaces; delegates MIG validation to the shared helper.
- **Docs & examples** — new user guide multi-gpu-instance.md (+ sidebar), mixed/single strategy examples, and a refreshed design proposal.

**Requirements**

Unit tests added for the `mig` utility package, MIG webhook validation (Workspace + InferenceSet, both API versions), the estimator MIG path, node-label MIG detection, and the CPU-offload-disable behavior. Will add E2E tests (require A100) once the capacity issue in swedencentral is resolved.
- [x] manually verified both mixed and single strategies worked on an H100.

**Issue Fixed**:

Fixes #1744

**Notes for Reviewers**:
Also fixed deploy failures for workspace-mistral-3-*b models in this PR. These failures were pre-existing and caused by https://github.com/mistralai/mistral-common/issues/247. The fix was to stop passing `reasoning-parser` to vLLM for these non-reasoning models.


website/docs/multi-gpu-instance.md
```diff
@@ -0,0 +1,137 @@
...
+  template:
+    inference:
+      preset:
+        accessMode: public
+        name: "microsoft/Phi-4-mini-instruct"
+    resource:
+      mig:
+        profile: "2g.24gb"
+```
+
```
Fei-Guo, line 97: Please show the results running kubectl such that two workspace pods are running in the same GPU node.

api/v1alpha1/inferenceset_types.go
```diff
@@ -22,8 +22,16 @@ import (
...
+	// Required when node auto-provisioning is enabled; must be empty for BYO nodes
+	// (including MIG, which requires disableNodeAutoProvisioning=true).
+	// +optional
+	InstanceType string `json:"instanceType,omitempty"`
+
+	// MIG specifies NVIDIA Multi-Instance GPU configuration applied to each replica.
+	// Requires the enableMIG feature gate and BYO nodes (disableNodeAutoProvisioning=true).
+	// Propagated verbatim to each child Workspace.
+	// +optional
+	MIG *kaitov1beta1.MIGSpec `json:"mig,omitempty"`
```
Fei-Guo: "partition": {
        "kind": "mig", 
        "profile": "3g.40gb"
      }
I propose the above API changes.

api/v1alpha1/inferenceset_validation.go
```diff
@@ -57,11 +62,42 @@ func (is *InferenceSet) validateCreate() (errs *apis.FieldError) {
...
 	errs = errs.Also(validateMaintenanceWindow(is.Spec.AutoUpgrade))
+	errs = errs.Also(validateInferenceSetMIG(&is.Spec.Template.Resource).ViaField("template", "resource"))
 	return errs
 }
 
-func (is *InferenceSet) validateUpdate(_ *InferenceSet) (errs *apis.FieldError) {
+func (is *InferenceSet) validateUpdate(old *InferenceSet) (errs *apis.FieldError) {
 	errs = errs.Also(validateMaintenanceWindow(is.Spec.AutoUpgrade))
+	// MIG is immutable once set.
+	if !reflect.DeepEqual(is.Spec.Template.Resource.MIG, old.Spec.Template.Resource.MIG) {
```
zhuangqh: reflect.DeepEqual is usually not what you want. use `apiequality.Semantic.DeepEqual` instead.

api/v1alpha1/inferenceset_types.go
zhuangqh: +1

api/v1beta1/inferenceset_types.go
```diff
@@ -20,8 +20,16 @@ import (
 
 type InferenceSetResourceSpec struct {
 	// InstanceType specifies the GPU node SKU.
-	// +required
-	InstanceType string `json:"instanceType"`
+	// Required when node auto-provisioning is enabled; must be empty for BYO nodes
+	// (including MIG, which requires disableNodeAutoProvisioning=true).
```
zhuangqh: please avoid mention `disableNodeAutoProvisioning=true` in api.

pkg/workspace/estimator/nodesestimator/estimator.go
```diff
@@ -80,6 +83,12 @@ func (c *NodeEstimator) EstimateNodeCount(ctx context.Context, req estimator.Nod
 		return 0, fmt.Errorf("failed to get model by name: %w", err)
 	}
 
+	// MIG path: when MIG is configured, the model must fit in a single MIG slice.
+	// No tensor parallelism or multi-node distribution is possible with MIG.
+	if featuregates.FeatureGates[consts.FeatureFlagEnableMIG] && req.ResourceProfile.MIGProfile != "" {
```
zhuangqh: this package was imported by external lib. please avoid introduce featuregates package here.
add field to req instead.

presets/workspace/inference/vllm/inference_api.py
```diff
@@ -425,16 +425,19 @@ def get_max_gpu_memory_utilization(device_index: int = 0) -> float:
...
-    handle = pynvml.nvmlDeviceGetHandleByIndex(device_index)
-    info = pynvml.nvmlDeviceGetMemoryInfo(handle)
-    pynvml.nvmlShutdown()
+    #
+    # Use torch.cuda.mem_get_info() instead of pynvml: on MIG partitions NVML
+    # device-memory queries fail with NVMLError_NoPermission and, when they do
+    # succeed, report the parent GPU's memory rather than the slice's. torch
+    # reports the per-device (per-slice under MIG) free/total memory correctly
+    # and requires no elevated NVML permissions.
+    free_bytes, total_bytes = torch.cuda.mem_get_info(device_index)
```
zhuangqh: the reason why we dont use torch.cuda.mem_get_info.

https://github.com/kaito-project/kaito/pull/1405#discussion_r2290621592

`calling this method will initialize the underlying cuda context, which will consume an extra 230MB vram and cannot be released unless process exit.`

need to figure out how to deal with this issue...

website/docs/multi-gpu-instance.md
```diff
@@ -0,0 +1,137 @@
+---
```
zhuangqh, line 1: you are awesome! 
all changes included

presets/workspace/inference/vllm/inference_api.py
zhehli688: Thanks for the context! I put the memory probe in a subprocess so the 240MB will be auto-released. However, I just found out that the result of get_max_gpu_memory_utilization() is not being used because we explicitly pass `--gpu-memory-utilization=0.84` to vLLM. I filed a bug for this: https://github.com/kaito-project/kaito/issues/2191.

api/v1alpha1/inferenceset_types.go
```diff
@@ -22,8 +22,16 @@ import (
...
+	// Required when node auto-provisioning is enabled; must be empty for BYO nodes.
+	// +optional
+	InstanceType string `json:"instanceType,omitempty"`
+
+	// Partition specifies GPU partitioning applied to each replica. When set, each
+	// replica is scheduled on a GPU partition (slice) instead of a full GPU.
+	// Requires the enableMIG feature gate and BYO nodes.
+	// Propagated verbatim to each child Workspace.
+	// +optional
+	Partition *kaitov1beta1.PartitionSpec `json:"partition,omitempty"`
```
zhuangqh: we should update v1beta1 api only. v1alpha1 is archived

Fei-Guo (approved):

PR #2190, opened 2026-07-15
feat: pluggable model cache framework

hasethuraman:
KAITO has no first-class extension point for caching model weights. Today any weight-cache behavior would have to be hardcoded into the workspace/inference controllers and gated on runtime-specific signals (e.g. a particular  --load-format ), which doesn't scale to multiple backends and couples core logic to one implementation.

This PR introduces a pluggable cache framework:

• A  cache.Provider  contract with an optional  PodApplicabilityChecker  ( AppliesTo(concern, ws, *StatefulSet) ) so each provider decides when it applies to a workspace, rather than the framework hardcoding conditions.
• Framework wiring in the workspace/inference controllers that collects and applies provider mutations to the StatefulSet/pod template when a provider declares the workspace applicable.
• A  noop  provider registered by default — the framework ships with no functional backend, so behavior is unchanged out of the box.
•  ModelCacheReady  condition, CRD/API additions ( v1alpha1  +  v1beta1 ), regenerated CRDs/RBAC, and feature-gate plumbing.
• Conformance + integration + e2e test harnesses that any real provider can reuse to validate itself.
• Provider-agnostic documentation ( website/docs/distributed-cache.md ) covering the contract and how to implement/register a provider.

This lands the scaffolding independently so provider implementations can be reviewed and wired in separately.

**Requirements**
✓ added unit tests and e2e tests (if applicable).

**Issue Fixed:**
_NA_

**Notes for Reviewers:**
• Framework-only, no behavior change: only the  noop  provider is registered in  cmd/workspace/main.go ; with no real provider compiled in, workspaces behave exactly as before. Verified by building this branch's controller image and deploying a plain vLLM Workspace streaming from Azure Blob ( runai_streamer , no  cache:  section) — reached  InferenceReady=True , confirming the base streaming path is intact.
• Applicability is provider-declared: the framework no longer inspects  --load-format /runai_streamer. Providers implement  AppliesTo  to opt in; providers that don't implement  PodApplicabilityChecker  are always applicable. Not-applicable providers are skipped silently (no block, even in Required mode).
• New tests: unit/conformance under  pkg/cache/  ( conformance_test.go ,  integration_test.go ,  mutations_test.go , etc.) plus  test/e2e/cache_integration_test.go  (CI/GPU-only; cache e2e env vars are  CACHE_VLLM_IMAGE ,  CACHE_MODEL_URL ,  CACHE_STORAGE_ACCOUNT ).
• Generated files ( config/crd/bases/* ,  charts/kaito/**/crds ,  zz_generated.deepcopy.go ) are committed from  make generate manifests .
• No  v1beta1  fields were removed or renamed; only optional additions.


api/v1alpha1/inferenceset_types.go
```diff
@@ -37,6 +37,11 @@ type InferenceSetTemplate struct {
 	// +optional
 	Resource  InferenceSetResourceSpec   `json:"resource"`
 	Inference kaitov1beta1.InferenceSpec `json:"inference"`
+
+	// Cache configures distributed caching for all Workspaces created by this InferenceSet.
+	// Propagated to child Workspaces at creation time.
+	// +optional
+	Cache *kaitov1beta1.CacheSpec `json:"cache,omitempty"`
```
zhuangqh: v1alpha1 was archived. update v1beta1 only. Thanks.

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,221 @@
+---
```
zhuangqh: I prefer to defer the document until this feature is implemented.

pkg/cache/mutations.go
```diff
@@ -0,0 +1,547 @@
...
+	}
+	return checker.AppliesTo(concern, ws, ss)
+}
+
+// SetCacheMutations returns a pod spec modifier that injects cache provider
+// env vars, volumes, volume mounts, and init containers into the inference pod.
+// Returns nil (no modifier) when the feature gate is disabled or cache is not configured.
+func SetCacheMutations() generator.TypedManifestModifier[generator.WorkspaceGeneratorContext, corev1.PodSpec] {
+	return func(ctx *generator.WorkspaceGeneratorContext, spec *corev1.PodSpec) error {
+		if !featuregates.FeatureGates[consts.FeatureFlagDistributedCache] {
```
zhuangqh: how about guarding the code in the workspacereconciler/operator side?
not in the lib code here.

website/docs/distributed-cache.md
hasethuraman: Kept this document from the framework aspect. The document basically says how to implement a provider and how the configuration can be done in KAITO. I will remove it from website/sidebars.js at the moment. Please let me know otherwise

api/v1alpha1/inferenceset_types.go
hasethuraman: Done.

pkg/cache/mutations.go
hasethuraman: Done. Feature gate is a global one & hence moved to reconciler

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -362,6 +363,27 @@ func (c *WorkspaceReconciler) addOrUpdateWorkspace(ctx context.Context, wObj *ka
...
+	// the cache library itself stays feature-gate agnostic.
+	if featuregates.FeatureGates[consts.FeatureFlagDistributedCache] {
+		cacheResult := cache.ReconcileCache(ctx, c.Client, wObj, &wObj.Status)
+		if cacheResult.BlockDeployment {
+			klog.V(2).InfoS("Cache not ready, blocking deployment", "workspace", klog.KObj(wObj))
+			if cacheResult.RequeueNeeded {
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+			return reconcile.Result{}, nil
+		}
```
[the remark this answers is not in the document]
hasethuraman: readiness gating is provider-level and intentionally runs before render;  PodApplicabilityChecker  is a render-time, StatefulSet-shape-based injection decision — a separate concern that is not meant to influence Required-mode blocking

charts/kaito/workspace/values.yaml
```diff
@@ -187,4 +188,11 @@ local-csi-driver:
...
 # See https://github.com/kaito-project/production-stack for more details
 productionstack:
-  enabled: false
\ No newline at end of file
+  enabled: false
+
+# Distributed cache provider configuration.
+# Only active when featureGates.distributedCache is true.
+# Each cache provider is configured under cache.providers.<name>. No provider is
+# bundled with this framework; install a provider and add its config here.
```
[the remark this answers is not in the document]
hasethuraman: This documents the Helm-configurable backends under `cache.providers.<name>`, which are brought in out-of-tree — none ships in this chart, so the guidance matches shipped behavior.

pkg/cache/mutations.go
```diff
@@ -0,0 +1,538 @@
...
+			cacheName := extractCacheName(ctx, kubeClient, ws.Namespace, ws.Cache.KVCache.Config)
+			// Check infrastructure availability before injecting mutations.
+			available, availErr := p.IsAvailable(ctx, cacheName)
+			if availErr != nil || !available {
+				if ws.Cache.KVCache.Mode == kaitov1beta1.CacheModeRequired {
+					return nil, fmt.Errorf("KV cache provider %q not available", ws.Cache.KVCache.Provider)
+				}
+				klog.V(2).InfoS("KV cache provider not available, skipping mutations",
+					"provider", ws.Cache.KVCache.Provider, "available", available, "error", availErr)
+			} else {
```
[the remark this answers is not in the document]
hasethuraman: This is intentional. A provider that doesn't want to engage KV cache for a given workload can already opt out inside `PodMutations(CacheConcernKVCache, ...)` by returning no mutations — so the opt-out capability the contract promises is preserved. The dedicated `PodApplicabilityChecker` pre-check exists for model weights specifically because non-applicability there must avoid Required-mode *blocking* (we must not hold up deployment for a weights cache that can't engage). KV cache injection isn't gated at that blocking level, so it doesn't need the
pre-check. The two concerns intentionally have different flows; they aren't required to be symmetric.

pkg/cache/mutations.go
```diff
@@ -0,0 +1,538 @@
...
+		if dst.Labels == nil {
+			dst.Labels = make(map[string]string, len(src.Labels))
+		}
+		for k, v := range src.Labels {
+			dst.Labels[k] = v
+		}
+	}
+
+	// Deduplicate env vars by name. For path-list vars, merge by appending
+	// with colon separator. For all others, first wins (dst takes priority).
```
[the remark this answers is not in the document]
hasethuraman: Good catch — the comment was stale. `mergeMutations` only deduplicates env vars
by name (first wins, dst takes priority); there's no colon-separated path-list
merging. Fixed the comment to describe the actual behavior

pkg/cache/reconcile.go
```diff
@@ -0,0 +1,263 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
+)
+
+const (
+	// DefaultCacheReadyTimeout is the maximum time to block deployment in Required mode
+	// waiting for cache to become ready. After this, deployment proceeds without cache.
+	DefaultCacheReadyTimeout = 10 * time.Minute
+)
```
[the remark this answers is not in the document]
hasethuraman: Fixed by removing the timeout entirely so the code matches the documented
semantics. Required mode now blocks deployment indefinitely until the cache is
ready — no 10-minute safety valve, no "proceed without cache" path. Dropped
`DefaultCacheReadyTimeout` and both timeout branches, and updated
`distributed-cache.md` (removed the readiness-timeout section and Configuration
Reference row; the Required-mode row now simply states it blocks until the cache
is ready). A Required cache that never becomes ready is a hard block by design

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,221 @@
...
+    mode: Opportunistic
+```
+
+### Cache Modes
+
+Each concern supports three modes:
+
+| Mode | Behavior |
+|------|----------|
+| `Required` | Block pod deployment until cache infrastructure is ready. Workspace enters a waiting state if cache is unavailable. |
```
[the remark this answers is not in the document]
hasethuraman: Fixed by removing the timeout entirely so the code matches the documented
semantics. Required mode now blocks deployment indefinitely until the cache is
ready — no 10-minute safety valve, no "proceed without cache" path. Dropped
`DefaultCacheReadyTimeout` and both timeout branches, and updated
`distributed-cache.md` (removed the readiness-timeout section and Configuration
Reference row; the Required-mode row now simply states it blocks until the cache
is ready). A Required cache that never becomes ready is a hard block by design

charts/kaito/workspace/templates/clusterrole.yaml
```diff
@@ -89,6 +89,14 @@ rules:
...
     verbs: ["get", "list", "watch", "create", "update"]
   {{- end }}
+  {{- if .Values.featureGates.distributedCache }}
+  - apiGroups: ["storage.azure.com"]
+    resources: ["caches"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["storage.azure.com"]
+    resources: ["caches/status"]
+    verbs: ["get"]
+  {{- end }}
```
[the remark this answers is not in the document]
hasethuraman: Agreed — removed the `storage.azure.com` (`caches`/`caches/status`) rules from the
core chart. No framework code reads that API group; the only consumer is the
out-of-tree provider, which ships the identical RBAC in its own chart under the
same feature gate. The core chart now stays provider-agnostic and a noop-only
install no longer gets unused Azure permissions.

Verified by reinstalling the chart with `featureGates.distributedCache=true`: the
rendered `kaito-workspace` ClusterRole contains zero `storage.azure.com` entries
and the controller runs fine.

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,221 @@
...
+resource:
+  instanceType: "Standard_NC24ads_A100_v4"
+  labelSelector:
+    matchLabels:
+      apps: phi-4
+inference:
+  preset:
+    name: "microsoft/phi-4"
+cache:
+  modelCache:
```
andyzhangx: does modelCache already work with this PR?

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,221 @@
...
+    provider: <provider-name>
+    mode: Opportunistic
+  kvCache:
```
andyzhangx: remove kvCache related doc since it's not working now, it requires more changes.

test/e2e/cache_integration_test.go
```diff
@@ -0,0 +1,1237 @@
...
+
+	kaitov1beta1 "github.com/kaito-project/kaito/api/v1beta1"
+	"github.com/kaito-project/kaito/pkg/cache"
+	_ "github.com/kaito-project/kaito/pkg/cache/noop"
+	"github.com/kaito-project/kaito/pkg/model"
+	"github.com/kaito-project/kaito/pkg/utils/consts"
+	"github.com/kaito-project/kaito/test/e2e/utils"
+)
+
+// These tests validate the distributed cache integration with the workspace
```
andyzhangx: if this e2e test won't work now, you could remove this since both dacs-client and dacs provider need to be public accessible.

test/e2e/e2e_test.go
```diff
@@ -116,11 +116,15 @@ var _ = BeforeSuite(func() {
 
 	loadTestEnvVars()
 
-	err = copySecretToNamespace(aiModelsRegistrySecret, namespaceName)
-	Expect(err).NotTo(HaveOccurred())
+	if aiModelsRegistrySecret != "" {
```
andyzhangx: what's this change for?

test/e2e/e2e_test.go
```diff
@@ -116,11 +116,15 @@ var _ = BeforeSuite(func() {
...
+		err = copySecretToNamespace(aiModelsRegistrySecret, namespaceName)
+		Expect(err).NotTo(HaveOccurred())
+	}
 
-	err = copySecretToNamespace(e2eACRSecret, namespaceName)
-	Expect(err).NotTo(HaveOccurred())
+	if e2eACRSecret != "" {
```
andyzhangx: what's this change for?

test/e2e/preset_vllm_test.go
```diff
@@ -471,6 +471,34 @@ var _ = Describe("Workspace Preset on vllm runtime", func() {
 		validateModelsEndpoint(workspaceObj)
 		validateChatCompletionsEndpoint(workspaceObj)
 	})
+
+	It("should serve a public preset GPU workspace with no cache mutations when no cache is installed", utils.GinkgoLabelFastCheck, func() {
```
andyzhangx: not necessary e2e test since by default this feature is not enabled.

api/v1beta1/workspace_types.go
```diff
@@ -270,6 +270,80 @@ type WorkspaceStatus struct {
...
+type CacheProvider string
+
+// CacheMode controls how a workspace interacts with the cache.
+type CacheMode string
+
+const (
+	// CacheModeRequired blocks workload deployment until the cache is ready.
+	CacheModeRequired CacheMode = "Required"
+	// CacheModeOpportunistic uses the cache if available, proceeds without if not.
+	CacheModeOpportunistic CacheMode = "Opportunistic"
```
andyzhangx, line 295: what if dacs cache is not available when set as `Opportunistic`? where is the model file downloaded to in that case?

api/v1beta1/workspace_types.go
```diff
@@ -270,6 +270,80 @@ type WorkspaceStatus struct {
...
+	// CacheModeDisabled disables cache interaction entirely.
+	CacheModeDisabled CacheMode = "Disabled"
```
andyzhangx, line 297: is `Disabled` mode really needed? e.g. if this feature is enabled, while user set as  `Disabled` mode, where is the model file downloaded to in that case?

pkg/cache/test_conformance.go
```diff
@@ -0,0 +1,182 @@
+// Copyright (c) KAITO authors.
```
andyzhangx, line 1: what are these 3 test_xx.go files? for e2e tests instead of unit tests? are they useful in the KAITO repo?

pkg/cache/integration_test.go
```diff
@@ -0,0 +1,507 @@
+// Copyright (c) KAITO authors.
```
andyzhangx, line 1: what is this file? is it useful in the KAITO repo?

andyzhangx (commented):
pls make sure the files in pkg/cache are all used in KAITO repo, if it's for other project only, pls remove those useless files especially non unit test files.

andyzhangx (commented):
there should be a fix in this PR:
disable local csi driver when distribution cache is enabled since there would be conflict

test/e2e/cache_integration_test.go
hasethuraman: done

test/e2e/e2e_test.go
hasethuraman: reverted

test/e2e/preset_vllm_test.go
hasethuraman: Done

hasethuraman:
> there should be a fix in this PR: disable local csi driver when distribution cache is enabled since there would be conflict

we enable dacs and disable local csi driver as part of dacs PR and not in this PR. We can enable it in a follow up PR before we merge integration if required. We dont want to disable it by default till the integration PR is merged

test/e2e/e2e_test.go
hasethuraman: Reverted

website/docs/distributed-cache.md
hasethuraman: @andyzhangx this is the schema of framework that expects and documentation how a provider will have to be implemented and plugged in. Please let me know if any concern.

website/docs/distributed-cache.md
hasethuraman: Same as my earlier comment. It is the schema expectation from framework. When we add a new provider, we will call out what wont work. This is just a reference for provider.

pkg/cache/integration_test.go
hasethuraman: It runs as part of UT. Any concern?

pkg/cache/test_conformance.go
hasethuraman: It used by framework UT also. Please let me know if any concerns.

api/v1beta1/workspace_types.go
hasethuraman: if the provider is unavailable/unregistered/not-applicable we skip injection and the model loads via the normal non-cached path — no blocking. That's the Opportunistic contract

api/v1beta1/workspace_types.go
hasethuraman: Disabled  is a per-workspace escape hatch: if a specific workspace hits an unforeseen cache-layer issue (e.g. corruption), the user can disable caching for just that workspace as a workaround, while other healthy workspaces keep using the cache. It's more selective than flipping the global feature gate off.

andyzhangx: if there is cache-layer issue, user could just remove the `cache` field. what if `Opportunistic` fails, where is the model file downloaded to in that case?

```
    cache:
      modelCache:
        provider: dacs
        mode: Opportunistic
```

api/v1beta1/workspace_types.go
andyzhangx: you mean download to the storage inside the inference pod? I think we need to test on this failed mode, if `modelCache` is not set, it's using nvme PVC, while if `modelCache` is set and it fails, the large model download may fail since os disk cannot host such big size? that's just my assumption, need to verify here.

website/docs/distributed-cache.md
```diff
@@ -0,0 +1,221 @@
+---
+title: Distributed Cache
+---
+
+This document explains how to enable distributed caching for model weights and KV cache in KAITO using a pluggable cache provider.
```
andyzhangx: remove this doc in the PR since only this PR would not land modelCache feature

pkg/cache/test_e2e_scenarios.go
```diff
@@ -0,0 +1,103 @@
+// Copyright (c) KAITO authors.
```
andyzhangx: this file is not used in ut, pls remove this.

and pls check whether other test_*.go files are really used, we don't want to add unused go file in KAITO repo

website/docs/distributed-cache.md
hasethuraman: Done.

pkg/cache/test_e2e_scenarios.go
hasethuraman: Done. test_expectations is trimmed and remaining is used in UT context. The E2E test related code is removed from test_expectations.

hasethuraman: there is one more test_ file which is used only UT

pkg/cache/mutations.go
```diff
@@ -0,0 +1,538 @@
...
+}
+
+// applyMutations injects the collected mutations into the pod spec.
+// It deduplicates by name to avoid invalid PodSpecs with duplicate entries.
+func applyMutations(spec *corev1.PodSpec, mutations *PodMutations) {
+	if mutations == nil || len(spec.Containers) == 0 {
+		return
+	}
+
+	container := &spec.Containers[0]
```
andyzhangx, line 225: there could be multiple containers for inference pod

api/v1beta1/workspace_types.go
```diff
@@ -282,6 +282,80 @@ type WorkspaceStatus struct {
...
 
+// CacheProvider identifies a cache backend implementation.
+type CacheProvider string
+
+// CacheMode controls how a workspace interacts with the cache.
+type CacheMode string
+
+const (
+	// CacheModeRequired blocks workload deployment until the cache is ready.
+	CacheModeRequired CacheMode = "Required"
```
andyzhangx, line 293: will this `Required` value be really used in the future?

api/v1beta1/workspace_types.go
```diff
@@ -282,6 +282,80 @@ type WorkspaceStatus struct {
...
+	// KVCache configures caching of attention key/value tensors.
+	// +optional
+	KVCache *KVCacheSpec `json:"kvCache,omitempty"`
+}
+
+// ModelCacheSpec controls how model weight files are cached.
+type ModelCacheSpec struct {
+	// Provider selects the cache implementation for model weights.
+	// +kubebuilder:validation:MinLength=1
+	Provider CacheProvider `json:"provider"`
```
andyzhangx, line 317: what's the other provider that could be used? in kaito, we only need to support one provider during KAITO installation, I don't think we need to expose such option to the end user.

api/v1beta1/workspace_types.go
```diff
@@ -282,6 +282,80 @@ type WorkspaceStatus struct {
...
+	Config string `json:"config,omitempty"`
+
+	// CleanupOnDelete invalidates cached model data when workspace is deleted.
+	//
+	// NOTE: Not yet implemented. This field is reserved for future use; setting it
+	// currently has no effect. Cached model chunks are managed by the cache
+	// provider's own TTL/eviction policy and are not explicitly invalidated on
+	// workspace deletion. See https://github.com/kaito-project/kaito for status.
+	// +optional
+	CleanupOnDelete bool `json:"cleanupOnDelete,omitempty"`
```
andyzhangx, line 337: since this is optional field and not used now, I would recommend deleting this field for simplicity.

api/v1beta1/workspace_types.go
```diff
@@ -282,6 +282,80 @@ type WorkspaceStatus struct {
...
+	// Mode controls cache behavior.
+	// +kubebuilder:default:="Opportunistic"
+	// +kubebuilder:validation:Enum=Required;Opportunistic;Disabled
+	Mode CacheMode `json:"mode,omitempty"`
+
+	// Config is the name of a ConfigMap in the same namespace containing
+	// provider-specific model cache configuration. The provider validates
+	// and merges this with its defaults from Helm values.
+	// +optional
+	Config string `json:"config,omitempty"`
```
andyzhangx, line 328: since this is optional field and not used now, I would recommend deleting this field for simplicity.

api/v1beta1/workspace_types.go
```diff
@@ -282,6 +282,80 @@ type WorkspaceStatus struct {
...
+	// NOTE: Not yet implemented. This field is reserved for future use; setting it
+	// currently has no effect. Cached model chunks are managed by the cache
+	// provider's own TTL/eviction policy and are not explicitly invalidated on
+	// workspace deletion. See https://github.com/kaito-project/kaito for status.
+	// +optional
+	CleanupOnDelete bool `json:"cleanupOnDelete,omitempty"`
+}
+
+// KVCacheSpec controls how attention KV tensors are cached.
+type KVCacheSpec struct {
```
andyzhangx, line 341: since this is optional field and not used now, I would recommend deleting this field for simplicity.

andyzhangx (commented):
pls remove kv cache related settings since they are not supported now, we could add it in future versions. And to make the KAITO InferenceSet simple enough, what about set like following:
> we also need to make sure that as long as the `inference.preset.name` is the same, the two KAITO InferenceSet could share the same dacs cache model data in one AKS cluster:

```yaml
piVersion: kaito.sh/v1beta1
kind: InferenceSet
metadata:
  name: qwen3-coder-30b-a3b-instruct
spec:
  replicas: 1
  labelSelector:
    matchLabels:
      apps: qwen3-coder-30b-a3b-instruct
  template:
    enableModelCache: true # default is false
    resource:
    inference:
      preset:
        name: Qwen/Qwen3-Coder-30B-A3B-Instruct
```

as comparison, the original yaml format is like following (we don't need to expose provider to the end user, and we only need to set true or false for this model cache field, `required` is not used IMO since as long as there is cache miss, just download from blob storage) :

```yaml
apiVersion: kaito.sh/v1beta1
kind: InferenceSet
metadata:
  name: qwen3-coder-30b-a3b-instruct
spec:
  replicas: 1
  labelSelector:
    matchLabels:
      apps: qwen3-coder-30b-a3b-instruct
  template:
    cache:
      modelCache:
        provider: dacs
        mode: Opportunistic
    resource:
    inference:
      preset:
        name: Qwen/Qwen3-Coder-30B-A3B-Instruct
```

Fei-Guo:
or a simple API like:
```
apiVersion: kaito.sh/v1beta1
kind: InferenceSet
metadata:
  name: qwen3-coder-30b-a3b-instruct
spec:
  replicas: 1
  labelSelector:
    matchLabels:
      apps: qwen3-coder-30b-a3b-instruct
  template:
    cacheSpec:
      enabled: true # default is false
    resource:
    inference:
      preset:
        name: Qwen/Qwen3-Coder-30B-A3B-Instruct
```
We can add extra field if extra info is needed to hook the model deployment to the cache server that runs inside the same namespace.

PR #2193, opened 2026-07-16
feat: make gpu usage configurable

henrichter-sap:
**Reason for Change**:

On BYO nodes a preset Workspace always requests every GPU on the matched node, so a small model like `phi-4-mini` occupies the whole box. This adds optional `resource.requests`/`resource.limits` to override the pod's GPU count; the count is also threaded into `SKUNumGPUs` so vLLM's parallelism matches the GPUs the pod gets. `limits` defaults to `requests`; behavior is unchanged when neither is set.

I saw the MIG work (#2186) for sub-GPU sharing, but we're on bare metal nodes without MIG enabled. This is a whole GPU override, complementary to MIG.

**Requirements**

**Issue Fixed**:

Related to #1744

**Notes for Reviewers**:

Only exercised on the BYO path. Unsure how this interacts with Karpenter/node auto-provisioning; open to BYO gating it.


Fei-Guo:
This only applies for BYO because it will largely impact how we calculate number of required nodes in the NAP case. 

Hence, I think the feature should be guarded by BYO feature gate. We are reluctant to open resource request/limit in API for user to configure because it is hard to be covered in tests and error prone. For this kind of customization, the workaround is to use pod template instead of preset. 

However, for your case, I am open to introduce a special annotation in IS/WS which is guarded by BYO feature gate to allow customizing the resource request/limit of the inference pod.

henrichter-sap:
Thanks for the detailed reply, this makes sense. Agreed on feature-gating this to BYO, and I understand the concern about exposing an arbitrary `ResourceList`.

On the pod template workaround: if I understood correctly, that requires overriding the *entire* template. We'd like to keep using the preexisting preset templates and only adjust the GPU count, so replacing the whole template isn't ideal for us.

I understand an annotation would be the simplest implementation for now. However, if you're not strictly against it, I'd prefer a first-class CRD field, consistent with how MIG is exposed.

One idea: since GPU partitioning (#2186) just landed, could we reuse that abstraction directly? We could introduce `partition.mode: count` alongside the existing `mig` mode, and add a `count` field:

```yaml
resource:
 partition:
   mode: count
   count: 2
```

This stays first-class and typed like MIG, and reuses the existing PartitionSpec validation. Would you be open to this?

Fei-Guo:
> Thanks for the detailed reply, this makes sense. Agreed on feature-gating this to BYO, and I understand the concern about exposing an arbitrary `ResourceList`.
> 
> On the pod template workaround: if I understood correctly, that requires overriding the _entire_ template. We'd like to keep using the preexisting preset templates and only adjust the GPU count, so replacing the whole template isn't ideal for us.
> 
> I understand an annotation would be the simplest implementation for now. However, if you're not strictly against it, I'd prefer a first-class CRD field, consistent with how MIG is exposed.
> 
> One idea: since GPU partitioning (#2186) just landed, could we reuse that abstraction directly? We could introduce `partition.mode: count` alongside the existing `mig` mode, and add a `count` field:
> 
> ```yaml
> resource:
>  partition:
>    mode: count
>    count: 2
> ```
> 
> This stays first-class and typed like MIG, and reuses the existing PartitionSpec validation. Would you be open to this?

This is definitely better than directly configuring request/limit. Something like 

 ```yaml
    resource:
    partition:
      mode: accelerator 
      count: 2
```
I am ok with it.

henrichter-sap:
Implemented like that in #2236

PR #2201, opened 2026-07-20
fix: scope workspace node listing to owning workspace

rambohe-ch:
**Reason for Change**:

Multiple workspaces created by an InferenceSet share the same Resource.LabelSelector, so listing nodes by that selector alone could count nodes owned by sibling workspaces, inflating readiness/status counts and letting custom-template pods schedule onto a sibling's node.

- Add WorkspaceNodeSelector/ListWorkspaceNodes helpers that merge the user selector with the provisioner's BuildNodeSelector ownership labels, and route GetReadyNodes, gpu/byo provisioners and the workspace controller node-status snapshot through them.
- countCoveredNodes now excludes nodes owned by a different workspace while keeping BYO/legacy nodes in scope.
- Extract ApplyProvisionerNodeSelector and apply it to the custom template inference path so those pods are pinned to workspace nodes.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


rambohe-ch:
@Fei-Guo @zhuangqh PTAL

zhuangqh (approved):
Thank you for addressing this.

PR #2204, opened 2026-07-21
feat: rework static mirror sas fetching logic

### LINKED ISSUE #2202: [Feature] Improve sas fetching logic
Obtain the blob uri, stream uri, and storage account name fetching from Kaito pod init container instead of from annotations.

**Additional context**

2170chm:
**Reason for Change**:

- Slashed the RP annotation set to three (stream-datarefs-url, stream-identity-client-id, stream-source-type), with source-type validation.

- fetch_sas.py now derives everything at runtime. It resolves the model, parses the blob URI/account, discovers the safetensors subpath by listing the container, and mints the SAS.

- sasblob.go delivers the script through a per-workspace ConfigMap and sets the model path to a $STREAM_MODEL_URI runtime placeholder.

- interface.go renders --model as the shell placeholder so vLLM streams from the derived URI at pod start.

Note:
Added a paginated blob-listing call to find the weights subpath (by matching .safetensors files). Different models weights are under different subpaths so we can't hard-code it.

**Requirements**

**Issue Fixed**:

Closes #2202 
**Notes for Reviewers**:


Fei-Guo (approved):

pkg/workspace/controllers/workspace_controller_test.go
```diff
@@ -1020,11 +1020,9 @@ func TestEnsureModelMirror_StaticWithPartialSASFails(t *testing.T) {
 			Namespace: "default",
 			Annotations: map[string]string{
 				modelstreaming.AnnotationStaticModelMirror: "true",
-				modelstreaming.AnnotationStreamURI:         "az://c/model",
-				modelstreaming.AnnotationStreamAccount:     "acct",
-				modelstreaming.AnnotationStreamDatarefsURL: "https://x/datarefs",
-				modelstreaming.AnnotationStreamBlobURI:     "https://acct.blob.core.windows.net/c/prefix",
-				// inference.kaito.sh/stream-identity-client-id intentionally omitted (4 of 5 core).
+				modelstreaming.AnnotationStreamDatarefsURL: "https://x/models/m/versions/1/credentials",
```
bfoley13: I think this wont include `/credentials` correct? You need the base URL and you can apply credentials based on it being Foundry.

pkg/workspace/inference/modelstreaming/azure/fetch_sas.py
```diff
@@ -11,38 +11,93 @@
...
+  5. List the container with the SAS to discover the safetensors subpath -> model streaming URI.
+  6. Write AZURE_STORAGE_SAS_TOKEN, AZURE_STORAGE_ACCOUNT_NAME, and STREAM_MODEL_URI to the
+     shared env file so the main container's entrypoint wrapper can source them.
 
 Required environment variables:
-    STREAM_DATAREFS_URL       - SAS-mint endpoint URL (datarefs for public, /credentials for BYO)
-    STREAM_BLOB_URI           - blob URI for the request body
-    STREAM_IDENTITY_CLIENT_ID - workload identity client ID to mint the SAS as
+    STREAM_DATAREFS_URL       - SAS-mint endpoint URL
+    STREAM_IDENTITY_CLIENT_ID - workload identity client ID to resolve/mint as
```
bfoley13, line 30: Will we ever need tenantId for the request as well? I think we might want to check on the case of multiple identities being assigned to a single service account, how to we fetch the correct perms with the correct identity? Is it only clientId?

pkg/workspace/inference/modelstreaming/azure/fetch_sas.py
```diff
@@ -11,38 +11,93 @@
...
+    SOURCE_PUBLIC: "https://management.azure.com",
+    SOURCE_BYO: "https://ai.azure.com",
+}
+
+
+def resolve_url_from_datarefs(datarefs_url: str, source_type: str) -> str:
+    """Derive the model-resolve URL from the SAS-mint URL.
+
+    byo:    .../models/{m}/versions/{v}/credentials?... -> strip the trailing '/credentials' segment.
+    public: .../registries/{r}/datarefs/{m}/versions/{v}?... -> swap '/datarefs/' for '/models/'.
```
bfoley13: I guess Im questioning why not just make the input this shape(env vars) for both byo and public? At least for the byo I dont think it should have credentials to begin with, you can add that when you need it

bfoley13 (approved):
Looks good

Fei-Guo (approved):

PR #2205, opened 2026-07-21
feat: add queue-depth rate limit guard to vLLM runtime

zhuangqh:
**Reason for Change**:

Installed via vLLM's --middleware extension. Returns HTTP 429 on guarded inference endpoints when the waiting queue exceeds max_num_seqs, hedged against the lagging Prometheus metric with a process-local in-flight counter.

Test: 4096 requests at 1024 concurrent via guidellm:

  guidellm benchmark \
    --target http://localhost:5000 \
    --rate-type concurrent \
    --rate 1024 \
    --max-requests 4096 \
    --processor google/gemma-4-e4b-it \
    --data '{"prompt_tokens": 512, "output_tokens": 64}'

Without rate limit — all 4096 pile into the waiting queue:

  Engine 000: Running: 120 reqs, Waiting: 904 reqs, KV cache: 6.3%
  Req: 32.2 req/s, 28.89s Lat, 930.7 Conc, 4096 Comp, 0 Err
  Tok: 2065.3 gen/s, 18876.6 tot/s

With rate limit — excess requests rejected early:

  Req: 31.9 req/s, 12.95s Lat, 413.8 Conc, 712 Comp, 3384 Err
  Tok: 2069.3 gen/s, 18912.8 tot/s

Note: throughput is unchanged (32.2 vs 31.9 req/s, 2065 vs 2069 gen/s, 18876 vs 18912 tot/s) — the guard only sheds work the scheduler would have queued anyway, so benchmark results measuring server throughput are not affected.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


docker/presets/models/tfs/Dockerfile
```diff
@@ -63,6 +63,7 @@ COPY presets/workspace/inference/vllm/inference_api.py \
     presets/workspace/inference/vllm/multi-node-health-check.py \
     presets/workspace/inference/vllm/list_supported_llm_archs.py \
     presets/workspace/inference/vllm/benchmark_entrypoint.py \
+    presets/workspace/inference/vllm/rate_limit.py \
```
zhehli688, line 66: Don't we need to bump the base image version for this change?

rambohe-ch (approved):

docker/presets/models/tfs/Dockerfile
zhuangqh: we should

presets/workspace/inference/vllm/rate_limit.py
```diff
@@ -0,0 +1,128 @@
...
+                local_waiting,
+                effective,
+                threshold,
+            )
+            return JSONResponse(
+                status_code=429,
+                content={
+                    "error": {
+                        "message": (
+                            f"Server overloaded: waiting queue "
```
Fei-Guo: I think standard msg for 429 is "Too many requests" instead of "Server overloaded".

docker/presets/models/tfs/Dockerfile
zhuangqh: regarding the base image tag bump, I would like to refactor the workflow. every time we update py code. we need to update the tag, this is annoying. so, leave it here.

Fei-Guo (approved):

PR #2239, opened 2026-07-30
fix: cve bump setuptools>=83.0.0 and pin msgpack>=1.2.1

andyzhangx:
## What this PR does

Bumps two Python dependencies in `presets/workspace/dependencies/requirements.txt` to pull in security fixes flagged by trivy:

| Library    | CVE / Advisory                                                       | Severity | Fixed in |
|------------|----------------------------------------------------------------------|----------|----------|
| msgpack    | GHSA-6v7p-g79w-8964 (OOB read / crash on `Unpacker` reuse)           | HIGH     | 1.2.1    |
| setuptools | CVE-2025-47273 (path traversal in `PackageIndex`)                    | HIGH     | 78.1.1   |
| setuptools | CVE-2026-59890 (`MANIFEST.in` exclusion bypass via NFC/NFD collision)| MEDIUM   | 83.0.0   |

## Changes

- `setuptools` was previously unpinned; changed to `setuptools>=83.0.0` (covers both CVE-2025-47273 and CVE-2026-59890) with a short inline comment referencing the CVE IDs.
- `msgpack` was not previously listed; it comes in transitively via `ray[default]==2.55.0` at version `1.1.2`. Added an explicit floor `msgpack>=1.2.1` so pip resolves at or above the fix.

## Why an explicit `msgpack` line

`ray[default]==2.55.0` currently pins `msgpack~=1.0` and picks the latest matching wheel (`1.1.2`, still vulnerable). Adding an explicit floor is the least invasive fix; if a future `ray` release moves off the vulnerable line we can drop this pin again.

## Validation

- `pip install -r requirements.txt` resolves cleanly to `msgpack==1.2.1+` and `setuptools>=83.0.0`.
- Trivy scan against the rebuilt image shows all three findings marked as resolved.

/kind bug
/sig storage
/assign @andyzhangx

### DIFF
--- a/docker/presets/models/tfs/Dockerfile
+++ b/docker/presets/models/tfs/Dockerfile
@@ -40,6 +40,7 @@ RUN pip3 install --no-cache-dir --upgrade "setuptools>=83.0.0" "msgpack>=1.2.1" 
            /usr/local/lib/python3.12/site-packages/pip/_vendor/msgpack* \
            /usr/local/lib/python3.12/site-packages/pip/_vendor/setuptools* \
            /usr/local/lib/python3.12/site-packages/pip/_vendor/pkg_resources* \
+           /usr/local/lib/python3.12/site-packages/pip/_vendor/vendor.txt \
            /root/.cache/pip && \
     find / -type d \( \
         -name 'setuptools-6*.dist-info' -o \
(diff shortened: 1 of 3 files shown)


docker/presets/models/tfs/Dockerfile
```diff
@@ -53,6 +53,17 @@ RUN --mount=type=cache,target=/root/.cache/pip \
     uv pip install --system -q --reinstall-package nixl-cu12 nixl-cu12 && \
     pip3 uninstall -y uv
 
+# Force-upgrade CVE-flagged packages that the resolver above leaves at
```
zhuangqh: you already pin the version in requirements.txt. why still run it here?

andyzhangx:
fixed by https://github.com/kaito-project/kaito/pull/2240

PR #2245, opened 2026-07-31
feat: support secrets redaction in streaming guardrails

xiaoqi-7:
**Reason for Change**:

Support safe streaming redaction for secrets using `redactMode: all`.

**Requirements**

**Issue Fixed**:

N/A

**Notes for Reviewers**:

Covers AWS keys, split chunks, multiple secrets, repeated scans, and leak prevention. `partial` and `hash` remain unsupported.


presets/ragengine/guardrails/scanner_schemas.py
```diff
@@ -58,12 +62,56 @@ class _SensitiveMatch:
...
+            secret_values,
+            key=lambda value: (-len(value), value),
+        ):
+            replacement = self._scanner.redact_value(
+                secret_value,
+                self._redact_mode,
+            )
+            sanitized_output = sanitized_output.replace(secret_value, replacement)
+
+        return sanitized_output, False, 1.0
```
Fei-Guo, line 89: Can you briefly explain what returned values mean?

presets/ragengine/guardrails/scanner_schemas.py
```diff
@@ -58,12 +62,56 @@ class _SensitiveMatch:
...
+                secrets.scan_file(temp_path)
+        finally:
+            os.remove(temp_path)
+
+        return {
+            found_secret.secret_value
+            for file_path in secrets.files
+            for found_secret in secrets[file_path]
+            if found_secret.secret_value
+        }
```
Fei-Guo, line 114: Also, what does this method _detect_secret_values do? why is a file involved?

presets/ragengine/streaming/guardrails.py
```diff
@@ -189,6 +212,28 @@ def scan(self, text: str) -> WindowScanResult:
...
+def _redact_secrets_and_verify(scanner: Any, prompt: str, text: str) -> str | None:
+    sanitized, results_valid, _ = scan_output([scanner], prompt, text, fail_fast=False)
+    if not isinstance(sanitized, str):
+        return None
+    if all(results_valid.values()):
+        return text if sanitized == text else None
+    if sanitized == text or len(sanitized) > len(text):
+        return None
+
+    verified, verified_valid, _ = scan_output(
```
Fei-Guo, line 224: Why do you need to verify the result? Is there a chance that previous step goes wrong?

xiaoqi-7: The first scan’s `results_valid` applies to the original text, not the sanitized output. We rescan to validate the postcondition that the sanitized output contains no secrets detectable by the scanner. This is a defense-in-depth check against incomplete redaction; if the output is modified again or remains invalid, we fail closed.

presets/ragengine/guardrails/scanner_schemas.py
xiaoqi-7: The returned tuple is (sanitized_output, is_valid, risk_score). Here, sanitized_output is the redacted text, False indicates that the original output contained at least one secret, and 1.0 is the fixed risk score returned by LLM Guard when a secret is detected; it is not a probability.

The third return value is risk_score. Although the scanner interface accepts a float, LLM Guard's Secrets scanner uses only two values: -1.0 when no secret is detected and 1.0 when at least one secret is detected. Our adapter preserves the same behavior.

presets/ragengine/guardrails/scanner_schemas.py
xiaoqi-7: _detect_secret_values collects the unique secret values detected in the output so that we can redact every occurrence. A temporary file is used because the SecretsCollection flow used here scans through scan_file(), which expects a file path; the file is deleted immediately after scanning.

Fei-Guo (approved):

Fei-Guo (approved):

PR #2260, opened 2026-08-06
feat: support word matching in streaming ban-substrings

xiaoqi-7:
## Reason for Change

PR #2254 added streaming `ban_substrings` support for `match_type: str`. This PR adds `match_type: word` support for both `block` and `redact`.

## What Changed

- Preserves the preceding character for left-boundary checks.
- Directly validates word boundaries across chunks.
- Handles split matches and end-of-response matches.
- Uses dynamic holdback based on the longest configured substring.
- Continues to reject `contains_all: true`.

## Testing

Added tests for word boundaries, cross-chunk matches, final flush, case sensitivity, long substrings, and both `block` and `redact` actions.


presets/ragengine/streaming/buffer_window.py
```diff
@@ -31,7 +31,13 @@ class WindowScanResult:
 
 
 class WindowScanner(Protocol):
-    def scan(self, text: str) -> WindowScanResult: ...
+    def scan(
+        self,
+        text: str,
+        *,
+        flush: bool = False,
+        left_context: str = "",
```
Fei-Guo: what does "left_context" mean? can you use a better name?

presets/ragengine/streaming/guardrails.py
```diff
@@ -237,16 +259,95 @@ def scan(self, text: str) -> WindowScanResult:
...
+            scanner_output = scanner_output[len(scan_prefix) :]
+        if boundary_guard:
+            if not scanner_output.endswith(boundary_guard):
+                return None
+            scanner_output = scanner_output[: -len(boundary_guard)]
+
+        return scanner_output, results_valid
+
+
+def _select_boundary_sentinel(
```
Fei-Guo: I think this is overengineered. 

You can just pick a unicode char that never appears in word as sentinel for simplicity.

presets/ragengine/streaming/guardrails.py
```diff
@@ -237,16 +259,95 @@ def scan(self, text: str) -> WindowScanResult:
...
+                return WindowScanResult(blocked=True)
+            _, results_valid = scan_result
             if not all(results_valid.values()):
                 return WindowScanResult(blocked=True)
 
         if sanitized_text == text:
             return WindowScanResult()
         return WindowScanResult(sanitized_text=sanitized_text)
 
+    def _scan_output(
```
Fei-Guo: The entire sentinel system is complicated. A few suggestions from AI:

A few simpler alternatives:

1. Don't use the library scanner for word matching. Do it yourself:


def check_word_match(text, substring, left_context):
    for match in re.finditer(re.escape(substring), text, flags):
        left_ok = (match.start() == 0 and not re.fullmatch(r"\w", left_context)) \
                  or (match.start() > 0 and not re.fullmatch(r"\w", text[match.start()-1]))
        right_ok = match.end() == len(text) and flush \
                   or (match.end() < len(text) and not re.fullmatch(r"\w", text[match.end()]))
        if left_ok and right_ok:
            # it's a word match
You skip the library's \b entirely and check boundaries yourself with direct character inspection. No sentinels needed. For redaction, do text.replace() on confirmed matches.

2. Keep holdback = len+1 and include left char in the buffer:

Never emit the very last character — always keep it as the leftmost char of the buffer. The regex then has real context on both sides. Costs 1 extra character of latency permanently, but zero complexity.

3. Two-pass: str match first, then verify boundaries:

Call the scanner with match_type: str. If no match, pass through. If match found, check left_context and text[match_end] manually to confirm it's a real word boundary.

All three avoid the sentinel system.

presets/ragengine/streaming/buffer_window.py
xiaoqi-7: `left_context` is the last emitted character before the current buffer, used to determine the left word boundary. I’ll rename it to `preceding_char` for clarity.

presets/ragengine/streaming/guardrails.py
xiaoqi-7: Agreed—the sentinel logic can be simplified. I’ll use the actual `preceding_char` for the left boundary and a fixed non-word sentinel for the synthetic right boundary, then remove `_select_boundary_sentinel()`.

presets/ragengine/streaming/guardrails.py
xiaoqi-7: Thanks for the suggestions. I agree the sentinel system is too complicated. I’ll use direct boundary checks and redact only confirmed match spans to preserve case-insensitive matching and avoid false replacements.

Fei-Guo (approved):

PR #2263, opened 2026-08-08
feat: surfaces granular model mirror and streaming errors and download metrics to status

### LINKED ISSUE #2258: Surface model mirror and streaming errors and metrics in status
Surface any missing errors and essential metrics (such as mirror download speed) to CR status.

**Additional context**

2170chm:
**Reason for Change**:

#### Problem

Model-mirror and streaming failures were invisible in CR status. A failed download left a ModelMirror with no Ready condition at all, and streaming misconfiguration surfaced only as an opaque reconcile error. There was no way for a separate troubleshoot controller to read status and emit events.

#### Download throughput
The download Job's shell script now brackets the download with timestamps and echoes a marker line, KAITO_DOWNLOAD_STATS seconds=N bytes=N — the same pattern as the existing KAITO_BENCHMARK_RESULT. A new downloadstats package reads the pod log and parses it into status.downloadDurationSeconds, downloadedBytes, and downloadThroughputMBps.

The parser is forgiving: a missing marker returns (nil, nil), not an error, so a successful download always reaches Ready even when stats are unavailable.

#### Failure reasons
ModelMirror failures now map to constants in consts.go: PVCPending, DownloadFailed, DownloadOOMKilled, DownloadEvicted, etc. Workspace streaming-validation failures flow through a typed streamingValidationError carrying a reason; streamingValidationReason lifts it onto the InferenceReady condition.

#### Bug discovered
checkJobStatus selected only non-failed Jobs, so a failed Job could never be selected and the entire failure branch was unreachable dead code. Now selects the newest Job regardless of state.

#### hf_transfer removal in download script
The download script previously installed hf_transfer and set HF_HUB_ENABLE_HF_TRANSFER=1. Both are gone; downloads use plain huggingface-hub with --max-workers 4.

Testing surfaced a FutureWarning in the pod logs showing the variable was deprecated and already inert. I also tried the recommended replacement, HF_XET_HIGH_PERFORMANCE, and measured no improvement, so I dropped both.

**Requirements**

**Issue Fixed**:

Closes #2258 
**Notes for Reviewers**:

#### Positive

P1 — happy path download, stats populated
P2 — small + large model, stats populated

#### Negative — ModelMirror

N1 — nonexistent model → DownloadFailed
N2 — gated model → DownloadFailed
N3 — bad StorageClass → PVCPending
N4 — memory limit too small → DownloadOOMKilled

#### Negative — Workspace streaming validation

N5 — missing StorageClass → ModelStreamingStorageClassNotFound
N6 — wrong CSI provisioner → ModelStreamingInvalidStorageClass
N7 — ServiceAccount absent → ModelStreamingServiceAccountInvalid
N8 — SA without workload identity → same reason, different message


api/v1alpha1/modelmirror_types.go
```diff
@@ -117,6 +117,17 @@ type ModelMirrorStatus struct {
 	Conditions       []metav1.Condition `json:"conditions,omitempty"`
 	FailureMessage   string             `json:"failureMessage,omitempty"`
 	LastDownloadTime *metav1.Time       `json:"lastDownloadTime,omitempty"`
+
+	// DownloadDurationSeconds is the wall-clock duration of the model download
```
zhuangqh: how about group them.

```go
type ModelMirrorDownloadStatus struct {
    // DurationSeconds is the wall-clock duration of the model download
    // phase only, excluding dependency installation and post-download cleanup.
    DurationSeconds int64 `json:"durationSeconds,omitempty"`

    // DownloadedBytes is the total size on disk of the downloaded model.
    DownloadedBytes int64 `json:"downloadedBytes,omitempty"`

    // ThroughputMBps is the average download throughput in MB/s,
    // derived from DownloadedBytes and DurationSeconds.
    ThroughputMBps string `json:"throughputMBps,omitempty"`
}

type ModelMirrorStatus struct {
    ...
    Download *ModelMirrorDownloadStatus `json:"download,omitempty"`
}
```

zhuangqh: fields with same prefix usually should be grouped.

api/v1alpha1/modelmirror_types.go
```diff
@@ -117,6 +117,17 @@ type ModelMirrorStatus struct {
 	Conditions       []metav1.Condition `json:"conditions,omitempty"`
 	FailureMessage   string             `json:"failureMessage,omitempty"`
 	LastDownloadTime *metav1.Time       `json:"lastDownloadTime,omitempty"`
```
andyzhangx, line 119: what does `LastDownloadTime` mean here? in fact, I would like to know download start time, and download end time

2170chm: Sure I can add the start and end time, which should be the last download's start and end time (currently it's only the end time). This is because a job can fail and a ModelMirror can spawn new download jobs until it succeeds, hence the "last".

pkg/modelmirror/controllers/modelmirror_controller.go
```diff
@@ -293,20 +326,106 @@ func (r *ModelMirrorReconciler) checkJobStatus(ctx context.Context, cr *kaitov1a
...
+func (r *ModelMirrorReconciler) handleJobSuccess(ctx context.Context, cr *kaitov1alpha1.ModelMirror, job *batchv1.Job, log logr.Logger) (ctrl.Result, error) {
 	modelID := cr.Spec.Source.ModelID
 	cr.Status.Phase = kaitov1alpha1.ModelMirrorPhaseReady
 	cr.Status.ModelPath = "/models/" + modelID
 	cr.Status.FailureMessage = ""
 	cr.Status.LastDownloadTime = ptr.To(metav1.Now())
 
-	setCondition(cr, mmconsts.ConditionTypeReady, metav1.ConditionTrue, "DownloadSucceeded", "Model download completed")
-	setCondition(cr, mmconsts.ConditionTypeStorageReady, metav1.ConditionTrue, "PVCBound", "PVC is bound")
+	r.recordDownloadStats(ctx, cr, job, log)
```
rambohe-ch: We should update download status at a cycle interval, such as 30s or 1min, not configure download status at the success end.

pkg/workspace/controllers/workspace_controller.go
rambohe-ch: Maybe we can use reason from ModelMirror status instead of a fixed reason string: ModelMirrorPending.

rambohe-ch:
@2170chm There are two options for productionstack-status-reporter to emit events for ModelMirror.
- option1: according to workspace.status.conditions["ModelMirrorReady"] to emit events
- option1: according to modelMioor.status.conditions to emit events

Which option do you prefer?

pkg/modelmirror/downloadstats/downloadstats.go
```diff
@@ -0,0 +1,145 @@
...
+	scanner.Buffer(make([]byte, 4096), maxScanTokenSize)
+	scanner.Split(scanLinesOrCR)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if idx := strings.Index(line, MarkerPrefix); idx != -1 {
+			lastMarker = strings.TrimSpace(line[idx+len(MarkerPrefix):])
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		if errors.Is(err, bufio.ErrTooLong) && lastMarker == "" {
```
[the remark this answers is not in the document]
2170chm: It is an exported Go API

2170chm:
> @2170chm There are two options for productionstack-status-reporter to emit events for ModelMirror.
> 
> * option1: according to workspace.status.conditions["ModelMirrorReady"] to emit events
> * option1: according to modelMioor.status.conditions to emit events
> 
> Which option do you prefer?

I think option 1 because events are namespaced but modelMirror is a cluster-scoped resource.

pkg/modelmirror/download/sampler.py
```diff
@@ -0,0 +1,264 @@
...
+    # operator may poll during it. Until the first sample lands the endpoint
+    # reports speed 0 / ETA -1, which is a valid "not yet known" answer.
+    serve(state)
+
+    state.set_total(fetch_total_bytes(model_id, exclude))
+
+    while True:
+        now = time.time()
+        state.update(directory_bytes(model_dir), now)
+        state.set_finished(is_finished(model_dir))
```
rambohe-ch, line 259: How about make these functions(like directory_bytes, is_finished) as common functions? so inference.py can also use them.

2170chm: I think we can leave them separated because the sampler is delivered as a single embedded script into a stock image while inference_api.py ships in the preset image, so there's no import path between them, if we do add a shared package it would not be worth it. The two also read different HF layouts (.cache/huggingface/download vs blobs/).

rambohe-ch (approved):

Fei-Guo (approved):

pkg/modelmirror/download/sampler.py
```diff
@@ -0,0 +1,264 @@
...
+REMAINING_METRIC = "model_mirror_download_remaining_seconds"
+
+DEFAULT_PORT = 9100
+SAMPLE_INTERVAL_SECONDS = 60
+
+
+def render_metrics(state):
+    """Prometheus text format. Values are emitted as-is; the operator truncates
+    toward zero when assigning to its int64 status fields."""
+    return (
```
zhuangqh, line 180: it's quite hacky here. is it because you dont want to introduce the Prometheus lib here?

pkg/modelmirror/controllers/modelmirror_controller.go
```diff
@@ -281,32 +311,149 @@ func (r *ModelMirrorReconciler) checkJobStatus(ctx context.Context, cr *kaitov1a
...
+		if pod.Status.Phase == corev1.PodFailed && pod.Status.Reason == "Evicted" {
+			return mmconsts.ReasonDownloadEvicted,
+				fmt.Sprintf("download pod was evicted: %s", pod.Status.Message)
+		}
+	}
+
+	return "", ""
+}
+
+func (r *ModelMirrorReconciler) handleJobSuccess(ctx context.Context, cr *kaitov1alpha1.ModelMirror, job *batchv1.Job, log logr.Logger) (ctrl.Result, error) {
```
zhuangqh, line 375: handleJobSuccess's job *batchv1.Job parameter is unused

zhuangqh (approved):

Issue #1641, opened 2025-11-05
RAGEngine status condition updates


pauldotyu:
**Describe the bug**

RAGEngine controller does not update `observedGeneration` in status conditions after reconciling spec changes, causing `kubectl wait --for=condition=ServiceReady` to timeout indefinitely even when the service is actually ready and all conditions show `Status: True`.

This makes it difficult to write robust deployment scripts and CI/CD pipelines that need to wait for RAGEngine resources to be fully reconciled after updates. The standard Kubernetes pattern of checking observedGeneration == generation is broken, as kubectl wait --for=condition=X expects the condition to be satisfied at the current generation.

**Steps To Reproduce**

1. Deploy a RAGEngine resource:

```bash
kubectl apply -f - <<EOF
apiVersion: kaito.sh/v1alpha1
kind: RAGEngine
metadata:
  name: rag-gpt-oss-20b
spec:
  compute:
    count: 1
    instanceType: Standard_D8ds_v6
    labelSelector:
      matchLabels:
        apps: rag-gpt-oss-20b
  embedding:
    local:
      modelID: BAAI/bge-small-en-v1.5
  inferenceService:
    contextWindowSize: 131072
    url: http://my-llm-service/v1/chat/completions
EOF
```

2. Wait for initial deployment to complete successfully (generation: 1, observedGeneration: 1):

```bash
kubectl wait --for=condition=ServiceReady ragengine/rag-gpt-oss-20b --timeout=10m
```

3. Update the RAGEngine spec (e.g., change the inference service URL):

```bash
kubectl patch ragengine rag-gpt-oss-20b --type=merge -p '{"spec":{"inferenceService":{"url":"http://updated-llm-service/v1/chat/completions"}}}'
```

4. Attempt to wait for the ServiceReady condition after the update:

```bash
kubectl wait --for=condition=ServiceReady ragengine/rag-gpt-oss-20b --timeout=10m
```

5. The command times out after 10 minutes, even though kubectl describe shows the resource is ready
 
The RAGEngine resource has `generation: X` (indicating X modifications to the spec) but all status conditions still show `observedGeneration: Y`, suggesting the controller successfully reconciles the spec changes but fails to update the observedGeneration field in the status subresource.

**Expected behavior**

After the controller reconciles the spec changes, it should update the observedGeneration field in all status conditions to match the current metadata.generation value. The kubectl wait command should succeed once the condition is satisfied at the current generation.

**Logs**

```text
[...]

PR #2270, opened 2026-08-11
fix: refresh observedGeneration for InferenceSet and Workspace status

andyzhangx:
## Reason for Change

This follows up on the stale `observedGeneration` issue discussed in #1641 / PR #1708.

The latest `main` branch still had two remaining gaps:

1. `pkg/utils/inferenceset/inferenceset.go` skipped status updates when `status/reason/message` matched, even if `ObservedGeneration` was stale.
2. Workspace benchmark status was write-once, but a previously recorded `BenchmarkCompleted=True` condition could keep an older `ObservedGeneration` across later spec generations.

## What Changed

- Update `InferenceSet` condition matching to also require `ObservedGeneration == metadata.generation` before skipping a status write.
- Refresh `Workspace` `BenchmarkCompleted` observedGeneration when a completed benchmark result is being preserved.
- Add regression tests for both cases.

## Why a dedicated `refreshBenchmarkObservedGenerationIfCompleted` helper?

`setWorkspaceCondition` / `meta.SetStatusCondition` already refreshes `ObservedGeneration` whenever we go through it. The problem is that the `BenchmarkCompleted` condition is intentionally **write-once**: once it flips to `True`, `applyBenchmarkStatus` short-circuits and returns early **without** calling `setWorkspaceCondition` again:

```go
// pre-PR
if c := meta.FindStatusCondition(status.Conditions, ...BenchmarkCompleted); c != nil && c.Status == metav1.ConditionTrue {
    return nil    // ← never touches setcondition after this
}
```

This is by design — we don't want to overwrite the recorded benchmark result or re-read logs on every reconcile. But the side effect is that `BenchmarkCompleted.ObservedGeneration` stays pinned to whatever `metadata.generation` was when the benchmark first completed. Every subsequent spec update (e.g. changing `resource.count`, adjusting adapters, anything that doesn't reset the benchmark) bumps `metadata.generation` but leaves `BenchmarkCompleted.ObservedGeneration` behind — that's exactly the stale-generation symptom #1641 / #1708 was tracking.

The naive fix — "just call `setWorkspaceCondition` in the short-circuit path too" — has two problems:

1. `setWorkspaceCondition` writes `Status/Reason/Message` too. Even if we pass the same values, it churns `LastTransitionTime` semantics and forces us to reconstruct the original `Reason`/`Message` at every call site.
2. The short-circuit lives in two places (`applyBenchmarkStatus` **and** an early guard at the top of `applyInferenceWorkspaceStatus` so that even InferenceStatus-only reconciles refresh it). Duplicating the "find condition + rebuild + set" logic in both is easy to get wrong.

So `refreshBenchmarkObservedGenerationIfCompleted` is deliberately narrow:

- only touches `ObservedGeneration`, preserves `Status`/`Reason`/`Message`/`LastTransitionTime`
- returns `bool` so `applyBenchmarkStatus` can reuse it as the same short-circuit predicate (`if refresh...IfCompleted { return nil }`) — one call, no duplicated `FindStatusCondition + Status==True` check
- single source of truth for both call sites

## Validation

- `git diff --check`
- Added/updated unit tests in:
  - `pkg/utils/inferenceset/inferenceset_test.go`
  - `pkg/workspace/controllers/workspace_controller_test.go`
- [blocked] Local `go test` / `gofmt` could not be run in this environment because `go` and `gofmt` are unavailable.

## Notes for Reviewers

- This PR intentionally does **not** touch `RAGEngine` yet.
- Workspace change is scoped to preserving write-once benchmark results while still advancing `ObservedGeneration`.

### DIFF
--- a/pkg/workspace/controllers/workspace_controller.go
+++ b/pkg/workspace/controllers/workspace_controller.go
@@ -1253,6 +1253,15 @@ func applyBenchmarkStatus(ctx context.Context, status *kaitov1beta1.WorkspaceSta
 	return nil
 }
 
+// refreshBenchmarkObservedGenerationIfCompleted preserves the write-once
+// BenchmarkCompleted=True state and recorded benchmark result, while advancing
+// the condition's ObservedGeneration to the current Workspace generation.
+//
+// This keeps the condition from looking stale after later reconciles or spec
+// updates that do not require re-running benchmark. Callers can use the return
+// value as a short-circuit signal: true means benchmark is already completed
+// (and its ObservedGeneration is now current), so no log re-read or result
+// overwrite should happen.
 func refreshBenchmarkObservedGenerationIfCompleted(status *kaitov1beta1.WorkspaceStatus, generation int64) bool {
 	c := meta.FindStatusCondition(status.Conditions, string(kaitov1beta1.WorkspaceConditionTypeBenchmarkCompleted))
 	if c == nil || c.Status != metav1.ConditionTrue {
--- a/pkg/workspace/controllers/workspace_controller.go
+++ b/pkg/workspace/controllers/workspace_controller.go
@@ -1185,6 +1185,8 @@ func applyInferenceWorkspaceStatus(ctx context.Context, status *kaitov1beta1.Wor
 	resourceReady := resourceConditionStatus == metav1.ConditionTrue
 	isInferenceEstablished := status.State == kaitov1beta1.WorkspaceStateReady || status.State == kaitov1beta1.WorkspaceStateNotReady
 
+	refreshBenchmarkObservedGenerationIfCompleted(status, generation)
+
 	if inferenceReady && resourceReady {
 		setWorkspaceCondition(status, generation, appendMessage,
 			kaitov1beta1.WorkspaceConditionTypeInferenceStatus, metav1.ConditionTrue, "WorkspaceInferenceStatusSuccess", "Inference has been deployed successfully")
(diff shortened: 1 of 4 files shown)


Fei-Guo (approved):

pkg/workspace/controllers/workspace_controller.go
```diff
@@ -1200,6 +1200,8 @@ func applyInferenceWorkspaceStatus(ctx context.Context, status *kaitov1beta1.Wor
 	resourceReady := resourceConditionStatus == metav1.ConditionTrue
 	isInferenceEstablished := status.State == kaitov1beta1.WorkspaceStateReady || status.State == kaitov1beta1.WorkspaceStateNotReady
 
+	refreshBenchmarkObservedGenerationIfCompleted(status, generation)
```
andyzhangx, line 1188: > The `refreshBenchmarkObservedGenerationIfCompleted` call at L1203 is there to cover the code paths where `applyBenchmarkStatus` is never reached.
>
> `applyInferenceWorkspaceStatus` has multiple branches and early returns (e.g. not-ready / error paths) that finish the reconcile without ever invoking the benchmark logic. On those paths, the refresh inside `applyBenchmarkStatus` doesn't fire, so a pre-existing `BenchmarkCompleted=True` condition would keep a stale `ObservedGeneration` even though the Workspace generation has advanced.
>
> Placing the refresh at the entry of `applyInferenceWorkspaceStatus` guarantees that — regardless of which branch the reconcile takes — the `BenchmarkCompleted` condition's `ObservedGeneration` is always brought up to the current generation, while the write-once status/result is preserved. The regression test `"not-ready path preserves benchmark condition and refreshes observedGeneration"` exercises exactly this path.

Fei-Guo (approved):

PR #2272, opened 2026-08-11
chore: nodeclass API cleanup

2170chm:
**Reason for Change**:

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:


pkg/utils/consts/consts.go
```diff
@@ -111,6 +111,10 @@ const (
 	AKSNodeClassAzureLinuxName = "image-family-azure-linux"
 	AKSNodeClassOSDiskSizeGB   = 300
 
+	// NodeClassConfigMapName is the ConfigMap in the release namespace holding the
+	// admin-defined NodeClass manifests KAITO creates and allows Workspaces to select.
+	NodeClassConfigMapName = "kaito-nodeclasses"
```
zhuangqh: dont quite understand why use a separate cm here? do you expect this list will be changed during operator lifecycle. if not, why dont read them from operator cli args instead.

pkg/utils/consts/consts.go
```diff
@@ -97,6 +98,22 @@ func IsKarpenterProvisioner() bool {
...
 
+// allowedNodeClassNames is the sorted set of NodeClass names declared via
+// --karpenter-node-classes. Set once during startup in main.go; read by the Workspace
+// admission webhook to validate the node-class-name annotation. Unexported so callers
+// cannot mutate the allowlist in place.
+var allowedNodeClassNames []string
+
+// SetAllowedNodeClassNames publishes the NodeClass allowlist. Startup only.
+func SetAllowedNodeClassNames(names []string) {
+	allowedNodeClassNames = slices.Clone(names)
```
zhuangqh, line 109: why deepcopy? I think it's unnecessary

zhuangqh:
### Nit: `allowedNames` re-derivation in `Start()`

```go
allowedNames := make([]string, 0, len(p.nodeClassConfig.NodeClasses))
for _, nc := range p.nodeClassConfig.NodeClasses {
    allowedNames = append(allowedNames, nc.Name)
    ...
}
...
consts.SetAllowedNodeClassNames(allowedNames)
```

`ParseNodeClasses` (`pkg/nodeprovision/manager/factory.go`) already walks the same entries once to validate names (non-empty, DNS1123, no dupes) and returns the fully-validated `[]NodeClassSpec`. `Start()` re-projects that slice down to `[]string` in its own pass just to hand names to `consts.SetAllowedNodeClassNames`.

Not a bug — it's folded into the existing loop that also derives `DefaultName` and creates each NodeClass object, so it's not a second traversal, just a few bytes of incidental output from a loop that's already there. But since `ParseNodeClasses` already has the validated name list in hand, it could return `(names []string, entries []NodeClassSpec, err error)` (or just let `Start()` map over the already-clean slice) and skip re-building `allowedNames` from scratch. Purely cosmetic — feel free to ignore if you'd rather keep `Start()` self-contained.

zhuangqh (approved):

PR #2301, opened 2026-08-24
feat: [BREAKING CHANGE] refactored modelmirror and mirror sharing logic to be namespaced

### LINKED ISSUE #2287: Switch ModelMirror to be namespace-scoped
Currently ModelMirror is cluster-scoped but indirectly owns the namespace-scoped resources such as download job pod and PVC. It is more maintainable and cleaner to make it namespace-scoped.
Switch modelmirror to namespace-scoped resource. Workspaces of a model should only share a mirror within a namespace.

**Additional context**

2170chm:
**Reason for Change**:

Currently ModelMirror is cluster-scoped but indirectly owns namespace-scoped resources such as the download Job pod and the PVC. It is more maintainable and cleaner to make it namespace-scoped. This PR switches ModelMirror to a namespace-scoped resource. Workspaces for a model now share a mirror only within the same namespace.

**Summary**

- `ModelMirror` is now **namespace-scoped**. Its PVC and download Job are created in `metadata.namespace`; `spec.jobNamespace` is removed.
- `spec.source.accessSecret` (an `ObjectReference`) becomes `spec.source.accessSecretName` (a string), resolved in the mirror's own namespace.
- A Workspace resolves its mirror at `{ws.Namespace, sha256(modelID)[:6]}`. The name hash is unchanged, so the same model in two namespaces yields two mirrors with the same name and independent storage.

**Lifecycle**

Because mirrors of the same name now coexist across namespaces and can be recreated in place, teardown ordering matters in a way it did not before:

- A managed mirror carries `kaito.sh/model-mirror-cleanup`. `finalizeMirror` deletes the PVC and download Jobs and only releases the finalizer once they are **gone** — not merely marked for deletion. Static mirrors provision nothing and are deliberately left unfinalized.
- That ordering is what lets the mirror identify its children by name and label alone: a live mirror can never share a name with another generation. The PVC and Job still carry a controller ownerReference, which drives the `Owns()` watches and garbage collection.
- Finalization and the download-Job existence check read through the **API reader** rather than the informer cache. Releasing the finalizer on a stale read would drop the mirror while children survive, and a stale read before `Create` would start a second `hf download` against the same `ReadWriteMany` PVC — whose cleanup step (`rm -rf .cache`, recursive directory delete) would run against files the other writer still holds.
- `ensureFinalizer` patches with an optimistic lock so a stale read cannot clobber `foregroundDeletion`.

**Explicitly out of scope**

Hand-editing a mirror's PVC or planting objects that match its name or `kaito.sh/model-mirror-name` label is treated as user error and is not guarded. Deleting the PVC destroys the downloaded weights (reclaim policy is `Delete`) and the mirror keeps reporting `Ready` against storage that is gone; recovery is deleting the ModelMirror so it rebuilds. A Ready-storage re-check, a PVC finalizer, and ownerReference-UID verification on read were all prototyped and dropped as more machinery than the failure mode warrants.

**Side changes**

- Reusing an existing mirror now requires agreement on **Mode** and **storageClassName** as well as model ID; a mismatch is reported instead of silently adopted. These matter to the consumer because the bound PV's `volumeHandle` supplies both the `az://` container and the `AZURE_STORAGE_ACCOUNT_NAME` injected into the serving pod.
- An `AlreadyExists` on create now requeues instead of returning success, so the workspace still validates a mirror another workspace built between its cached read and its create.
- A workspace whose mirror is draining reports `ModelMirrorDeleting` and retries on a flat interval; returning it as a reconcile error would feed the exponential backoff and idle the workspace for minutes after a teardown that took seconds.
- `ModelMirrorReady` no longer reports `True` for a mirror the workspace was refused the use of, or for one that is terminating. It now agrees with `InferenceReady`.
- `progress.Fetch` reads the sampler through a bounded stream instead of buffering the whole response.
- The sampler sidecar builds its two gauges with `prometheus_client` instead of hand-formatting Prometheus text; the payload is unchanged apart from the speed value no longer being truncated to six significant figures. (Addresses a legacy issue from https://github.com/kaito-project/kaito/pull/2263)

**Upgrade**

`spec.scope` is immutable on an established CRD, so `helm upgrade` alone cannot perform this change. With the **old controller still running**, delete the CRD and let it drain:

```bash
kubectl delete crd modelmirrors.kaito.sh --timeout=300s
# the old handleDeletion deletes each mirror's Jobs and PVC and releases the CR finalizer;
# the CRD's customresourcecleanup finalizer blocks until every CR is gone
helm upgrade ...
```

**This destroys the mirrored weights, by design** — reclaim policy is `Delete`, so each blob container goes with its PVC and every namespace re-downloads once. There is no migration that carries weights across the scope change. Acceptable because the ModelMirror feature gate is off by default, so affected clusters have the CRD but no `ModelMirror` objects.

If the delete hangs, the old controller is not reconciling (scaled to 0, or the gate is off) so nothing removes `kaito.sh/model-mirror-cleanup`. Start it, or clear the CR finalizers by hand — but note that route strands the PVCs, since the old scheme gave them no ownerReference and they keep a `kaito.sh/model-mirror-protection` finalizer the new controller does not know.

**Requirements**

**Issue Fixed**:

Closes #2287

**Notes for Reviewers**:

**Manual E2E — namespace-scoped ModelMirror** (guide: `dev/docs/namespaced-model-mirror-e2e.md`)

Prerequisites — CRD served `Namespaced`; spec reduced to `mode / serviceAccountName / source / storage`; controller SA has PVC `create`/`delete`.

**Placement**
- A1 — PVC + Job land only in the CR's namespace; CR carries `kaito.sh/model-mirror-cleanup` and the PVC's ownerRef points at the CR
- A2 — standalone mirror (no Workspace) reaches Ready in a non-default namespace
- A3 — same model in 2 namespaces → 2 mirrors, same name, distinct PVs, both independently Ready

**Ownership / lifecycle**
- L1 — delete mirror → Jobs + PVC reaped, then the finalizer released; PV reclaimed (2s)
- L2 — `--cascade=foreground` completes without deadlocking (2s)
- L3 — namespace containing a live mirror + PVC + running Job deletes (18s)
- L5 — mirror deleted under a live Workspace; workspace gates, then recreates it
- L7 — exactly one download Job for the life of a download

**Regressions — real completions served**
- B1 — WI-blob streaming: `--model=az:///… --load-format=runai_streamer`
- B2 — Workspace in ns B builds its **own** mirror + PV; ns A's untouched
- S1 — SAS/static: mirror created **with** a namespace, `/mnt/streaming`, 0 `empty namespace` errors
- D1 — opt-out: no mirror created, `--download-dir=/workspace/weights`, 0 `az://`

**Metrics** (covers `DoRaw` → `Stream` + `ParseLimited`)
- E1 — live progression on `Qwen/Qwen3-32B`: `-1/0` → `{241, 216756000}` → `{176,…}` → `{108,…}` → `{43,…}`
- E2 — zeroes on success, `lastDownloadTime` stamped, 0 parse errors

**Negative paths**
- N4 — Mode disagreement → rejected; `ModelMirrorReady` / `InferenceReady` / `ResourceReady` all agree, and no nodes are provisioned
- N5 — StorageClass disagreement → rejected, names both classes and the annotation
- N7 — workspace gates on a draining mirror with `ModelMirrorDeleting` and recreates it 1s after teardown completes

**Health** — 0 controller restarts, 0 panics, 0 namespaces stuck Terminating, 0 mirrors stuck holding their finalizer, 0 orphaned PVCs.

Two notes on the guide itself: N2's `PVCPending` branch is no longer reachable from a bogus StorageClass name (admission rejects it first), and the per-namespace federated identity credential is required for any namespace that must complete a download — placement/teardown cases (L1–L3) do not need it.


Fei-Guo (approved):

zhuangqh (commented):
1. ensurePVC adopts PVCs without ownership check — compares only StorageClass, unlike the UID-based isOwnedBy() used for Jobs, allowing adoption of a terminating PVC from a deleted-and-recreated CR of the same name.
  2. releasePVC ownership check omits UID — inconsistent with isOwnedBy(), risking premature finalizer removal on a PVC actually owned by a newer CR generation.
  3. Misleading/dead modelmirrors/finalizers RBAC grant — comment claims it's needed for blockOwnerDeletion, but no finalizer is ever placed on the CR anymore.

Fei-Guo (approved):

zhuangqh (approved):

Fei-Guo (approved):

PR #2302, opened 2026-08-26
fix: optimize node estimation on common Azure SKUs

zhehli688:
**Reason for Change**:

Today KAITO's node estimator under- and over-provisioned nodes on common Azure GPU SKUs (e.g. estimates two Standard_NC48ads_A100_v4 nodes for gpt-oss-120b while it fits on one, estimates one Standard_NV36ads_A10_v5 node for NVIDIA-Nemotron-Nano-9B-v2 while it needs two due to Mamba cache overhead). Two root causes:

1. **`gpu-memory-utilization` is too conservative on A100/H100.** The estimator and vLLM launcher hardcoded `0.84` while vLLM's default is `0.92`.
2. **Hybrid Mamba/Attention models were under-sized.** NemotronH models allocate a per-sequence Mamba-2 state cache *in addition* to the attention KV cache, which the estimator ignored — so a model like `Nemotron-Nano-9B-v2` was estimated at 1×A10 but OOMs at KV-cache allocation.

Key changes:
- Bumped gpu-memory-utilization (https://docs.vllm.ai/en/latest/configuration/engine_args/#-gpu-memory-utilization) from 0.84 to 0.92 while keeping the 0.82 override for A10.
- Stored Mamba cache size in model_catalog.yaml and take it into account in node estimation.
- Overrided base overhead to 1.5 GiB (default 2.4 GiB) for A10.
- added real-catalog node estimation unit tests for A100 (80 GiB), H100 (94 GiB), and A10 (24 GiB) SKUs.
- refreshed e2e tests to use latest catalog models (`Qwen3.8-27B`, `granite-4.1-8b`, `Nemotron-3-Nano-4B-BF16`); consolidated adapter tests into `inference_with_adapters_test.go`.

**Requirements**

- [x] manually tested the following models:
    - DeepSeek-V4-Flash-0731 on one 2xH100
    - gpt-oss-120b on one 2xA100
    - Llama-3.3-70B-Instruct on one 2xA100
    - Mistral-medium-3.5 on one 2xA100
    - granite-4.1-8b on one 1xA10
    - NVIDIA-Nemotron-Nano-9B-v2 on one 2xA10
    -  Qwen3.8-27B on one 1xA100

**Issue Fixed**:

**Notes for Reviewers**:


zhuangqh (approved):
AWESOME change! fitting in one node has much better cost and effiency.
And great e2e test cases compaction.

Fei-Guo (approved):

PR #2306, opened 2026-08-28
fix: report runtime max model length in status

### LINKED ISSUE #2257: Report actual max-model-len in Workspace status
The actual max-model-len of a serving model is determined by vLLM based on the results of memory profiling and can be different than the model's advertised max-model-len. When a inference request with `prompt_tokens + max_tokens > max_model_len` is submitted to vLLM, vLLM will return a 400 error with error messages like:
* **max_tokens > max_model_len**: max_tokens=300 cannot be greater than max_model_len=max_total_tokens=256
* **prompt_tokens + max_tokens > max_model_len**: This model's maximum context length is 256 tokens. However, you requested 16 output tokens and your prompt contains at least 241 input tokens, for a total of at least 257 tokens…

Reporting the actual max-model-len in Workspace status can help users better understand this limit.

Report actual max-model-len in Workspace/InferenceSet status, which is available via the /v1/models endpoint.

**Additional context**

mikemikimike:
**Reason for Change**:

vLLM can reduce the usable context window during memory profiling, so the configured or model-advertised limit may not match the serving limit. Exposing the runtime value in Workspace and InferenceSet status lets users diagnose request-limit errors.

**Requirements**

**Issue Fixed**:

Fixes #2257
https://github.com/kaito-project/kaito/issues/2257

**Notes for Reviewers**:

The controller reads the OpenAI-compatible `/v1/models` metadata after the inference workload is ready and records the reported `max_model_len` in `status.maxModelLen`. Missing or temporarily unavailable metadata is treated as unknown and does not make reconciliation fail. InferenceSet status reports the value only when all ready replicas agree.

Validation: `go test ./pkg/workspace/controllers ./pkg/inferenceset`, `go vet ./...`, `golangci-lint run --new-from-rev=HEAD ./api/... ./pkg/workspace/controllers/... ./pkg/inferenceset/...`, and `git diff --check`.

Known validation limits: the full repository suite still has unrelated existing failures in tuning-config directory fixtures and Windows shell-script newline assertions; the E2E suites require an external registry and cluster, and race tests require a local C compiler. The changed packages pass their focused tests.


zhuangqh (commented):
Nice change. please update v1beta1 only since v1alpha1 is depracated. should not have new fields.

zhuangqh:
@mikemikimike  could you follow up this PR? discard v1alpha1 change and fix ut. Thanks

PR #2327, opened 2026-09-07
fix: default Karpenter capacity to on-demand and add spot opt-in

zhuangqh:
**Reason for Change**:

New Karpenter NodePools previously inherited the provider default capacity type, which could place inference workloads on Spot nodes unexpectedly. Every generated NodePool now carries an explicit karpenter.sh/capacity-type requirement that defaults to on-demand.

Spot is opt-in via the kaito.sh/capacity-type annotation on a Workspace, InferenceSet pod template, or MultiRoleInference. Validation webhooks reject unsupported values and treat the annotation as immutable; existing NodePools are left untouched.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

image

### DIFF
--- a/pkg/nodeprovision/karpenter/nodepool.go
+++ b/pkg/nodeprovision/karpenter/nodepool.go
@@ -74,16 +74,38 @@ func isInferenceSetWorkspace(ws *kaitov1beta1.Workspace) bool {
 	return ok
 }
 
+func resolveCapacityType(ws *kaitov1beta1.Workspace) (string, error) {
+	capacityType := ws.Annotations[kaitov1beta1.AnnotationCapacityType]
+	if capacityType == "" {
+		return consts.KarpenterCapacityTypeOnDemand, nil
+	}
+	switch capacityType {
+	case consts.KarpenterCapacityTypeOnDemand, consts.KarpenterCapacityTypeSpot:
+		return capacityType, nil
+	default:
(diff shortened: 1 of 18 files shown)


pkg/nodeprovision/karpenter/nodepool.go
```diff
@@ -74,17 +74,39 @@ func isInferenceSetWorkspace(ws *kaitov1beta1.Workspace) bool {
...
+func resolveCapacityType(ws *kaitov1beta1.Workspace) (string, error) {
+	capacityType := ws.Annotations[kaitov1beta1.AnnotationCapacityType]
+	if capacityType == "" {
+		return consts.KarpenterCapacityTypeOnDemand, nil
+	}
+	switch capacityType {
+	case consts.KarpenterCapacityTypeOnDemand, consts.KarpenterCapacityTypeSpot:
+		return capacityType, nil
+	default:
+		return "", fmt.Errorf("annotation %s=%q is invalid: supported values are %q and %q",
```
rambohe-ch, line 86: How about set to `KarpenterCapacityTypeOnDemand` for unknown capacity type? because return error can not help resolve this unknown capacity setting.

zhuangqh: We should explicitly raise an error if it's wrong instead of fallback silently I think.
The error msg tell what options are available.

Fei-Guo (approved):

PR #2328, opened 2026-09-08
fix: prefetch SAS model assets before Ray workers start

zhuangqh:
**Reason for Change**:

Run:ai downloads non-weight model files only where vLLM constructs its model configuration. It then distributes the resolved node-local path to remote workers without populating their filesystems. Multi-node models that initialize tokenizers, multimodal processors, or custom code on workers therefore fail at startup, even when requests are text-only. See https://github.com/vllm-project/vllm/issues/50616.

KAITO runs inference_api.py only on the leader, so downloading there would leave the same gap. Instead, extend the SAS-fetch init container, which runs in every pod before the Ray head or workers start, to download the complete non-weight asset set under the resolved model prefix.

Reuse the existing pod-local shared volume and populate vLLM's model_streamer cache using the SHA256-derived directory for the exact remote model URI. Export VLLM_ASSETS_CACHE and disable cache cleanup through the existing SAS env-file wrapper. Every worker consequently starts with the assets at the path it receives from the driver. Keep the remote model URI unchanged and exclude tensor files so Run:ai weight streaming retains its startup benefits.

Reuse the paginated blob listing, preserve SAS query encoding, reject unsafe destination paths, and publish each file only after its download completes. Failed or truncated downloads stop initialization; retries re-download files instead of trusting a partial cache. Add regression coverage for filtering, cache paths, pagination, failed downloads and retries, and the actual entrypoint env handoff.

This is a temporary SAS-path workaround, not a vLLM modification or a shared-PVC requirement. Separate remote tokenizers and draft models are out of scope. Leave a TODO to remove the prefetch and cache exports once KAITO ships the upstream fix.

Validation: .venv/bin/python pkg/workspace/inference/modelstreaming/azure/fetch_sas_test.py (16 existing helper tests and 11 prefetch tests passed); targeted SAS provider and pod-wiring Go tests passed; ruff check and format checks passed. A live multi-node deployment was not exercised.

AI assistance was used to implement this change.

**Requirements**

**Issue Fixed**:

**Notes for Reviewers**:

### DIFF
--- a/pkg/workspace/inference/modelstreaming/azure/fetch_sas.py
+++ b/pkg/workspace/inference/modelstreaming/azure/fetch_sas.py
@@ -21,8 +21,9 @@ script resolves everything else at pod runtime using the workload identity:
   3. Derive the storage account and container from the blobUri.
   4. Mint a SAS at the mint endpoint with {blobUri[, assetId]} -> SAS token.
   5. List the container with the SAS to discover the safetensors subpath -> model streaming URI.
-  6. Write AZURE_STORAGE_SAS_TOKEN, AZURE_STORAGE_ACCOUNT_NAME, and STREAM_MODEL_URI to the
-     shared env file so the main container's entrypoint wrapper can source them.
+  6. Prefetch non-weight model files into the shared volume's vLLM asset cache.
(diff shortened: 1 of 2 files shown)


zhuangqh:
should support WI model streaming. not sas only. closed.

PR #2329, opened 2026-09-08
fix: prefetch object-storage model assets on every Ray node

### LINKED ISSUE #2320: Runai_streamer metadata is missing on peer nodes
**Describe the bug**

A multi-node Workspace using `runai_streamer` fails because vLLM downloads non-tensor model files only on the Ray head. Worker actors receive the same node-local path, but that directory is empty on peer nodes.

**Steps To Reproduce**

1. Deploy an `az://` model across two one-GPU nodes.
2. Use Ray with PP=2, TP=1, and `runai_streamer`.
3. Observe the worker fail while reading the generated cache path.

**Expected behavior**

Required model metadata should be materialized on every node while weights remain streamed.

**Logs**

```text
OSError: Repo id must be in the form 'repo_name' or 'namespace/repo_name':
'/root/.cache/vllm/assets/model_streamer/8d53e465'
```

**Environment**

- Kubernetes version: Client `v1.35.0`, Server `v1.35.7`
- OS: Ubuntu 24.04.4 LTS
- Install tools: KAITO on AKS
- Others: 2× H100 nodes, vLLM `0.25.1`, Ray PP=2/TP=1

**Additional context**

Ray cluster formation and actor placement succeed. This is an upstream vLLM/Run:ai issue:

- [vLLM issue #50616](https://github.com/vllm-project/vllm/issues/50616)
- [vLLM PR #50633](https://github.com/vllm-project/vllm/pull/50633)

zhuangqh:
**Reason for Change**:

vLLM pulls a remote model's non-weight files (config.json, tokenizer,
processor, trust_remote_code modules) only where ModelConfig is built, then
hands the resolved node-local path to remote workers without populating their
filesystems. Multi-node models that initialize tokenizers, multimodal
processors, or custom code on workers fail at startup even for text-only
requests. See https://github.com/vllm-project/vllm/issues/50616.

KAITO runs inference_api.py on the Ray leader only, so the leader's own pull
leaves that gap. Before starting the server, fan out a short Ray task to every
alive node that calls vLLM's own ObjectStorageModel.pull_files() for each
remote URI the driver will resolve: the model, a separate --tokenizer, and a
speculative draft model. The cache directory is derived from the URI alone, so
each node reproduces the exact path the driver ships to its workers. Reusing
vLLM's helper keeps the hash and layout from drifting. Tensor files stay
excluded so Run:ai weight streaming keeps its startup benefit.

Pin one absolute VLLM_ASSETS_CACHE for the driver and the tasks, keep the
driver's runtime_env without copying the leader's short-lived storage
credentials, bound the barrier with --kaito-model-asset-prefetch-timeout, and
cancel stragglers on failure so a restart re-downloads instead of racing a
half-written cache.

Because the fix is a temporary upstream workaround, isolate it in a dedicated
presets/workspace/inference/vllm/_workarounds/model_asset_prefetch.py module
(CLI flag registered via register_args) so inference_api.py keeps only a
three-line seam and the whole thing deletes cleanly once the upstream fix
ships. The tfs Dockerfile copies the package into the image.

Chain the leader's Ray startup and the API server with '&&' instead of ';'.
multi-node-serving.sh leader exits non-zero when the cluster does not reach
ray_cluster_size, and the upstream example chains the same way; with ';' the
server started against an incomplete cluster and the prefetch would only reach
the nodes that happened to have joined.

**Requirements**

**Issue Fixed**:

Closes #2320 
**Notes for Reviewers**:

Tested in a live cluster. Qwen/Qwen3.5-122B-A10B-GPTQ-Int4  + 2 A100 + modelstreamer

### DIFF
--- a/pkg/model/interface.go
+++ b/pkg/model/interface.go
@@ -584,8 +584,8 @@ func (p *PresetParam) buildMultiNodeRayCommand(rc RuntimeContext) []string {
 	rayLeaderCommand := utils.BuildCmdStr(p.VLLM.RayLeaderBaseCommand, p.VLLM.RayLeaderParams)
 	modelRunCommand := utils.BuildCmdStr(p.VLLM.BaseCommand, p.VLLM.ModelRunParams)
 	result := utils.BuildIfElseCmdStr(
-		`[ "${POD_INDEX}" = "0" ]`,                                      // leader if pod index is 0, otherwise worker
-		strings.Join([]string{rayLeaderCommand, modelRunCommand}, "; "), // leader: start ray head + model
+		`[ "${POD_INDEX}" = "0" ]`,                                        // leader if pod index is 0, otherwise worker
+		strings.Join([]string{rayLeaderCommand, modelRunCommand}, " && "), // start model only after Ray is ready
 		map[string]string{},
 		p.VLLM.RayWorkerBaseCommand, // worker: join the cluster
 		p.VLLM.RayWorkerParams,
--- a/pkg/model/interface_test.go
+++ b/pkg/model/interface_test.go
@@ -14,6 +14,7 @@
 package model
 
 import (
+	"os/exec"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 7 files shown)


zhehli688:
Nice fix!

zhehli688 (approved):

2170chm (approved):
lgtm

Fei-Guo (approved):

Issue #2325, opened 2026-09-07
Inference cold start: measured breakdown and proposed optimizations


zhuangqh:
## Summary

A single-GPU vLLM Workspace took **966 seconds (16m06s)** from creation to `WorkspaceSucceeded=True`.

| Item | Value |
|---|---|
| Model | `Qwen/Qwen3.8-27B` (51.75 GiB, 18 safetensors shards) |
| Instance | `Standard_NC24ads_A100_v4` (A100 80GB PCIe) |
| Image | `mcr.microsoft.com/aks/kaito/kaito-base:0.5.0` (5.60 GB) |
| vLLM | 0.25.1 |
| Runs | 1 |

## Measured time cost

| Phase | Time | Share | Measurement |
|---|---:|---:|---|
| GPU node provisioning | 284 s | 29.4% | Workspace created → Node object created |
| GPU node initialization | 126 s | 13.0% | Node created → `ResourceReady=True` |
| Inference resource preparation | 7 s | 0.7% | `ResourceReady=True` → image pull start |
| Container image pull | 101 s | 10.5% | kubelet `Pulling` → `Pulled` |
| Container creation and start | 3 s | 0.3% | image pulled → container start |
| Framework bootstrap | 74 s | 7.7% | container start → vLLM engine init |
| Engine pre-load init | 21 s | 2.2% | engine init → model loading start |
| Model download and loading | 62 s | 6.4% | loading start → weights loaded |
| **Engine init, compile, warmup** | **195 s** | **20.2%** | weights loaded → HTTP server up |
| Benchmark + drain | 92 s | 9.5% | HTTP server up → benchmark result |
| Readiness propagation | 1 s | 0.1% | benchmark result → `WorkspaceSucceeded=True` |
| **Total** | **966 s** | **100%** | |

<details>
<summary>Detailed timeline (UTC, 2026-09-07)</summary>

| Event | Timestamp | Elapsed |
|---|---|---:|
| Workspace created | 05:05:46 | 0 s |
| NodeClaim created | 05:05:47 | 1 s |
| VM launched | 05:05:53 | 7 s |
| Kubernetes Node created | 05:10:30 | 284 s |
| Node `Ready` | 05:11:00 | 314 s |
| NodeClaim `Ready` | 05:11:03 | 317 s |
| GPU registered, `ResourceReady=True` | 05:12:36 | 410 s |
| Inference pod and PVC created | 05:12:36 | 410 s |
[...]

PR #2362, opened 2026-09-18
feat: support request abortion during startup benchmark drain

zhehli688:
**Reason for Change**:

This PR adds support for aborting in-flight requests during startup benchmark drain through vLLM's engine.pause_generation api, which helps reduce benchmark drain latency and improve inference cold-start speed.

**Requirements**

- [x] verified drain latency reduced from 12.1s to 1.2s for model Qwen3.8-27B.

**Issue Fixed**:

**Notes for Reviewers**:


zhuangqh:
good job!  link #2325

presets/workspace/inference/vllm/benchmark_entrypoint.py
```diff
@@ -585,12 +599,16 @@ def _run_benchmark() -> tuple:
...
-    """Spin until vllm:num_requests_running reaches zero.
+    """Abort in-flight requests, then wait until none remain.
 
     Raises ``TimeoutError`` after *timeout* seconds so the pod can still
     become Ready rather than hanging forever if vLLM gets stuck.
     """
     _log("drain_start")
+    try:
+        _abort_requests()
+    except Exception as exc:
```
2170chm: Should we separately handle the case if the failure is due to `engine.resume_generation()` since vllm is paused?

zhehli688: Good point! Updated the implementation to raise an error when engine fails to resume.

Fei-Guo (approved):

Fei-Guo (approved):

